PR #206, opened 2014-04-17
add try-in-cloudcare to Tools

sravfeyn:
see https://github.com/dimagi/commcare-hq/pull/3169


src/js/util.js
```diff
@@ -218,6 +218,17 @@ formdesigner.util = (function(){
         throw new Error("Unable to copy obj! Its type isn't supported.");
     }
     that.clone = clone;
+    
+    function invert(obj) {
```
orangejenny: http://underscorejs.org/#invert

orangejenny: ha, i clicked a link from slack and didn't realize this is an old PR. nevermind, Sravan :)

PR #277, opened 2014-11-24
Highlight textarea when tabbing into it

emord:
http://manage.dimagi.com/default.asp?105508#794143

### DIFF
--- a/src/javaRosa.js
+++ b/src/javaRosa.js
@@ -909,6 +909,7 @@ define([
             })
             .on('change keyup', function (e) {
                 widget.updateValue();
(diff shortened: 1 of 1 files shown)


millerdev:
@emord the test added in this PR is failing intermittently when I run the tests in a browser. After a quick look and a few experiments I was not able to fix it, and I'm not sure what's causing the failure. I'm notifying you in case you want to look into it more. I have not seen the failure yet with phantomJS.

emord:
@millerdev looks like the focus() isn't changing selectionStart/selectionEnd. This always fails for me when I run all tests, but when I run just this test it works every time. Very strange.

PR #294, opened 2015-01-07
Add option to display question tree as IDs

orangejenny:


orangejenny:
@millerdev Oof, I didn't even notice that "(_ids)" was displaying. Fixed that, and addressed other comments.

amsagoff:
[?]

orangejenny:
@amsagoff Objections? Concerns?

amsagoff:
nope - looks great with daniel's suggestions! Excited to see your idea
become reality

millerdev:
One more thing. The Display chooser is not visible if there is only one language on the form. You can test this in the stand-alone version by temporarily changing [tests/options.js](https://github.com/dimagi/Vellum/blob/master/tests/options.js#L88)

```js
        javaRosa: {
            langs: ['en'], //'hin'],
            displayLanguage: 'en'
        },
```

orangejenny:
@millerdev Good catch, thank you. Fixed now.

The return statement should now never execute, so I could have just deleted it, but figured maybe someone will remove or change the _ids logic at some point.

PR #301, opened 2015-01-19
Add PATH to bower list

emord:
Matches this: https://github.com/dimagi/Vellum/blob/f9cbcc5575c764f1fbd9a8c2e4b18fcf7bd34e56/Makefile#L17

@millerdev


millerdev:
Looks good to me. I added a commit, so you should review that (and wait for the build).

PR #302, opened 2015-01-19
Disallow duplicate choice

emord:
http://manage.dimagi.com/default.asp?116075#841220

Choice values shouldn't have the same value and different labels

@millerdev

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1020,6 +1020,11 @@ define([
         this.data.core.duplicateIsForMove = forMove;
     };
 
+    fn.setUnsavedDuplicateChoiceValue = function (value, forMove) {
+        this.data.core.unsavedDuplicateChoiceValue = value;
+        this.data.core.duplicateIsForMove = forMove;
+    };
+
     // Attempt to guard against doing actions when there are unsaved or invalid
     // pending changes. In the case of an invalid duplicate sibling ID, it tries
     // to call 'callback' after the user automatically fixes the invalid state,
@@ -1033,7 +1038,10 @@ define([
         var _this = this,
             mug = this.getCurrentlySelectedMug(),
             duplicate = this.data.core.unsavedDuplicateNodeId,
-            duplicateIsForMove = this.data.core.duplicateIsForMove;
+            duplicateChoice = this.data.core.unsavedDuplicateChoiceValue,
+            duplicateIsForMove = this.data.core.duplicateIsForMove,
+            verb = duplicateIsForMove ? 'would have' : 'has';
+
 
         if (this.data.core.hasXPathEditorChanged) {
             this.alert(
@@ -1042,8 +1050,7 @@ define([
                 "changes before continuing.");
             return false;
         } else if (duplicate) {
-            var verb = duplicateIsForMove ? 'would have' : 'has',
-                newQuestionId = this.data.core.form.generate_question_id(duplicate);
+            var newQuestionId = this.data.core.form.generate_question_id(duplicate);
 
             this.alert(
                 "Duplicate Question ID",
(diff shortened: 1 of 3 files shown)


src/form.js
```diff
@@ -353,6 +353,16 @@ define([
                 return m.p.nodeID === nodeID;
             });
         },
+        getMugChildrenByChoiceValue: function (mug, nodeID) {
```
millerdev: Would probably be clearer to rename `nodeID` to `value` since it's not a node ID.

src/form.js
```diff
@@ -353,6 +353,16 @@ define([
...
+            var parentNode = (mug ? this.dataTree.getNodeFromMug(mug)
+                                  : this.dataTree.rootNode);
+            return _.filter(parentNode.value._node_control.children, function (m) {
```
millerdev: Did you consider `_.filter(this.getChildren(mug), ...` instead of the above three lines?

emord: D'oh. I tried too hard to model it after  getMugChildrenByNodeID.

Yours is much better

src/core.js
```diff
@@ -1059,28 +1052,58 @@ define([
...
+
+            this.alert(
+                "Duplicate Question ID",
+                "'" + duplicate + "' " + verb + " the same Question ID as " +
+                "another question in the same group. Please change '" + 
+                duplicate + "' to a unique Question ID before continuing.",
+                alertDuplicateButtons(_this.setUnsavedDuplicateNodeId.bind(_this), "nodeID", newQuestionId)
+            );
+            return false;
+        } else if (duplicateChoice) {
```
millerdev: I'm not sure if it's legal to have a choice with a blank value, but I can't think of why it wouldn't be. In a quick test a choice didn't retain a blank value, but it also wasn't clear why, which may indicate a bug. Anyway, in the case that two choices have blank values this will evaluate to false even though it should evaluate to true, correct? Do you think that's worth fixing?

millerdev: A possible solution may be to say `if (duplicateChoice !== false)`, which is ugly but should work since a choice value is always a string.

emord: https://github.com/dimagi/Vellum/blob/8cc6d2472434b3b6eb76a5e7f9dfdc8859425d28/src/widgets.js#L68

defaultValue and nodeID can't have blank values. In the future if we want blank choice then this will break, but I don't think that changing that code is likely so I don't think we need to handle that situation.

millerdev: Can you add a test that will fail for this if we ever allow an empty default value? That way if we ever do change it we'll know about it. That test should have enough information (a comment?) to remind someone that this is now broken.

tests/core.js
```diff
@@ -33,10 +33,10 @@ require([
             util.init({
                 core: {
                     onReady: function () {
-                        var dup;
                         util.addQuestion("Text", "question1");
-                        dup = util.addQuestion("Text", "question2");
-                        dup.p.nodeID = "question1";
+                        util.addQuestion("Text", "question2");
+                        var target = $("[name='property-nodeID']");
+                        target.val('question1').change();
```
emord: I found when writing my other tests that by modifying the mug directly doesn't go through the set method, which can have side effects. Went ahead and changed this one, but there may be others.

src/core.js
emord: https://github.com/dimagi/Vellum/pull/302/files#diff-905dcd5f60ed0d9e12b350401d356a0aR102

tests/core.js
millerdev: Are you referring to [this set method](https://github.com/dimagi/Vellum/blob/disallow-duplicate-choice/src/mugs.js#L75)? Should investigate why `dup.p.nodeID = "question1";` does trigger change events, etc. since as far as I know many things depend on that.

emord: Hmm. That does seem to be working now. Now I'm not sure why I came to that conclusion anymore

emord: Reverting this last commit as it's unnecessary.

src/form.js
```diff
@@ -624,6 +632,13 @@ define([
                 this.vellum.setUnsavedDuplicateNodeId(value);
                 return null;
             }
+            else if (property === 'defaultValue' && previous &&
```
millerdev: This PR seems to detect choice mugs (*Items*) by whether the mug has a `defaultValue` or not. I think it would be better to also check the class name (`mug.__className === "Item"`), since as far as I know there is nothing preventing other mug types from having a `defaultValue`.

emord: fixed

src/core.js
```diff
@@ -1245,6 +1268,8 @@ define([
             // existing duplicate warning state.
             if (e.property === 'nodeID') {
                 _this.setUnsavedDuplicateNodeId(false);
+            } else if (e.property === 'defaultValue') {
```
millerdev: Should also check `e.mug.__className === "Item"`

src/form.js
```diff
@@ -353,6 +353,14 @@ define([
                 return m.p.nodeID === nodeID;
             });
         },
+        getMugChildrenByChoiceValue: function (mug, value) {
+            return _.filter(this.getChildren(mug), function (m) {
+                if (m.p.defaultValue) {
```
millerdev: should also check if `m.__className === "Item"`

emord: Fixed

PR #306, opened 2015-01-21
Model iteration

millerdev:
This is phase 1 of model iteration. Notable things that are missing:

- easy access to iteration id and index (currently they must be typed by hand in an XPath expression)
- user-friendly data source selector widget
- Update *Model Iteration IDs Query* on rename referenced node (iterate on multi-select)

To use model iteration first create a *Repeat* question, then expand the *Advanced* properties and configure *Model Iteration ID Query*. You now have a model iteration repeat group.


src/mugs.js
```diff
@@ -924,6 +952,7 @@ define([
         typeName: 'Question List',
         icon: 'icon-reorder',
         isSpecialGroup: true,
+        isNestableGroup: false,
```
emord: Should be true. Simplify code below

czue:
hiyo!!!

PR #309, opened 2015-01-22
Reset error messages after question deletion

emord:
http://manage.dimagi.com/default.asp?75697#858316

@millerdev

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1206,6 +1206,7 @@ define([
         }).on('remove-question', function (e) {
             if (!e.isInternal) {
                 var prev = _this.jstree("get_prev_dom", e.mug.ufid);
+                _this.showVisualValidation(null);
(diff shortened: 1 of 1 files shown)


src/core.js
```diff
@@ -1206,6 +1206,7 @@ define([
         }).on('remove-question', function (e) {
             if (!e.isInternal) {
                 var prev = _this.jstree("get_prev_dom", e.mug.ufid);
+                _this.showVisualValidation(null);
```
millerdev: You can simply omit the first argument like so: `_this.showVisualValidation();`

Will this clear all errors, including errors on questions that still exist? If yes, probably should make `showVisualValidation()` re-validate all mugs when no mug is provided.

millerdev: On further investigation it looks like [the bug](http://manage.dimagi.com/default.asp?75697) only appears when deleting the last question on the form.

PR #311, opened 2015-01-26
Preserve unknown attributes on ODK intent

millerdev:
This improves intent handling so Derek doesn't need to use ignore/retain anymore. See also [FB #149925](http://manage.dimagi.com/default.asp?149925).

### DIFF
--- a/src/intentManager.js
+++ b/src/intentManager.js
@@ -19,6 +19,7 @@ define([
             xmlns: "http://opendatakit.org/xforms",
             extra: {},
             response: {},
+            unknownAttributes: {},
             initialNodeID: nodeID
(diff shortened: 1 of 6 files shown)


src/intentManager.js
```diff
@@ -124,7 +137,6 @@ define([
                 };
             intents = dataTree.treeMap(getIntentMugs);
             if (intents.length > 0) {
-                xmlWriter.writeComment('Intents inserted by Vellum:');
```
millerdev: Removed because this was not actually writing a comment (`writeComment` is broken).

millerdev:
@emord

PR #312, opened 2015-01-27
Don't allow group questions to be required

orangejenny:
[3:24:33 PM] Andrea Fletcher: is there ever a time when you would want group questions to be required?
[3:24:45 PM] Andrea Fletcher: does checking that box break things? It seems to....
[3:25:09 PM] Clayton Sims: group questions cannot be required
[3:25:20 PM] Clayton Sims: the UI will not enforce any constraints if you set them to be

### DIFF
--- a/src/mugs.js
+++ b/src/mugs.js
@@ -930,10 +930,7 @@ define([
         init: function (mug, form) {
             Group.init(mug, form);
             mug.p.appearance = 'field-list';
-        },
-        spec: {
-            requiredAttr: { presence: "notallowed" },
-        },
+        }
     });
 
     var Repeat = util.extend(Group, {
@@ -991,8 +988,7 @@ define([
             rawRepeatAttributes: {
                 presence: 'optional',
                 lstring: "Extra Repeat Attributes"
-            },
-            requiredAttr: { presence: "notallowed" },
+            }
         }
     });
    
--- a/src/mugs.js
+++ b/src/mugs.js
@@ -930,7 +930,7 @@ define([
         init: function (mug, form) {
             Group.init(mug, form);
             mug.p.appearance = 'field-list';
-        }
+        },
     });
 
     var Repeat = util.extend(Group, {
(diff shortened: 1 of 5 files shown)


src/mugs.js
```diff
@@ -930,6 +931,9 @@ define([
             Group.init(mug, form);
             mug.p.appearance = 'field-list';
         },
+        spec: {
+            requiredAttr: { presence: "notallowed" },
+        },
```
millerdev: This shouldn't be necessary since FieldList extends Group. Ditto for Repeat below.

millerdev:
Tests are failing. You can run tests in your local dev environment with the `./test` command.
See `./test --help` for more options.

orangejenny:
fixed tests

snopoke:
:+1:

src/mugs.js
```diff
@@ -913,7 +913,8 @@ define([
             calculateAttr: { presence: "notallowed" },
             constraintAttr: { presence: "notallowed" },
             constraintMsgAttr: { presence: "notallowed" },
-            dataValue: { presence: "notallowed" }
+            dataValue: { presence: "notallowed" },
+            requiredAttr: { presence: "notallowed" },
```
millerdev: Comma after last item is not valid JS syntax, unfortunately. Not sure if it matters in practice.

orangejenny: I'd argue it's a good practice to do trailing commas even though they're unofficial - it's minor, but it makes for slightly less code change and slightly fewer "oops, missed a comma" changes while developing.

millerdev: Completely agree in Python where it's allowed. However some JS (maybe only JSON?) parsers will choke on it. Doesn't seem worth much debate if tests pass and it works in Chrome and Firefox.

orangejenny: Yeah, definitely illegal in JSON. I think IE8 and earlier are the only browsers that can't handle it.

PR #313, opened 2015-01-28
Steady cursor when replacing spaces with underscores

orangejenny:
http://manage.dimagi.com/default.asp?61268

Tested in Chrome and Firefox. Not a great candidate for automated testing.

### DIFF
--- a/src/widgets.js
+++ b/src/widgets.js
@@ -135,7 +135,18 @@ define([
                 value = value.replace(
                     new RegExp(String.fromCharCode(10), 'g'), '&#10;');
             }
+            var start = input[0].selectionStart;
+            var end = input[0].selectionEnd;
+            var oldvalue = input.val();
             input.val(value);
+            if (
+                input.is(":focus")
+                && start === end
+                && oldvalue.length === value.length
+            ) {
+                input[0].setSelectionRange(start, start);
+console.log("setting selection range: " + start);
+            }
         };
 
         widget.getValue = function() {
--- a/src/widgets.js
+++ b/src/widgets.js
@@ -142,11 +142,7 @@ define([
 
             // If this input has focus and value hasn't changed much,
             // keep the cursor in the same position
-            if (
-                input.is(":focus")
-                && start === end
-                && oldvalue.length === value.length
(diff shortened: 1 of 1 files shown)


emord:
Why wouldn't this be a good candidate for automated testing? Seems like it could be a pretty easy test to add

orangejenny:
@emord 

```javascript
            ...
            var $input = $("[name='property-nodeID']");
            vellum_util.setCaretPosition($input[0], 3, 3);
            var oldPosition = vellum_util.getCaretPosition($input[0]);
            $input.val("que stion1").change();
            var newPosition = vellum_util.getCaretPosition($input[0]);
            assert.equal(oldPosition, newPosition);
```

This fails, because `val` moves the cursor to the end of the input. I'd originally written the test triggering a keypress on `$input`, but that doesn't seem to actually put the character into the box, so it also fails.

Thoughts?

emord:
@orangejenny 

You tried using the trigger like this: https://github.com/dimagi/Vellum/blob/8cd18e4f3d298748def79be8e7944912a63d95a2/tests/javaRosa.js#L318-L322 ?

Hmm. I'm not sure why that won't work ( @millerdev any ideas?), but if that turns out to be impossible then this looks good to merge

orangejenny:
@emord 

http://jsfiddle.net/z6v4xgef/

Test fiddle that fires off keydown events...the characters don't actually appear in the input. I'm mystified how the test you linked is consistently passing.

Merge?

emord:
Yeah that is weird. Not sure how that one is working

millerdev:
The FB ticket says this fix should only apply to the Question ID field (`mug.p.nodeID`). However, the way it's implemented it will apply to *any* text widget when `setValue` is called and the length of the value hasn't changed. This could have broader effects, although I can't think of how it would matter right now. It's something to keep in mind and possibly (if it causes a problem) implement a new widget type for the Question ID field so this only applies to question ids.

orangejenny:
@millerdev Fair point, I didn't realize that. Agree that it's good to keep in mind but unlikely to be a problem for the time being.

PR #314, opened 2015-01-28
Show correct language when node is opened

orangejenny:
http://manage.dimagi.com/default.asp?148701

Are there any (performance) reasons to be more surgical about this, rather than just refreshing the whole tree?

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -841,6 +841,7 @@ define([
             _this.activateQuestionTypeGroup(mug);
         }).bind("open_node.jstree", function (e, data) {
             _this.activateQuestionTypeGroup(_this.data.core.form.getMugByUFID(data.node.id));
+            _this.refreshVisibleData();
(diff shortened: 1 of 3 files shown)


czue:
(quite likely yes)

src/core.js
```diff
@@ -841,6 +841,7 @@ define([
             _this.activateQuestionTypeGroup(mug);
         }).bind("open_node.jstree", function (e, data) {
             _this.activateQuestionTypeGroup(_this.data.core.form.getMugByUFID(data.node.id));
+            _this.refreshVisibleData();
```
czue: how often does this event fire?

emord: Should only be called when a node is opened from being collapsed

emord:
I don't think there will be a huge performance hit as it's only occurring on opening a node. Could also only update the children nodes of the mug being opened though.

This seems like it would be a good automated test to write though.

PR #315, opened 2015-01-28
Focus on ID when adding new question

orangejenny:
http://manage.dimagi.com/default.asp?59401

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1329,7 +1329,7 @@ define([
             mug = _this.data.core.form.createQuestion(foo[0], foo[1], qType);
             var $firstInput = _this.$f.find(".fd-question-properties input:text:visible:first");
             if ($firstInput.length) {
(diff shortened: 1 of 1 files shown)


src/core.js
```diff
@@ -1327,6 +1327,10 @@ define([
             var foo = _this.getInsertTargetAndPosition(
                 _this.getCurrentlySelectedMug(), qType);
             mug = _this.data.core.form.createQuestion(foo[0], foo[1], qType);
+            var $firstInput = _this.$f.find(".fd-question-properties input:text:visible:first");
+            if ($firstInput.length) {
+                util.setCaretPosition($firstInput[0], 0, $firstInput.val().length);
```
millerdev: This line can be shortened to `$firstInput.focus().select();`, a more standard way to do this.

orangejenny: thanks, fixed

PR #317, opened 2015-01-30
Testing "Question ID" option in languages dropdown

orangejenny:
Test for https://github.com/dimagi/Vellum/pull/294

### DIFF
--- a/tests/javaRosa.js
+++ b/tests/javaRosa.js
@@ -197,29 +197,6 @@ require([
             assert.equal(hinLabel.attr("placeholder"), "English");
         });
 
-        describe("the language selector", function() {
-            before(function(done) {
-                util.init({
-                    javaRosa: { langs: ['en'] },
-                    core: {onReady: done}
-                });
-            });
-    
-            it("should have option to display IDs", function() {
-                util.loadXML("");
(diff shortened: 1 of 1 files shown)


tests/javaRosa.js
```diff
@@ -197,6 +197,29 @@ require([
...
+                $("[name='itext-en-label']").val('english').change();
+    
+                var treeSelector = ".fd-question-tree .jstree-anchor";
+                assert.equal($(treeSelector).text(), "english");
+                var $dropdown = $(".fd-question-tree-lang select");
+                $dropdown.val('_ids').change();
+                assert.equal($dropdown.find("option").length, 2);
+                assert.equal($(treeSelector).text(), "question1");
+            });
+        });
```
millerdev: Looks like a nested `describe` block is executed after all tests in the parent block have finished, which works great for what you're doing here. However, having it at this location in the file could be confusing. What do you think of moving this nested `describe` block outside of it's parent "The javaRosa plugin with multiple languages" block?

I realize the parent `describe` block is not doing setup correctly with `before`, as you are here, and that's unfortunate, but we can address that later/separately.

orangejenny: Sounds logical. Moved it.

PR #318, opened 2015-01-30
Non group child

emord:
https://trello.com/c/wXNt4IL2/26-have-an-element-of-a-group-that-is-not-a-child-of-the-group-s-parent


src/core.js
```diff
@@ -1288,10 +1288,20 @@ define([
                 _this.setUnsavedDuplicateNodeId(false);
             } else if (e.property === 'defaultValue' && e.mug.__className === 'Item') {
                 _this.setUnsavedDuplicateChoiceValue(false);
+            } else if (e.property === 'dataParent') {
+                // TODO Dirty write hack that needs to be replace
+                e.mug.options.isDataOnly = true;
```
emord: Did this to make sure the control tree doesn't change. Could also extend this to use another param or refactor this into another method to insert into data only.

Thoughts on what would be better?

src/core.js
```diff
@@ -1288,10 +1288,20 @@ define([
...
+                _this.data.core.form.insertMug(e.mug.parentMug, e.mug, 'into');
+                e.mug.options.isDataOnly = false;
             }
 
             _this.refreshMugName(e.mug);
             _this.toggleConstraintItext(e.mug);
+        }).on('question-move', function(e) {
+            if (e.mug.spec.dataParent &&
```
emord: If a question is moved and the parent isn't valid reset the data parent to be undefined

emord:
@millerdev 

This should have proper functionality, but I think I may have not followed proper Vellum patterns in some places, so this is probably more of a review that I'm writing maintainable code.

src/mugs.js
```diff
@@ -293,7 +293,45 @@ define([
...
+            dataParent: {
+                lstring: 'Data Parent',
+                visibility: function(mug) {
+                    function recFunc(mug) {
+                        if (!mug) {
+                            return true;
+                        } else if (mug.__className === 'Repeat') {
+                            return false;
+                        }
+                        return recFunc(mug.form.controlTree.getNodeFromMug(mug).parent.value);
```
millerdev: What happens if this is called on a data-only mug (one that is in the data tree but not the control tree)? Will `getNodeFromMug(mug) -> null/undefined` and then `.parent -> Error`?

src/mugs.js
```diff
@@ -293,7 +293,45 @@ define([
...
+
+                    if (dataParent &&
+                        form.getBasePath().slice(0, -1) !== dataParent &&
+                        !dataParentMug) {
+                        return "Must be valid path";
+                    } else if (dataParentMug && dataParentMug.__className !== 'Group') {
+                        return "Data must be a child of a group";
+                    } else if (dataParent && !mug.spec.dataParent.visibility(mug)) {
+                        return "Children of repeat groups cannot have a different data parent";
+                    }
```
millerdev: It seems that there would be no need to do any of these checks if `dataParent` evaluates to false. Could indent them and put them inside `if (dataParent) {`, correct?

tests/diffDataParent.js
```diff
@@ -0,0 +1,95 @@
...
+    util,
+    PARSE_XML
+) {
+    var assert = chai.assert,
+        call = util.call;
+
+    describe("Control elements with different data parents", function() {
+        before(function (done) {
+            util.init({
+                javaRosa: {langs: ['en', 'hin']},
```
millerdev: What do you think of removing one of these languages? I don't believe the features being tested here have any impact on multi-language forms, and it would reduce the size of test XML and possibly make the tests a bit faster.

tests/diffDataParent.js
```diff
@@ -0,0 +1,95 @@
...
+            util.loadXML("");
+            var text1 = util.addQuestion("Text", 'text1'),
+                group1 = util.addQuestion("Group", 'group1'),
+                group2 = util.addQuestion.bind({prevId: text1.p.nodeID})("Group", 'group2'),
+                form = call("getData").core.form;
+
+            text1.p.dataParent = '/data/group1';
+            form.moveMug(group1, group2, 'into');
+            util.clickQuestion(text1.p.nodeID);
+            assert(!util.isTreeNodeValid(text1), "text1 should not be valid");
```
millerdev: Would it be possible to make `dataParent` automatically update to the new parent location when the parent is moved?

tests/diffDataParent.js
```diff
@@ -0,0 +1,95 @@
...
+
+        it("should clear the data parent when moving to a repeat group", function() {
+            util.loadXML("");
+            var text1 = util.addQuestion("Text", 'text1');
+            util.addQuestion("Group", 'group1');
+            var repeat1= util.addQuestion.bind({prevId: text1.p.nodeID})("Repeat", 'repeat1'),
+                form = call("getData").core.form;
+            text1.p.dataParent = '/data/group1';
+            form.moveMug(text1, repeat1, 'into');
+            assert(util.isTreeNodeValid(text1), "text1 should be valid");
```
millerdev: Should also `assert.equal(text1.p.dataParent, '')`?

tests/diffDataParent.js
```diff
@@ -0,0 +1,95 @@
...
+        });
+
+        it("should parse and write XML to have the same order", function() {
+            util.loadXML(PARSE_XML);
+            util.assertXmlEqual(call("createXML"), PARSE_XML);
+        });
+
+        it("should have proper data parent after being loaded from xml", function() {
+            util.loadXML(PARSE_XML);
+            var question1 = call("getMugByPath", "/data/question4/question1");
```
millerdev: FYI there's a nice convenience function you can use here: `util.getMug("question4/question1")`. It basically does the same thing, but it's more concise and it can get select items too.

tests/diffDataParent.js
```diff
@@ -0,0 +1,95 @@
...
+            assert.equal(question1.p.dataParent, "/data/question4");
```
millerdev: Would it make sense to also assert the JSTree state (`util.assertJSTreeState`) to ensure that `question1` is outside `question4` in the control tree?

src/parser.js
```diff
@@ -606,6 +606,11 @@ define([
             if (mug.options.isControlOnly) {
                 // HACK fix abstraction broken by direct tree insert
                 form.mugMap[mug.ufid] = mug;
+            } else if (form.getAbsolutePath(mug) &&
+                       form.getAbsolutePath(mug) !== form.getControlPath(mug)) {
+                var parentPath = form.getAbsolutePath(mug).split('/');
```
millerdev: Splitting the path might be problematic since removing the last element of a mug's path is not guaranteed to get the parent mug's path (this is at least true for model iteration Repeat questions). Would be better to get the parent path like this:
```js
// might need to check if the node is null/undefined before traversing to .parent
parentNode = form.dataTree.getNodeFromMug(mug).parent
mug.p.dataParent = parentNode.getAbsolutePath()
```
This is going to require a refactor of `Tree.getAbsolutePath`: move/rename [`pathOf`](https://github.com/dimagi/Vellum/blob/non-group-child/src/tree.js#L279) to `Node.getAbsolutePath()` so you can call `parent.getAbsolutePath()` even if parent happens to be the root node.

tests/diffDataParent.js
millerdev: Don't forget to remove the corresponding translation block from parse.xml if you remove one of these.

src/core.js
```diff
@@ -1288,10 +1288,20 @@ define([
                 _this.setUnsavedDuplicateNodeId(false);
             } else if (e.property === 'defaultValue' && e.mug.__className === 'Item') {
                 _this.setUnsavedDuplicateChoiceValue(false);
+            } else if (e.property === 'dataParent') {
+                // TODO Dirty write hack that needs to be replace
+                e.mug.options.isDataOnly = true;
+                _this.data.core.form.insertMug(e.mug.parentMug, e.mug, 'into');
```
millerdev: Insert the parent as a child of it's child? That doesn't seem right. Seems like we need a new function on the form that knows how to change the data path of a mug: `e.mug.form.updateDataParent(e.mug)`.

tests/diffDataParent.js
emord: Glad you mentioned this. Looks like there is a pretty nasty bug here that doesn't allow the data parents to be moved.

src/mugs.js
emord: I tested this with hidden values. That code doesn't get to this function because dataParent is not in the spec. Should I also test it with any other mugs (didn't see any others when I first searched through)?

emord:
Closing this as I originally put this before model-iteration was in master and you can't change target branches. Will open a new one once I get issues brought up in this PR completed and cross reference the comments

millerdev:
It might have worked to rebase this branch on top of master after `model-iteration` was merged. No worries, an new branch and PR is fine too.

src/mugs.js
millerdev: I can't think of any others

PR #319, opened 2015-02-02
155189 fix tsv encoding

millerdev:
http://manage.dimagi.com/default.asp?155089
http://manage.dimagi.com/default.asp?155190

### DIFF
--- a/src/javaRosa.js
+++ b/src/javaRosa.js
@@ -1257,63 +1257,49 @@ define([
     };
 
     var generateItextXLS = function (vellum, Itext) {
-        // todo: fix abstraction barrier
-        vellum.beforeSerialize();
-
-        // TODO move TSV generation logic into tsv module
         function getItemFormValues(item, languages, form) {
-
-            var ret = [], i, value, language;
-
+            var ret = [], i, language;
             for(i = 0; i < languages.length; i++) {
                 language = languages[i];
-                value = item.hasForm(form) ? (item.get(language, form) || "") : "";
-
-                if (specialChars.test(value)) {
-                    // quote field
-                    value = '"' + value.replace(/"/g, '""') + '"';
-                }
-                ret.push(value);
+                ret.push(item.hasForm(form) ? item.get(language, form) : "");
             }
-            return ret.join("\t");
+            return ret;
         }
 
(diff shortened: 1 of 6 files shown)


src/javaRosa.js
```diff
@@ -1257,64 +1257,49 @@ define([
...
+            for (i = 0; i < exportCols.length; i++) {
+                for (j = 0; j < languages.length; j++) {
                     header_row.push(exportCols[i] + '-' + languages[j]);
                 }
             }
-            return header_row.join("\t");
+            return header_row;
         }
 
-        var specialChars = /[\r\n\u2028\u2029"]/g;
```
millerdev: This (specifically the `g` flag) was the problem for bulk translations. Global RegExp objects are stateful, which caused some cells to not be escaped even though they needed it!

All other modifications to this file are cleanup and refactoring to use the tsv module for TSV encoding.

src/tsv.js
```diff
@@ -3,6 +3,8 @@ define([
 ], function (
     _
 ) {
+    var specialChars = /[\r\n\u2028\u2029"]/;
```
emord: Went through the backlog yesterday. Adding \t here should fix http://manage.dimagi.com/default.asp?104522#794142 as well correct?

src/tsv.js
```diff
@@ -72,8 +74,28 @@ define([
         return rows;
     }
 
+    function escape(value) {
+        if (value === null || _.isUndefined(value)) {
```
orangejenny: Should also test for value === false. Or do `if (!value)` - this seems like a case where writing non-strict code actually does the right thing.

millerdev: I had originally written `if (!value)`, but that converts `0` to an empty string which is almost certainly not desired (see [the test for that](https://github.com/dimagi/Vellum/blob/155189-fix-tsv-encoding/tests/tsv.js#L101)). This is meant to handle the things you'd likely want to serialize in a TSV file, which includes strings and numbers. `false` falls into the category of stuff that will probably just not do the right thing if you serialize it. Objects and arrays also fall into that category.

src/tsv.js
millerdev: Good catch! How did I miss that one?

PR #320, opened 2015-02-03
Revert "Disallow duplicate choice"

millerdev:
Reverts dimagi/Vellum#302


millerdev:
Reverted because this feature should only warn about duplicate values, not prevent them from being created/saved.

PR #321, opened 2015-02-04
Fix save button activation on load form

millerdev:
@czue @emord

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1214,6 +1214,10 @@ define([
         this.data.core.form = form = parser.parseXForm(
             formXML, options, this, _this.data.core.parseWarnings);
         form.formName = this.opts().core.formName || form.formName;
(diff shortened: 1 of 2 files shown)


src/core.js
```diff
@@ -1214,6 +1214,10 @@ define([
         this.data.core.form = form = parser.parseXForm(
             formXML, options, this, _this.data.core.parseWarnings);
         form.formName = this.opts().core.formName || form.formName;
+        if (formXML) {
+            _this._resetMessages(_this.data.core.form.errors);
+            _this._populateTree();
+        }
```
millerdev: Populate tree before binding form event handlers so mug property changes in `handleMugParseFinish` do not trigger a save button state change.

czue: :+1:

PR #323, opened 2015-02-05
Only one tree

emord:
No longer have to keep track of both the data and control tree.

The control tree is the master and we use the flags isControlOnly and isDataOnly whenever we need to discern between those.

### DIFF
--- a/src/form.js
+++ b/src/form.js
@@ -395,30 +395,13 @@ define([
          * single argument is the mug.
          */
         walkMugs: function (valueFunc) {
-            function makeGenerator(items) {
-                var i = 0;
-                if (!items) {
-                    return function () { return null; };
+            function callback(mug, nodeID, processChildren) {
+                if (mug !== null) {
+                    valueFunc(mug);
                 }
-                return function () {
-                    return i < items.length ? items[i++] : null;
-                };
+                processChildren();
             }
-            function visitNode(node) {
-                var nodeChildren = node.getChildren(),
-                    nextChild = makeGenerator(nodeChildren),
-                    child = nextChild(),
-                    value = node.getValue();
-
-                if (value) {
-                    valueFunc(value);
-                }
-                while (child) {
-                    visitNode(child);
-                    child = nextChild();
-                }
-            }
-            visitNode(this.tree.getRootNode());
+            this.tree.walk(callback);
         },
         getDescendants: function (mug) {
             var desc = this.getChildren(mug), i;
--- a/src/form.js
+++ b/src/form.js
@@ -384,17 +384,16 @@ define([
                 this.dataTree.insertMug(newMug, position, refMug);
             }
 
-            if (!newMug.options.isDataOnly) {
-                if (refMug && refMug.options.isDataOnly) {
-                    if (position !== 'after' && position !== 'before') {
-                        // should never happen
-                        throw new Error("cannot insert " + position + " " + refMug.__className);
-                    }
-                    // find alternate insert refMug and position
-                    var newRefMug = this.dataTree.findSibling(
-                                    refMug,
-                                    (position === 'after' ? 'before' : 'after'),
-                                    function (mug) { return !mug.options.isDataOnly; });
+            if (refMug && refMug.options.isDataOnly) {
+                if (position !== 'after' && position !== 'before') {
+                    // should never happen
+                    throw new Error("cannot insert " + position + " " + refMug.__className);
+                }
+                // find alternate insert refMug and position
+                var newRefMug = this.dataTree.findSibling(
+                    refMug,
+                    (position === 'after' ? 'before' : 'after'),
+                    function (mug) { return !mug.options.isDataOnly; });
                     if (newRefMug) {
                         refMug = newRefMug;
                     } else {
@@ -402,9 +401,8 @@ define([
                         refMug = refMug.parentMug;
                         position = (position === 'after' ? 'first' : 'last');
                     }
-                }
-                this.controlTree.insertMug(newMug, position, refMug);
             }
+            this.controlTree.insertMug(newMug, position, refMug);
         },
         /**
          * Move a mug from its current place (in both the Data and Control trees) to
--- a/src/form.js
+++ b/src/form.js
@@ -313,19 +313,13 @@ define([
          * autocomplete skip logic wizard.
          */
         getMugList: function () {
-            var treeFunc, cList;
-
-            treeFunc = function (node) {
+            return this.controlTree.treeMap( function (node) {
                 if(node.isRootNode) {
                     return;
                 }
 
                 return node.getValue();
-            };
-
-            cList = this.controlTree.treeMap(treeFunc);
-
-            return cList;
+            });
         },
         updateError: function (errObj, options) {
             errObj = FormError(errObj);
@@ -719,7 +713,7 @@ define([
         /**
          * Get the logical path of the mug's node in the data tree
          *
-         * It is not always possible to lookup a mug by traversing either the
+         * It is not always possible to lookup a mug by traversing the
          * control tree using it's absolute path. For example, some
          * mugs encapsulate multiple levels of XML elements. This Form object
          * maintains a hash table to quickly get a mug by its path.
--- a/src/form.js
+++ b/src/form.js
@@ -406,23 +406,12 @@ define([
             this.fireChange(mug);
         },
         /**
-         * Walk merged data and control trees calling valueFunc for each mug
-         *
-         * If common nodes in merged tree branches are not in the same order
-         * then the order of nodes in the control tree will be preserved, and
-         * all nodes in the data tree that are not also in the control tree will
-         * be visited last.
+         * Walk through the mugs of the tree and call valueFunc on each mug
          *
          * @param valueFunc - a function called for each mug in the tree. The
          * single argument is the mug.
          */
-        mergedTreeMap: function (valueFunc) {
-            function getID(node) {
-                return node.getID() || node.getValue().ufid;
-            }
-            function hasControl(node) {
-                return !node.getValue().options.isDataOnly;
-            }
+        walkMugs: function (valueFunc) {
             function makeGenerator(items) {
                 var i = 0;
                 if (!items) {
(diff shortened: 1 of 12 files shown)


tests/static/all_question_types.tsv
```diff
@@ -28,6 +28,6 @@ Question	Type	Text (en)	Text (hin)	Audio (en)	Audio (hin)	Image (en)	Image (hin)
 /question22/question23/question28	Password	question28	question28									no
 /question22/question23/question29	Signature Capture	question29	question29									no
 /question22/question23/question7	Android App Callout	question7	question7									no
-	Base											no
 /question20	Hidden Value											no
-/question32	Hidden Value										1 + 2	no
\ No newline at end of file
+/question32	Hidden Value										1 + 2	no
+	Base											no
```
emord: I changed this because it seemed to just fall out of the old implementation and I didn't see any comments about how we want this to stay the same.

Basically if something is only in the control path and at the end of the control body (like this piece of custom xml) it will end up getting put at the end.

Is this acceptable? I could only find this one edge case and it didn't seem to affect the end product (didn't change the xml on this same test)

src/form.js
```diff
@@ -313,32 +307,13 @@ define([
             return writer.createXForm(this);
         },
         /**
-         * Goes through and grabs all of the data nodes (i.e. nodes that are
-         * only data nodes, possibly with a bind) without any kind of control.
-         * Returns a flat list of these nodes (list items are mugs).
-         */
-        getDataNodeList: function () {
```
emord: This was never used in vellum

src/form.js
```diff
@@ -155,10 +155,6 @@ define([
 
         this.formName = 'New Form';
         this.mugMap = {};
-        this.dataTree = new Tree('data', 'data');
-        this.dataTree.on('change', function (e) {
-            _this.fireChange(e.mug);
-        });
         this.controlTree = new Tree('data', 'control');
```
millerdev: Can we rename this to `tree` now that there is only one?

src/form.js
```diff
@@ -177,8 +173,7 @@ define([
 
     Form.prototype = {
         getBasePath: function () {
-            // the choice to use dataTree instead of controlTree is arbitrary
```
millerdev: Heh, when I wrote this comment I only meant it within the context of this function. Kind of amazing that you were able to run with it and figured out a way to apply it to all of Vellum.

src/form.js
```diff
@@ -313,45 +307,19 @@ define([
...
                     return;
                 }
 
                 return node.getValue();
-            };
-
-            cList = this.controlTree.treeMap(treeFunc);
-            dList = this.dataTree.treeMap(treeFunc);
-
-            return util.mergeArray(cList, dList); //strip dupes and merge
```
millerdev: Looks like this was the only use of `mergeArray`, and  `_.uniq` does the same thing. Feel free to remove it from util.js.

src/form.js
```diff
@@ -389,53 +357,28 @@ define([
...
-                        refMug = refMug.parentMug;
-                        position = (position === 'after' ? 'first' : 'last');
-                    }
-                }
-                this.controlTree.insertMug(newMug, position, refMug);
-            }
+            this.controlTree.insertMug(newMug, position, refMug);
         },
         /**
          * Move a mug from its current place (in both the Data and Control trees) to
```
millerdev: Should these documentation references to data and control trees be removed?

src/form.js
```diff
@@ -446,23 +389,12 @@ define([
             this.fireChange(mug);
         },
         /**
-         * Walk merged data and control trees calling valueFunc for each mug
-         *
-         * If common nodes in merged tree branches are not in the same order
-         * then the order of nodes in the control tree will be preserved, and
-         * all nodes in the data tree that are not also in the control tree will
-         * be visited last.
```
millerdev: I assume this behavior has moved into the parser? (haven't gotten there yet) I think it's important that we preserve this behavior and explicitly mention it in the place where it is implemented.

src/form.js
```diff
@@ -446,23 +389,12 @@ define([
...
          * single argument is the mug.
          */
-        mergedTreeMap: function (valueFunc) {
-            function getID(node) {
-                return node.getID() || node.getValue().ufid;
-            }
-            function hasControl(node) {
-                return !node.getValue().options.isDataOnly;
-            }
+        walkMugs: function (valueFunc) {
```
millerdev: This got quite a bit simpler, which is nice. It should be possible to make it simpler yet with `Tree.walk`. This example has not been tested, so may need to be tweaked:

```js
walkMugs: function (valueFunc) {
    function callback(mug, nodeID, processChildren) {
        if (mug !== null) {
            valueFunc(mug);
        }
        processChildren();
    }
    this.tree.walk(callback);
}
```

src/form.js
millerdev: Got to the parser. It looks like the data tree order will be preserved, and the control tree order will be ignored/lost. I'm surprised that there are no tests that verify this comment. Please add a test that has data and control trees with mismatched order and verify that the control tree order is preserved after a parse/write cycle.

I'm not so worried about the last part of this comment "all nodes in the data tree that are not also in the control tree will be visited last." I think it's safe to omit that requirement if that's hard to implement.

src/writer.js
```diff
@@ -156,7 +157,7 @@ define([
     };
 
     var createBindList = function (form, xmlWriter) {
-        form.dataTree.walk(function (mug, nodeID, processChildren) {
+        form.controlTree.walk(function (mug, nodeID, processChildren) {
             if(mug) {
```
millerdev: This should probably say `&& !mug.isControlOnly`

tests/form.js
```diff
@@ -111,40 +110,5 @@ define([
...
-        it("should prefer control tree order on merge", function () {
-            var form = new Form({}),
-                values = [];
-
-            form.dataTree = makeTree("data", ["a", "x1", "b", "x2", "c"]);
-            form.controlTree = makeTree("control", ["a", "c", "b"]);
-
-            form.mergedTreeMap(function (v) { values.push(v.id); });
-            assert.equal(values.join(" "), "a c b x1 x2");
-        });
```
millerdev: Ha! I knew there were tests for this. These should be converted to XForm with equivalent structure and write tests to verify that the parser preserves order as specified here.

tests/static/all_question_types.tsv
millerdev: I'm not quite understanding why content from custom XML is showing up in this file, but what you're saying seems reasonable. I think we should probably make it skip this content entirely, but we can address that in a different PR.

emord: I think it's because custom xml is turned into a ReadOnly Mug that can't be edited in Vellum, but still shows up. I'm not sure of it's use here, but didn't really investigate further

tests/form.js
emord: https://github.com/dimagi/Vellum/commit/03af95757856b736ead20fd0be2802a7fcafc2b8

src/form.js
emord: Yep. that worked perfectly

tests/static/form/mismatch-tree-order.xml
```diff
@@ -0,0 +1,55 @@
...
+        </input>
+        <group ref="/data/question2">
+            <label ref="jr:itext('question2-label')" />
+            <input ref="/data/question2/question3">
+                <label ref="jr:itext('question2/question3-label')" />
+            </input>
+        </group>
+        <input ref="/data/question4">
+            <label ref="jr:itext('question2/question4-label')" />
+        </input>
```
millerdev: The data tree order here is 1, 5, 2, 6, 4 with 5 and 6 being data-only, and control tree order is 1, 2, 4. Those are not mismatched trees, correct? I think you want to change this so the control tree order is 1, 4, 2.

The old test did this:
```js
form.dataTree = makeTree("data", ["a", "x1", "b", "x2", "c"]); // a b c
form.controlTree = makeTree("control", ["a", "c", "b"]);       // a c b
```

src/parser.js
```diff
@@ -598,27 +598,55 @@ define([
...
+                if (numControlStructs > index) {
+                    break;
+                }
             }
-            if (mug.options.controlNodeChildren) {
+
+            form.tree.insertMug(mug, 'index', parentMug, realIndex);
+
+            if (mug.__className !== "ReadOnly" && mug.options.controlNodeChildren) {
+                var index1 = 0;
```
millerdev: Very minor: might want to rename this to `child_index`. I would not even bring this up except it made code review a little harder. It wasn't easy to spot the difference between `index` and `index1`.

emord: Variable names are hard

src/parser.js
```diff
@@ -598,27 +598,55 @@ define([
...
+                    numControlStructs++;
+                }
+
+                if (numControlStructs > index) {
+                    break;
+                }
             }
-            if (mug.options.controlNodeChildren) {
+
+            form.tree.insertMug(mug, 'index', parentMug, realIndex);
```
millerdev: Can this second insert be avoided (if `realIndex` is the same as the the mug's current index)? In most cases it will be unnecessary. It's a shame to have to add this algorithmic complexity to the parser, given that it's only aimed at rare forms with mismatched control and data trees.

emord: I think I solved this in O(n) now

emord: Err. maybe not O(n), but at least not as terrible as it once once.

tests/form.js
```diff
@@ -112,39 +113,16 @@ define([
...
-        it("should merge data-only nodes with control nodes", function () {
-            var form = new Form({}),
-                values = [];
-
-            form.dataTree = makeTree("data", ["a", "x1", "b", "x2", "c"]);
-            form.controlTree = makeTree("control", ["a", "b", "c"]);
-
-            form.mergedTreeMap(function (v) { values.push(v.id); });
-            assert.equal(values.join(" "), "a x1 b x2 c");
-        });
```
millerdev: After a bit more reflection, I think it would be good to rewrite this test as well. I'm not sure if there are any other tests that verify that the order of hidden values interspersed with other questions is preserved.

millerdev:
Looks good. I'm tempted to be nitpicky about [names again](https://github.com/dimagi/Vellum/commit/317c7286026717552e3f564f040c3c1ba27511ae#diff-c1133de887698faf967f05bb89743782R133), but oh well... We can merge this once QA passes.

PR #324, opened 2015-02-06
Display brackets whenever label is question ID

orangejenny:
http://manage.dimagi.com/default.asp?156126

Just do a string comparison, since the itext widget sometimes populates "blank" labels with the id.

### DIFF
--- a/src/mugs.js
+++ b/src/mugs.js
@@ -574,7 +574,7 @@ define([
             }
 
             if (!itextItem || lang === '_ids') {
-                return '[' + nodeID + ']';
+                return nodeID;
             }
             lang = lang || defaultLang;
 
@@ -592,7 +592,7 @@ define([
                 return disp;
             }
 
-            return '[' + nodeID + ']';
+            return nodeID;
         },
         // todo: move these into javarosa
         getItextAutoID: function (propertyPath) {
--- a/src/mugs.js
+++ b/src/mugs.js
@@ -563,7 +563,8 @@ define([
                 Itext = this.form.vellum.data.javaRosa.Itext,
                 defaultLang = Itext.getDefaultLanguage(),
                 disp,
-                defaultDisp;
+                defaultDisp,
+                nodeID;
 
             if (this.__className === "ReadOnly") {
                 return "Unknown (read-only) question type";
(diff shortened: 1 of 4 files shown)


orangejenny:
Updated so that whenever an id is shown, for any reason, it has brackets. This changes these cases:
- The original FB ticket
- Hidden variables will now display in brackets
- Selecting "Question ID" from the languages dropdown will give everything brackets

I figured this is at least consistent, but am open to changing it if a reasonable person ( @amsagoff ? ) thinks this is too many brackets.

amsagoff:
Jenny, what are your feelings on the brackets in general? I'm pro
consistency (either all bracketed or all not bracketed), though not sure
it's important to have a visual distinction between question IDs and labels
(it's generally pretty clear).

orangejenny:
@amsagoff 

I don't like the brackets - if we're going to differentiate, we should do it with a style (color, italics, etc.), not something text-based.

But if you don't think differentiating is adding value, I like the thought of getting rid of them altogether...simpler is better. Let's do that.

src/mugs.js
```diff
@@ -573,7 +574,7 @@ define([
             }
 
             if (!itextItem || lang === '_ids') {
-                return this.getNodeID();
+                return '[' + nodeID + ']';
```
millerdev: What do you think about *not* displaying brackets when `lang === '_ids'`? Similar to displaying the default language in brackets when the selected translation is empty, the node ID is wrapped in brackets only when it is used as a fallback because the labels are empty. This minimizes visual noise when the display is set to _ids.

orangejenny: Sounds fine to me - I agree that it's getting to be a lot of brackets.

What do you think of the possibility of removing brackets altogether?

millerdev:
Heh, should have read the discussion before jumping into the code.

+1 on not using brackets, ever. I think they're ugly.

millerdev:
I should clarify:
- Do not use brackets to wrap Question IDs
- Continue to use brackets to wrap the translation abbreviation if the selected language is empty.

amsagoff:
Like this outcome! Will try to do a pass of all vellum p4 and higher bugs &
easy_wins to make sure there aren't any other terrible ideas I once had
still lying around...

orangejenny:
Consensus, great. Dropped all brackets around IDs.

PR #325, opened 2015-02-09
Mismatched parents

emord:
@millerdev 

Not ready to merge yet. Right now it adds an extra item in model iteration and I haven't been able to find out where that is quite yet. Also on every successive save it saves more items under item so it ends up like:

```
<item>
  <item />
  <item />
  <question1>
</item>
```

any ideas?


src/form.js
```diff
@@ -172,6 +172,40 @@ define([
     }
 
     Form.prototype = {
+        dataTree: function() {
```
emord: Seems almost silly to have this after taking out the original data tree, but I like that we can see how data tree relates to the tree in one function. Thought about also providing a helper for control tree.

src/form.js
```diff
@@ -701,7 +736,10 @@ define([
          * maintains a hash table to quickly get a mug by its path.
          */
         getAbsolutePath: function (mug, excludeRoot) {
-            if (!mug.options.isControlOnly) {
+            if (mug && !mug.options.isControlOnly) {
```
millerdev: I realize this is a convenience, but it should not be necessary. It's a bug to call this function with anything but a valid mug. It is the caller's responsibility to know that it's passing a valid mug.

src/mugs.js
```diff
@@ -295,7 +295,44 @@ define([
...
+            dataParent: {
+                lstring: 'Data Parent',
+                visibility: function(mug) {
+                    function recFunc(mug) {
+                        if (!mug) {
+                            return true;
+                        } else if (mug.__className === 'Repeat') {
+                            return false;
+                        }
+                        return recFunc(mug.form.tree.getNodeFromMug(mug).parent.value);
```
millerdev: Does `mug.form.tree.getNodeFromMug(mug).parent.value` return the same thing as `mug.parentMug`?

src/mugs.js
```diff
@@ -295,7 +295,44 @@ define([
...
+                        dataParentMug;
+
+                    if (dataParent) {
+                        dataParentMug = form.getMugByPath(dataParent);
+
+                        if(!dataParentMug &&
+                           form.getBasePath().slice(0, -1) !== dataParent) {
+                            return "Must be valid path";
+                        } else if (dataParentMug && dataParentMug.__className !== 'Group') {
+                            return "Data must be a child of a group";
```
millerdev: Should the visibility function above enforce this too?

src/writer.js
```diff
@@ -157,8 +156,8 @@ define([
     };
 
     var createBindList = function (form, xmlWriter) {
-        form.tree.walk(function (mug, nodeID, processChildren) {
-            if(mug && !mug.options.isControlOnly) {
+        form.dataTree().walk(function (mug, nodeID, processChildren) {
```
millerdev: The writer should probably construct the data tree once and pass it into `createDataBlock` as well as this function.

src/mugs.js
millerdev: If not, is there any reason why we can't declare an invariant that they will always be the same?

src/parser.js
```diff
@@ -339,6 +339,9 @@ define([
             mug = form.getMugByPath(path);
         }
         mug = adapt(mug, form);
+        if (parentMug !== mug.parentMug) {
+            mug.p.dataParent = form.getAbsolutePath(mug.parentMug);
+        }
         mug.parentMug = parentMug;
```
millerdev: This can move into `if` block above it.

src/mugs.js
millerdev: Nevermind.

tests/diffDataParent.js
```diff
@@ -0,0 +1,133 @@
...
+            util.addQuestion("Group", 'group1');
+            var repeat1= util.addQuestion.bind({prevId: text1.p.nodeID})("Repeat", 'repeat1'),
+                form = call("getData").core.form;
+            text1.p.dataParent = '/data/group1';
+            form.moveMug(text1, repeat1, 'into');
+            assert(util.isTreeNodeValid(text1), "text1 should be valid");
+            assert.isUndefined(text1.p.dataParent);
+        });
+
+        it("should have proper data parent after change through UI", function() {
```
millerdev: I don't see where the UI is being changed in this test. Can you explain the name?

Possible suggested name: "should have proper data parent on reload form"

tests/diffDataParent.js
```diff
@@ -0,0 +1,133 @@
...
+
+        it("should have proper data parent after being loaded from xml", function() {
+            util.loadXML(PARSE_XML);
+            var question1 = util.getMug("/data/question4/question1");
+            assert.equal(question1.p.dataParent, "/data/question4");
+            util.assertJSTreeState(
+                "question1",
+                "question4",
+                "  question3"
+            );
```
millerdev: This feature (the whole PR) is so unintuitive. I can't believe they're having us implement it :)

tests/diffDataParent.js
```diff
@@ -0,0 +1,133 @@
...
+
+        it("should use the control parent when inserting after a mug with different data parent", function() {
+            util.loadXML("");
+            var form = call("getData").core.form,
+                text1 = util.addQuestion("Text", 'text1');
+            util.addQuestion("Group", 'group');
+            util.addQuestion("Text", 'text2');
+
+            text1.p.dataParent = '/data/group';
+            util.addQuestion.bind({prevId: text1.p.nodeID})("Text", 'text3');
```
millerdev: Might be useful to `assertJSTreeState` here as well, so it's easier to see that text3 got inserted at the correct place in the control tree.

tests/utils.js
```diff
@@ -75,6 +75,29 @@ define([
         }
     }
 
+    function assertDataTreeState(tree) {
```
millerdev: What do you think of renaming this to `assertTreeState` since the provided tree could also be a control tree?

src/mugs.js
millerdev: Although what about `FieldList`?

src/mugs.js
millerdev: It does now: [mugs.js:Mug.prototype.parentPath](https://github.com/dimagi/Vellum/pull/325/files#diff-fb1b9e010259d16ab20a8186e2d67b23R684)

src/parser.js
```diff
@@ -339,12 +335,16 @@ define([
             mug = form.getMugByPath(path);
         }
         mug = adapt(mug, form);
-        mug.parentMug = parentMug;
+        var node = form.tree.getNodeFromMug(mug);
+        if (node && node.parent.value !== parentMug) {
+            mug.p.dataParent = node.parent.getAbsolutePath();
+            form.tree.insertMug(mug, 'into', parentMug);
```
millerdev: Hmm, this could be problematic. Will the control tree order be preserved in all cases? May need to add a test or two where the special mug with a dataParent comes before, after, and/or sandwiched between other normal mugs.

src/mugs.js
emord: Can't check `isNestableGroup` as repeat groups have that same attribute. Will create a new one called `possibleDataParent` though

tests/diffDataParent.js
emord: Yeah. I think that test name may have made sense when I first wrote this before the refactor, but no longer makes sense

src/parser.js
emord: Added three tests that test that. Control tree order was preserved in all.

tests/diffDataParent.js
```diff
@@ -0,0 +1,196 @@
...
+        });
+
+        it("should properly load mug after other mugs", function() {
+            util.loadXML(DATA_PARENT_MUG_AFTER);
+            var text = util.getMug("/data/question4/question1");
+            assert.equal(text.p.dataParent, "/data/question4");
+            util.assertJSTreeState(
+                "question3",
+                "question1",
+                "question4"
```
millerdev: This PR is looking good. One final nitpick. The question names in these tests are rather incomprehensible at first glance. What do you think of using names like
```js
"before-child",
"child",
"parent"
```
Also might want to rename the local `text` variable to `child`.

emord:
@millerdev 

I think all of your comments have been addressed, and I think the functionality is there and tested.

Should this go through QA, or do you think the refactor QA + our testing is good enough?

millerdev:
Excellent! It looks good. I think a QA run would be a good idea. We changed a few more pretty wide-reaching things in the parser since we QA'd [only-one-tree](https://github.com/dimagi/Vellum/pull/323). Also, there's [one more thing](https://github.com/dimagi/Vellum/pull/323#discussion-diff-24529244) to do on that branch, which could be done here just as easily.

Also, this is merging into only-one-tree, which is why we got all that activity from master. You'll probably want to create a new PR against master.

emord:
Will reopen based on master

PR #328, opened 2015-02-12
Limit type changes

orangejenny:
http://manage.dimagi.com/default.asp?67177

Left the alert alone to catch any programmatic attempts to change a select with children to a non-select.


src/core.js
```diff
@@ -1228,6 +1228,10 @@ define([
         }).on('parent-question-type-change', function (e) {
             _this.jstree("set_icon", e.childMug.ufid, e.childMug.getIcon());
         }).on('remove-question', function (e) {
+            var currentMug = _this.getCurrentlySelectedMug();
+            if (e.mug && e.mug.parentMug && e.mug.parentMug === currentMug) {
+                _this.displayMugProperties(currentMug);
+            }
```
millerdev: Was there a particular reason this was put at the beginning of this callback? Seems like `selectSomethingOrHideProperties` will possibly call `displayMugProperties` again. Would be good to not display properties more than once.

millerdev: Not only `selectSomethingOrHideProperties`, but anything that calls `jstree("select_node", ...)` will likely call `displayMugProperties`.

src/core.js
```diff
@@ -1242,6 +1246,10 @@ define([
             _this._resetMessages(e.errors);
         }).on('question-create', function (e) {
             _this.handleNewMug(e.mug, e.refMug, e.position);
+            var currentMug = _this.getCurrentlySelectedMug();
+            if (e.mug && e.mug.parentMug && e.mug.parentMug === currentMug) {
+                _this.displayMugProperties(currentMug);
+            }
             if (!e.isInternal) {
                 _this.jstree("deselect_all", true)
                      .jstree('select_node', e.mug.ufid);
```
millerdev: Same

src/core.js
```diff
@@ -1655,14 +1663,17 @@ define([
...
+                isSelect = /^M?Select$/.test.bind(/^M?Select$/);
 
             for (var i = 0; i < questions.length; i++) {
                 var typeName = questions[i],
                     q = _this.data.core.mugTypes[typeName];
                 if (q.isTypeChangeable && currentType !== typeName &&
                     (!q.limitTypeChangeTo || 
-                     q.limitTypeChangeTo.indexOf(currentType) !== -1))
+                     q.limitTypeChangeTo.indexOf(currentType) !== -1) &&
+                    (!hasChildren || !isSelect(mug.__className) || isSelect(typeName)))
```
millerdev: I know I was the one to refactor similar code over [here](https://github.com/dimagi/Vellum/blob/limit-type-changes/src/mugs.js#L1080), but since then I have started to delegate this kind of decision to the mug so we don't have to hard-code type names in places like this. This also makes it easier for plugins like itemsets to write custom mug types that need to have a say in this decision. What do you think of adding a new mug option like this:

```js
if (q.isTypeChangeable && currentType !== typeName &&
    (!q.limitTypeChangeTo ||
     q.limitTypeChangeTo.indexOf(currentType) !== -1) &&
    mug.options.canChangeType(mug, typeName))
{
    ...
}
```
`defaultOptions` in mugs.js will have this:
```js
/**
 * Decide if this mug can have its type changed to typeName
 *
 * The value of `limitTypeChangeTo` takes precedence over this option.
 *
 * @param mug - the mug that would change type.
 * @param typeName - the name of the new type for mug (e.g., MSelect).
 * @returns - true if the mug can change to type, otherwise false.
 */
canChangeType: function (mug, typeName) { return true; },
```

Then MSelect, and by inheritance Select also, can provide a custom value for this option that checks if the mug has children, etc.

orangejenny: I like that much better, will implement it.

src/core.js
orangejenny: This one's necessary because when the two child options are auto-added to a new select, that should cause the type dropdown to change, but those events are internal, so the select_node doesn't get triggered.

src/core.js
orangejenny: This isn't necessary, I did it for symmetry with the code below in case we ever programmatically remove children. I'm fine with removing it.

orangejenny: ...actually, the one place we programmatically remove children is in the test I added in this PR. Leaving this code in here unless you strongly object.

orangejenny:
@millerdev 

Addressed comments. Implemented canChangeType as a function that returns an error message on failure. It's a little non-standard, but it lets me move the actual error message into mugs.js. Also eliminated limitTypeChangeTo, because having three different ways to specify type changing seemed like a bit much.

src/core.js
millerdev: It probably won't matter, but we should be mindful of this if we start to notice the properties flashing or other general sluggishness on creating/moving/removing questions.

millerdev:
Looks good. Waiting for build.

PR #330, opened 2015-02-12
Add help placeholder

emord:
http://manage.dimagi.com/default.asp?156218

selection_004

### DIFF
--- a/src/javaRosa.js
+++ b/src/javaRosa.js
@@ -992,6 +992,7 @@ define([
         widget.isDefaultLang = widget.language === widget.defaultLang;
         widget.isSyncedWithDefaultLang = false;
         widget.hasNodeIdPlaceholder = options.path === 'labelItext';
+        widget.customPlaceholder = options.customPlaceholder;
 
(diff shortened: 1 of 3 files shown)


emord:
There wasn't really another example of this type of help text in vellum. The other help texts were on the actual type of this question, where this was a help for one input in a question

orangejenny:
This is kind of weird. Can we chat about it?

orangejenny:
To be more articulate: I don't think that placeholders are a good place for information like this. Would much prefer we pick a standard for marking android-only functionality, which pops up all over HQ. Maybe the robot icon that we currently use for callout questions, with a hover that says "Android only"?

amsagoff:
Agree with Jenny - though also this is under "advanced" so we really dont
have to put too much effort into the UX. Fine with removing any indication
it's android only if this is going to be an effort.

emord:
I think this will be more effort to come up with a new standard way to do this than we would normally put in a qa bug ticket. I'll make a new ticket for coming up with something like this though.

PR #331, opened 2015-02-17
Change save button if id changes

emord:
http://manage.dimagi.com/default.asp?156179#875788

Tried to write out a test for this, but for some reason the save button wouldn't be triggered in the test. (tried both $("[name='property-helpItextID']").val('blah') and text.p.helpItextID.id = 'blah')

### DIFF
--- a/src/javaRosa.js
+++ b/src/javaRosa.js
@@ -576,7 +576,12 @@ define([
 
         widget.save = function () {
             // override save to call out to rename itext
-            widget.mug.setItextId(widget.path, widget.getValue());
+            if (widget.mug.setItextId(widget.path, widget.getValue())) {
+                mug.form.fire({
+                    type: 'change',
+                    mug: mug
+                });
+            }
         };
 
         widget.mug.on('property-changed', function (e) {
--- a/src/javaRosa.js
+++ b/src/javaRosa.js
@@ -576,12 +576,7 @@ define([
 
         widget.save = function () {
             // override save to call out to rename itext
-            if (widget.mug.setItextId(widget.path, widget.getValue())) {
-                mug.form.fire({
-                    type: 'change',
-                    mug: mug
-                });
-            }
+            widget.mug.setItextId(widget.path, widget.getValue());
         };
 
         widget.mug.on('property-changed', function (e) {
--- a/src/mugs.js
+++ b/src/mugs.js
@@ -608,7 +608,8 @@ define([
             return rootId + nodeId + "-" + itextType;
         },
         setItextId: function (propertyPath, id, unlink) {
-            var itext = this.p[propertyPath];
+            var itext = this.p[propertyPath],
+                _this = this;
 
             if (id !== itext.id) {
                 if (unlink) {
@@ -620,9 +621,11 @@ define([
                 // Is this necessary, since itext is a reference?
                 // It probably triggers handlers.
                 this.p[propertyPath] = itext;
-                return true;
+                this.form.fire({
+                    type: 'change',
+                    mug: _this
+                });
             }
-            return false;
         },
         unlinkItext: function () {
             var _this = this;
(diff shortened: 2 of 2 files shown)


src/javaRosa.js
```diff
@@ -576,7 +576,12 @@ define([
 
         widget.save = function () {
             // override save to call out to rename itext
-            widget.mug.setItextId(widget.path, widget.getValue());
+            if (widget.mug.setItextId(widget.path, widget.getValue())) {
```
emord: Is it kosher in vellum to test the return value of a function in the if statement like this, or is that a dirty C habit I've picked?

src/mugs.js
```diff
@@ -620,7 +620,9 @@ define([
                 // Is this necessary, since itext is a reference?
                 // It probably triggers handlers.
                 this.p[propertyPath] = itext;
+                return true;
```
emord: true if changed. false otherwise.

orangejenny:
On the test, do you need to fire something like `_this.data.core.saveButton.fire('change')` or maybe explicitly blur `$("[name='property-helpItextID']")`?

src/javaRosa.js
millerdev: Coming from the Python world, this is generally discouraged. One reason being if someone adds another condition using `||` (or) or `&&` (and) it can cause unexpected things to happen due to short-circuit logical evaluation. For example (silly example):
```js
if (true && widget.mug.setItextId(widget.path, widget.getValue())) {
    // we're here, but setItextId did't get called
}
```
My rule of thumb is if you really want that function to be called then you should assign it's return value to a variable and test the value of that. If it doesn't matter if it gets called, then testing it in a condition is probably fine.

millerdev: Do you know who is calling this `save()` method? Typically [`widget.handleChange()` -> `updateValue()` -> `save()`](https://github.com/dimagi/Vellum/blob/save-button-fix/src/widgets.js#L61-L87), and somewhere in there the save button gets activated for most widgets. Do you know why the normal save button activation is not working for this widget? It's possible that this widget does not implement the widget protocol correctly, and you may be papering over a bigger problem with this fix.

millerdev: The answer came to me after asking all those questions. Mug properties usually fire change events when they are set, however `setItextId` is not doing that. It may make sense to simply have it fire that event unconditionally, or maybe if there are some places (for example during parsing) where change events should not be fired, then maybe you'll want to add a third parameter where you can indicate whether it should fire a change event or not. In this case it should. With the current design of Vellum it seems like it is not the widget's job to fire internal mug change events.

emord: Good point. I moved it to the mug. No need for having a function call in a if statement now

src/mugs.js
```diff
@@ -620,6 +621,10 @@ define([
                 // Is this necessary, since itext is a reference?
                 // It probably triggers handlers.
                 this.p[propertyPath] = itext;
+                this.form.fire({
+                    type: 'change',
+                    mug: _this
+                });
```
millerdev: I think this will fire a duplicate change event if `unlink` is true ([mugs.js:79-100](https://github.com/dimagi/Vellum/blob/save-button-fix/src/mugs.js#L79-L100)), and omit other parts of the typical change process (the property change logic is kind of a complicated/messy: [form.js:661](https://github.com/dimagi/Vellum/blob/save-button-fix/src/form.js#L661)) if `unlink` is false. What do you think of this implementation (replaces lines 618-627):
```js
} else if (itext.id !== id) {
    // HACK ensure that the property really changes so change handlers fire.
    this.__data[propertyPath] = null;
}
itext.id = id;
this.p[propertyPath] = itext;
```

millerdev: I edited the suggested code above after the initial submission.

PR #332, opened 2015-02-18
Make XPath widget properly handle newlines

millerdev:
http://manage.dimagi.com/default.asp?157728

This bug was introduced by https://github.com/dimagi/Vellum/commit/1cf6eb559f9937b6705b594ce2fa0f99c1bf3966

### DIFF
--- a/tests/widgets.js
+++ b/tests/widgets.js
@@ -1,0 +1,93 @@
+require([
+    'chai',
+    'jquery',
+    'underscore',
+    'tests/utils',
+    'vellum/expressionEditor',
+    'vellum/util'
+], function (
+    chai,
+    $,
+    _,
+    util,
+    expressionEditor,
+    vellumUtil
+) {
+    var assert = chai.assert,
+        real_showXPathEditor = expressionEditor.showXPathEditor,
+        events = {};
+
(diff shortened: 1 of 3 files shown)


tests/widgets.js
```diff
@@ -0,0 +1,93 @@
...
+                escaped = '"line 1 &#10; line 2"';
+            util.addQuestion("DataBindOnly", "hidden", {
+                calculateAttr: value
+            });
+            util.clickQuestion("/data/hidden");
+            var input = $("[name=property-calculateAttr]");
+            assert.equal(input.val(), escaped);
+
+            // click Edit button
+            input.closest("div").siblings(".fd-edit-button").click();
```
orangejenny: Picky, but could this be `input.closest(".control-group").find(".fd-edit-button")`?  I think both looking for a generic ancestor div, and using siblings, are kind of fragile.

millerdev: Haha! Yes, it's very fragile. The whole test stinks, but this issue has bitten me so many times in the past that I feel like the cost/fragility is worth it. Happy for suggestions on how to improve it. Thanks!

snopoke:
:+1: for tests

PR #333, opened 2015-02-18
Commtrack questions

orangejenny:


src/commtrack.js
```diff
@@ -0,0 +1,377 @@
...
+                requiredAttr: { presence: "notallowed" },
+                constraintAttr: { presence : "notallowed" },
+                calculateAttr: { presence: "notallowed" }
+            }
+        };
+
+    $.vellum.plugin("commtrack", {}, {
+        getAdvancedQuestions: function () {
+            var questions = this.__callOld();
+            if (this.opts().features.transaction_question_types) {
```
orangejenny: I feel like this is probably not the best place to check the feature flag - like it should be somewhere much higher up - but wasn't successful. Open to suggestions.

millerdev: Since the entire `commtrack` plugin contains only these two questions for now, what do you think of removing all references to this feature flag from Vellum, and instead use the feature flag in HQ (somewhere around [here](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/app_manager/views.py#L1152)) to enable/disable the plugin entirely?

tests/options.js
```diff
@@ -107,10 +107,11 @@ define(function () {
             },
             objectMap: {}  // todo
         },
-        plugins: ['itemset', 'modeliteration'],
+        plugins: ['itemset', 'modeliteration', 'commtrack'],
         features: {
             'add_help_text': true,
-            'group_in_field_list': true
+            'group_in_field_list': true,
+            'transaction_question_types': true,
```
millerdev: This can be removed

millerdev:
Ran tests locally, and they passed. Not waiting for Travis.

czue:
:trophy:

czue:
can we get this on HQ tonight?

millerdev:
yep, it's on HQ master now, so will go out with the deploy

czue:
awesome! thanks @millerdev and @orangejenny!

orangejenny:
minor FYI: the balance icon will be missing until the next HQ deploy, that PR didn't make it in time today

PR #334, opened 2015-02-20
Fix bugs in data source instance tracking

millerdev:
@czue

### DIFF
--- a/src/modeliteration.js
+++ b/src/modeliteration.js
@@ -284,24 +284,20 @@ define([
     }
 
     function updateDataSource(mug, value, previous) {
-        var value_src = value && value.instance && value.instance.src,
-            prev_src = previous && previous.instance && previous.instance.src;
-        if (value_src !== prev_src) {
-            var ref = mug.ufid + ".dataSource.instance";
-            if (prev_src) {
-                mug.form.dropInstanceReference(prev_src, ref);
-            }
-            if (value_src) {
-                var instanceId = mug.form.addInstanceIfNotExists(value.instance, ref);
-                if (instanceId !== value.instance.id) {
-                    // is it too magical to replace the instance id in the query?
-                    // there might be edge cases where a user is entering a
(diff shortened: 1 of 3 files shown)


czue:
not sure i feel totally qualified to review this - maybe @emord or @orangejenny can, or you can put some more information about how this change fixes the issue?

millerdev:
Previously the instance name in the UI and IDs query could get out of sync with the instance on the form. This change does the following:

- Properly update the instance id and query with the existing instance ID when the instance ID is changed (but not the src) while editing a model iteration data source and there is at least one other question on form referencing the same instance.
- Rename the instance ID to the new value when the instance ID is changed (but not the src) while editing a model iteration data source and there are no other questions referencing the instance.
- Also improve the data source widget. This is not due to a bug really, but rather just a better/more internally consistent implementation.

czue:
thanks. looks good to me but will give @emord or @orangejenny some time to review before merging.

src/modeliteration.js
```diff
@@ -284,24 +284,20 @@ define([
...
-            if (value_src) {
-                var instanceId = mug.form.addInstanceIfNotExists(value.instance, ref);
-                if (instanceId !== value.instance.id) {
-                    // is it too magical to replace the instance id in the query?
-                    // there might be edge cases where a user is entering a
-                    // custom instance and query and does not want this to
-                    // happen
-                    value.idsQuery = value.idsQuery.replace(
-                        instanceRegexp, "instance('" + instanceId + "')");
-                    mug.p.dataSource = value; // fire change handler
```
emord: Is firing the change handler no longer needed?

millerdev: Correct, this code is triggered by/part of a change handler already. Also, I think this never caused anything to happen anyway because the change handler only fires if the value actually changes (`value !== old`), which I'm pretty sure is never true.

I also manually tested this to verify that the UI acts the way we expect.

PR #335, opened 2015-02-20
Repeat count reference

emord:
http://manage.dimagi.com/default.asp?158002#878245


millerdev:
nice!

millerdev:
@emord I think this caused an issue with model iteration. I now get a warning when I configure a Repeat Group for model iteration:

screenshot 2015-02-23 17 10 23

PR #336, opened 2015-02-23
Warn on same choice value

emord:
http://manage.dimagi.com/default.asp?116075


millerdev:
Odd that tests are failing. Doesn't look related, but I can't figure out why it would be happening here and not on other PRs like [itemsets](https://github.com/dimagi/Vellum/pull/337).

Note that this is the second time the "push" tests failed on this PR. I had noticed they failed earlier and kicked them, but ended with the same result.

PR #337, opened 2015-02-23
Itemsets

millerdev:
- Use the same data source widget as for model iteration.
- Implement proper instance ref counting for itemsets.
- Clean up existing- and add some new tests.
- Do not allow appearance attribute on (control-only) itemset mug.


src/core.js
```diff
@@ -2046,7 +2045,6 @@ define([
         saveType: 'full',
         staticPrefix: "",
         allowedDataNodeReferences: [],
-        externalInstances: [],
```
millerdev: No longer used now that itemsets implement proper instance ref counting.

millerdev:
Changed my mind on QA for this. The core bits it touches are pretty small and not heavily used (instance ref counting), and I think they are well tested by the automated test suite. The itemsets plugin has been on prod (disabled) for ages, so that's not likely to introduce anything new/unexpected. I'm planning add a feature flag in HQ so QA/UAT can be done on prod, but this can be merged before that.

PR #340, opened 2015-02-24
Non group child2

emord:
@millerdev 

Had to add a couple of commits at the end to fix tests, but mostly just rebased on master

### DIFF
--- a/src/mugs.js
+++ b/src/mugs.js
@@ -524,11 +524,6 @@ define([
             } else {
                 var path = this.form.getAbsolutePath(this, true);
                 if (!path) {
-                    // fall back to control tree if mug not in data tree
-                    // this can happen with malformed XForms
-                    path = this.form.getControlPath(this, true);
-                }
-                if (!path) {
                     // fall back to nodeID if mug path still not found
                     // this can happen with malformed XForms
                     path = "/" + this.getNodeID();
--- a/src/mugs.js
+++ b/src/mugs.js
@@ -302,7 +302,8 @@ define([
                     function recFunc(mug) {
                         if (!mug) {
                             return true;
-                        } else if (mug.__className === 'Repeat') {
+                        } else if (mug.__className === 'Repeat' ||
+                                   mug.__className === "Item") {
                             return false;
                         }
                         return recFunc(mug.form.tree.getNodeFromMug(mug).parent.value);
--- a/src/mugs.js
+++ b/src/mugs.js
@@ -307,9 +307,6 @@ define([
                         }
                         return recFunc(mug.parentMug);
                     }
-                    if (mug.__className === "Item") {
-                        return false;
-                    }
 
                     return recFunc(mug.parentMug);
                 },
--- a/src/mugs.js
+++ b/src/mugs.js
@@ -305,7 +305,7 @@ define([
                         } else if (mug.__className === 'Repeat') {
                             return false;
                         }
-                        return recFunc(mug.form.controlTree.getNodeFromMug(mug).parent.value);
+                        return recFunc(mug.form.tree.getNodeFromMug(mug).parent.value);
                     }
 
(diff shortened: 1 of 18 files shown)


src/commtrack.js
```diff
@@ -105,6 +105,15 @@ define([
             getPathName: function (mug, name) {
                 return mug.options.getTagName() + "[@type='" + name + "']";
             },
+            adjustPath: function (mug, path) {
```
millerdev: `adjustPath` was renamed to `getPathName`

src/tree.js
```diff
@@ -139,6 +141,28 @@ define([
...
+                parentPath = mug.p.dataParent;
+            } else {
+                parentPath = this.parent.getAbsolutePath(excludeRoot);
+                if (parentPath === null) {
+                    return null;
+                }
+            }
+            var path = parentPath + '/' + this.getID();
+            if (mug.options.adjustPath) {
+                path = mug.options.adjustPath(mug, path);
```
millerdev: I suspect the rebase clobbered something or maybe omitted something. This should use `getPathName` instead of `adjustPath`.

src/modeliteration.js
```diff
@@ -110,6 +110,12 @@ define([
                 }
                 return binds;
             },
+            adjustPath: function (mug, path) {
```
millerdev: see other notes on `adjustPath`

src/commtrack.js
emord: Fixed that bad rebase

emord:
@millerdev passed QA http://manage.dimagi.com/default.asp?158079

I think there were a couple of commits to address QA bugs since you last looked at this, so may want to take a pass on the past 3 commits

src/mugs.js
```diff
@@ -295,7 +295,45 @@ define([
...
-            }
+            },
+            dataParent: {
+                lstring: 'Data Parent',
+                visibility: function(mug) {
+                    function recFunc(mug) {
+                        if (!mug) {
+                            return true;
+                        } else if (mug.__className === 'Repeat' ||
+                                   mug.__className === "Item") {
```
millerdev: Could this be changed to `mug.paretMug && !mug.parentMug.options.possibleDataParent`

Is this a correct statement?
A repeat node can have a data parent, but a repeat cannot itself be a data parent.

Hmm, is this code enforcing that?

src/mugs.js
```diff
@@ -295,7 +295,45 @@ define([
...
+                            return false;
+                        }
+                        return recFunc(mug.form.tree.getNodeFromMug(mug).parent.value);
```
millerdev: If you're changing the code above, you could change this to `recFunc(mug.parentMug)` since `mug.parentMug` is now a calculated property that does what this is doing.

src/mugs.js
emord: A repeat node can have a different data parent, but cannot be a data parent. Also children of the repeat node cannot have different data parents. This is checked in the validation function. There are also two tests:  

https://github.com/dimagi/Vellum/pull/340/files#diff-597c6d5c9e1efe9adf37cd233ade5331R32

https://github.com/dimagi/Vellum/pull/340/files#diff-597c6d5c9e1efe9adf37cd233ade5331R80

src/mugs.js
```diff
@@ -295,7 +295,47 @@ define([
...
+                visibility: function(mug) {
+                    function recFunc(mug) {
+                        if (!mug) {
+                            return true;
+                        } else if (!mug.options.possibleDataParent) {
+                            return false;
+                        }
+                        return recFunc(mug.parentMug);
+                    }
+                    if (mug.__className === "Item") {
```
millerdev: An "Item" can never be a parent mug, and it's parent will never be a possible data parent, so this check seems unnecessary. Correct?

PR #341, opened 2015-02-25
Repeat count reference

emord:
https://github.com/dimagi/Vellum/pull/335#issuecomment-75648494

### DIFF
--- a/src/modeliteration.js
+++ b/src/modeliteration.js
@@ -131,7 +131,9 @@ define([
                     widget: idsQueryDataSourceWidget,
                 }
             },
-            noLogicReference: true
+            ignoreReferenceWarning: function(mug) {
+                return isModelRepeat(mug);
+            }
         };
 
     $.vellum.plugin("modeliteration", {}, {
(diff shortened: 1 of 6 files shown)


src/modeliteration.js
```diff
@@ -130,7 +130,8 @@ define([
                     presence: 'optional',
                     widget: idsQueryDataSourceWidget,
                 }
-            }
+            },
+            noLogicReference: true
```
emord: I think this should be temporary until model iteration gets a nice user friendly UI. Then maybe we could extend the logic expression to include the cases/fixtures that it will pull from.

millerdev: When the `modeliteration` plugin is enabled this `modelRepeatMugOptions` object applies to all Repeat questions. How does this not disable `repeat_count` validation on all Repeat questions?

I see you have added tests to verify that it is/is not working at the correct times, but I can't figure out why they're passing. ... ok, on further inspection I think I get it. This disables the validation error message (for all Repeat questions), but it does not disable the automatic reference updates. So, unlike other mugs, there will be no validation error if you delete the node that a regular Repeat question's `repeat_count` references. If you wanted to fix that issue you could make `noLogicReference` a function instead of a constant. Something like this:
```js
noLogicReference: function (mug) {
    return !mug.p.dataSource.idsQuery;
}
```

PR #343, opened 2015-02-26
Improve XML normalization

millerdev:
http://manage.dimagi.com/default.asp?158705

### DIFF
--- a/tests/xml.js
+++ b/tests/xml.js
@@ -1,17 +1,23 @@
 require([
     'jquery',
+    'underscore',
     'chai',
     'vellum/xml'
 ], function (
     $,
+    _,
     chai,
     xml
 ) {
     var assert = chai.assert;
 
     describe("The XML normalizer", function () {
-        function eq(value, escaped) {
-            assert.strictEqual(xml.normalize(value), escaped);
+        function eq(value, escaped, humanize) {
+            var normal = xml.normalize(value);
+            assert.strictEqual(normal, escaped);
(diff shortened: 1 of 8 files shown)


tests/xml.js
```diff
@@ -0,0 +1,128 @@
...
+        });
+
+        it("should escape > character", function () {
+            eq('your visit count must be > 5', 'your visit count must be &gt; 5');
+        });
+
+        it("should escape < character before tag", function () {
+            // immortalizing https://github.com/dimagi/Vellum/pull/212
+            eq('your visit count must be < <output value="/path" />',
+               'your visit count must be &lt; <output value="/path" />');
```
emord: Not sure I understand how this one figures out not to convert the second < but I like the fact that we're testing that

tests/xml.js
```diff
@@ -0,0 +1,128 @@
...
+
+        it("should accept jquery node", function () {
+            eq($('<output />'), '<output />');
+        });
+
+        /* -- The following are lossy -- */
+
+        it("should handle malformed tag", function () {
+            // big loss here
+            eq('<output value="/path"', ''); //'<output value="/path" />');
```
emord: Is this commented, because we want to eventually make it to return the comment?

I noticed you had this comment on the other linked PR.

millerdev: Removed comment. No plans to change this since it's a result of the "pure magic" part of the serialization process.

PR #344, opened 2015-02-27
Fix transfer src/dest attribute writing

millerdev:
Remove pre-existing src/dest attribute from the data node when value its removed.
http://manage.dimagi.com/default.asp?158056

@czue @emord


czue:
:+1: from me - thanks for the fast fix!

PR #345, opened 2015-02-27
Enable save

orangejenny:
http://manage.dimagi.com/default.asp?158065

getValue was modifying value, so it got updated prematurely and the later test that value doesn't equal previous  was always failing.

### DIFF
--- a/src/commtrack.js
+++ b/src/commtrack.js
@@ -379,7 +379,7 @@ define([
         };
 
         widget.getValue = function () {
-            var val = _value || {};
+            var val = _.clone(_value) || {};
             val.value = _getValue();
             return val;
         };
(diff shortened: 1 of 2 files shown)


czue:
thanks!

millerdev:
Heh, I thought @orangejenny had handed this one off to me, my bad for not reassigning the bug to myself. I had started working on [a different fix for this issue](https://github.com/dimagi/Vellum/compare/activate-save-button-on-widget-change), but I'm not finished with it yet.

While simple, the problem with this fix is that it makes an unnecessary clone of the value that does not reflect a real change. I can't remember exactly how, but I had decided this approach would cause unnecessary save button activations under some circumstances.

Instead of triggering save button activation based on the mug property value inequality, I am planning to trigger it on the widget change event. There are a few edge cases I still need to verify before my work is complete.

orangejenny:
@millerdev 

Hey, sorry Daniel, I did hand it off to you, then had some time Friday afternoon and started looking at it again. Sounds like you have a more well-thought-out fix.

PR #349, opened 2015-03-04
Escape mug display name

emord:
Escaping the mugs display name 

The fun things you can do without this:

out

out-1

### DIFF
--- a/src/mugs.js
+++ b/src/mugs.js
@@ -613,7 +613,7 @@ define([
                 if (lang !== defaultLang && disp === defaultDisp) {
                     disp += " [" + defaultLang + "]";
                 }
-                return disp;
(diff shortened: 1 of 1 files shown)


millerdev:
How did you stumble upon this?

emord:
There was a bug for for <output value> in labels. I noticed it was blank in the tree. Inspected the element and saw that they were just being inserted as html.

So naturally I defaced HQ with marquee rainbows. Also I found out that Chrome disabled blink tags.

Also inserting javascript does nothing.

emord:
Actually I take that back. Arbitrary javascript can be run on form load

millerdev:
> So naturally I defaced HQ with marquee rainbows.

quote of the week!

PR #351, opened 2015-03-05
Enable save button on bulk update translations

millerdev:


src/javaRosa.js
```diff
@@ -1520,6 +1521,8 @@ define([
             this._makeLanguageSelectorDropdown();
 
             this.__callOld();
+
+            Itext.on('change', function () { _this.onFormChange(); });
```
millerdev: Kinda sad that this had to come back.

PR #352, opened 2015-03-05
Diff on overwrite

emord:
@millerdev 

Making this diff pretty with green/red colors will probably be a PITA, but this has all the info one would need right now

selection_012

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -1,14 +1,12 @@
 define([
     'json!langCodes',
     'underscore',
-    'equivalent-xml',
     'jsdiff',
     'jquery',
     'jquery.bootstrap-popout'
 ], function (
     langCodes,
     _,
-    EquivalentXml,
     jsdiff,
     $
 ) {
--- a/src/util.js
+++ b/src/util.js
@@ -1,11 +1,15 @@
 define([
     'json!langCodes',
     'underscore',
+    'equivalent-xml',
+    'jsdiff',
     'jquery',
     'jquery.bootstrap-popout'
 ], function (
     langCodes,
     _,
+    EquivalentXml,
+    jsdiff,
     $
 ) {
     RegExp.escape = function(s) {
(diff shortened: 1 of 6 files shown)


emord:
The dependencies increases vellums size on prod about 10K after minification

src/util.js
```diff
@@ -268,6 +272,33 @@ define([
         pos = pos + text.length;
         that.setCaretPosition(ctrl, start, pos);
     };
+
+
+    that.xmlDiff = function (localForm, serverForm, opts) {
```
emord: slimmed down version of https://github.com/dimagi/Vellum/blob/master/tests/utils.js#L28

Slimmed it down because I don't think color will work right now. Would have to do a bit of work or use some pretty diff coloring library.

src/util.js
```diff
@@ -268,6 +272,33 @@ define([
...
+        }
+        opts = opts || {};
+        if (opts.normalize_xmlns) {
+            var xmlns = $($.parseXML(serverForm)).find('data').attr('xmlns');
+            localForm = localForm.replace(/(data[^>]+xmlns=")(.+?)"/,
+                                    '$1' + xmlns + '"');
+        }
+        localForm = cleanForDiff(localForm);
+        serverForm = cleanForDiff(serverForm);
+        var patch = jsdiff.createPatch("", localForm, serverForm, "Local Form", "Server Form");
```
millerdev: What do you think of swapping the order of these? Right now it renders as
```
--- Local Form
+++ Server Form
```
but *Local Form* will overwrite *Server Form*. Seems more intuitive to me to show `+` beside lines that will be kept and `-` beside lines that will be removed.

src/util.js
```diff
@@ -1,11 +1,15 @@
...
     'json!langCodes',
     'underscore',
+    'equivalent-xml',
+    'jsdiff',
     'jquery',
     'jquery.bootstrap-popout'
 ], function (
     langCodes,
     _,
+    EquivalentXml,
```
millerdev: This does not appear to be used. Do we need it?

millerdev:
Awesome!

millerdev:
It's a little off-topic for this PR, but while we're improving this dialog would it be easy to make Enter/Return keys overwrite and Escape cancel?

orangejenny:
Nice. Can the XML go behind a link? This is awesome for advanced users and is going to freak out non-advanced users.

millerdev:
Sorry for bike-shedding. Jenny's suggestion seems like a good idea. Mine is not that important, especially if it's hard.

emord:
@orangejenny @millerdev 

Here's how it looks when it comes up now:

selection_013

Info seemed like the best option here: http://getbootstrap.com/css/#buttons-options

Once pressed it still looks like the first screenshot except local and server are now switched and there's a disabled show xml button on the bottom.

millerdev:
Build failure looks real. Looks like the build is trying to include equivalent-xml, but can't find it.

emord:
Yeah it was real. It passed ./test but not the make

PR #354, opened 2015-03-09
Work around XML escaping bug in browser

millerdev:
http://manage.dimagi.com/default.asp?122203
http://manage.dimagi.com/default.asp?159671


czue:
nice tests

PR #360, opened 2015-03-20
Untangling javaRosa itext

millerdev:
I apologize for the size of this thing, but it really cleans up javaRosa itext, which was something that had been itching for a while. This makes serializing itext easier, and I'll be building my copy-paste work on top of it. Copy/paste support is well underway, but I decided to go back and do this independently so it did not pollute that PR.

Major new feature: automatic Itext IDs are now detected on reading XML and re-calculated just in time before the the form is serialized. This greatly simplifies handling of auto IDs.

Also, mug properties that used to end with the suffix `ItextID` have been renamed so they now end with the suffix `Itext` (there are still virtual properties that end with `ItextID` that are used to produce widgets to edit the corresponding itext ID). The naming was very confusing before this refactor. Properties that held an `ItextItem` were named `labelItextID` (for example) and had a widget to edit the ItextItem ID, and the corresponding virtual property (`labelItext`) was used to create widgets for editing the itext content. That madness is hopefully behind us now.

Another simple change that maybe should have it's own PR is moving `tagName` and `dataType` from mug properties to mug options. They are immutable, so it makes more sense to have them as options than as properties, and it make serializing mugs easier. It accidentally got into this PR because it's another copy/paste re-req.

There are a few remnants in the mugs.js that still do Itext-related stuff, and I plan to gradually move that over to javaRosa.js as I find it convenient.

As a bonus, this fixes http://manage.dimagi.com/default.asp?154742


src/core.js
```diff
@@ -1368,7 +1368,10 @@ define([
         this.ensureCurrentMugIsSaved(function () {
             var foo = _this.getInsertTargetAndPosition(
                 _this.getCurrentlySelectedMug(), qType);
-            mug = _this.data.core.form.createQuestion(foo[0], foo[1], qType);
+            if (!foo) {
+                throw new Error("cannot add " + qType + " at the current position");
+            }
```
millerdev: There is a test that tries to insert an Item (normally a child of a Select question) at the top level and asserts that an error is thrown. Prior to this refactor somewhere in javaRosa (I think) something tried to access a property of null, which happened to throw an error in that case. However, by the time the error was thrown the mug had been created and was at least partly inserted into the vellum core data structures, leaving the form in a dirty state. This cleans up that mess.

src/core.js
```diff
@@ -1409,7 +1421,41 @@ define([
...
+        }
+        return {mug: refMug, position: position};
+    };
+
+    /**
+     * Check if a question of the given `type` can be inserted at `position`
+     * relative to `refMug`
+     *
+     * WARNING the bare minimum has been implemented to support
+     * getInsertTargetAndPosition(). Needs to be fleshed out for other uses.
```
millerdev: This will be needed by and completed for copy/paste.

src/javaRosa.js
```diff
@@ -218,7 +224,6 @@ define([
         util.eventuality(this);
         
         this.languages = [];
-        this.items = [];
```
millerdev: The ItextModel no longer keeps a list of items. Itext items are now always accessed via mugs.

src/javaRosa.js
```diff
@@ -384,123 +282,137 @@ define([
...
+                    // ignore empty item with duplicate ID
+                    return;
                 }
-            }    
-        };
-        
-        form.tree.treeMap(appendItemsIfPresent);
+                while (byId.hasOwnProperty(id)) {
+                    id = origId + count;
+                }
```
millerdev: This used to be done by [`deduplicateIds`](https://github.com/dimagi/Vellum/pull/360/files#diff-e5b3eca8407b45b132e0147b8392192cL267)

src/javaRosa.js
```diff
@@ -1452,9 +1331,10 @@ define([
         loadXML: function (xmlString) {
             var _this = this,
                 langs = this.opts().javaRosa.langs,
-                Itext;
+                Itext, itextMap;
 
             this.data.javaRosa.Itext = Itext = new ItextModel();
+            this.data.javaRosa.itextMap = itextMap = {};
```
millerdev: This temporary mapping of itext IDs to itext items is constructed during XML parsing, and used below in `populateControlMug`, which is called by the parser in the super call. The mapping is discarded once the form has been parsed. One possible issue is that itext from the XForm that is not referenced by a mug is discarded early (just after parsing). Previously it did not get discarded until the form was saved (see code removed from [`beforeSerialize`](https://github.com/dimagi/Vellum/pull/360/files#diff-e5b3eca8407b45b132e0147b8392192cL1643)).

src/javaRosa.js
```diff
@@ -1523,8 +1407,80 @@ define([
...
+                return "";
+            }
+
+            function getItextItem(id, property) {
+                var auto = !id || id === getDefaultItextId(mug, property);
+                if (id) {
+                    var item = itextMap[id];
+                    if (item && itextMap.hasOwnProperty(id)) {
+                        if (!auto) {
+                            item.autoId = false;
```
millerdev: All itext items have `autoId` turned on by default. This turns it off selectively for any ID that differs from the default. This used to be done later (during XML writing) leaving lots opportunity for the mug path to get out of sync with the itext ID.

From this point forward we depend on the `autoId` property of the itext rather than matching it ID against the default itext id for the mug/property.

src/javaRosa.js
```diff
@@ -1523,8 +1407,80 @@ define([
...
+            if (helpEl.length && mug.spec.label.presence !== 'notallowed') {
+                mug.p.helpItext = parseItextRef(helpEl, "help");
+            }
+            if (mug.p.constraintMsgAttr) {
+                var id = getITextID(mug.p.constraintMsgAttr);
+                if (id) {
+                    mug.p.constraintMsgItext = getItextItem(id, "constraintMsg");
+                    mug.p.constraintMsgAttr = null;
+                }
+            }
```
millerdev: All of the above had been in parser.js, intermingled with non-itext label parsing.

src/javaRosa.js
```diff
@@ -1602,8 +1558,8 @@ define([
             // non-duplicates
 
             var Itext = this.data.javaRosa.Itext,
+                items = this.data.javaRosa.itextItemsFromBeforeSerialize,
```
millerdev: Hmm, maybe this should be deleted after XML writing is finished?

src/javaRosa.js
```diff
@@ -1725,76 +1680,54 @@ define([
...
                 lstring: "Question Itext ID",
-                widget: function (mug, options) {
-                    return iTextIDWidget(mug, $.extend(options, {
-                        displayName: "Add Other Content",
-                        itextType: "label",
-                        getItextByMug: function (mug) {
-                            return mug.p.labelItextID;
-                        },
-                        forms: ['long', 'short'],
-                        isCustomAllowed: true
```
millerdev: These were dead options. None of them were used by `iTextIDWidget` (the same applies below). The history of this code is ... tangled.

src/javaRosa.js
```diff
@@ -1976,10 +1912,18 @@ define([
...
-            $textarea.val(generateItextXLS(this, Itext));
+            $textarea.val(generateItextXLS(vellum, Itext));
 
             $modal.modal('show');
             $modal.one('shown', function () { $textarea.focus(); });
         }
     });
+
+    return {
+        // TODO push these out to all places that use mug.getDefault...
```
millerdev: this comment should have been deleted...

src/mugs.js
```diff
@@ -458,7 +441,7 @@ define([
                 "jr:preloadParams": mug.p.preloadParams
             };
             _.each(mug.p.rawBindAttributes, function (value, key) {
-                if (!attrs.hasOwnProperty(key)) {
+                if (!attrs.hasOwnProperty(key) || _.isUndefined(attrs[key])) {
```
millerdev: Dear JavaScript...

src/mugs.js
```diff
@@ -667,37 +628,21 @@ define([
...
-        },
         unlinkItext: function () {
             var _this = this;
             _.each([
-                "labelItextID",
-                "constraintMsgItextID",
-                "hintItextID"
+                "labelItext",
+                "constraintMsgItext",
+                "hintItext"
```
millerdev: Looks like we need to add `helpItext` here. And move it over to javaRosa.js too.

src/core.js
```diff
@@ -2030,6 +2076,19 @@ define([
     };
 
     /**
+     * Extension point for mug setup during control node parsing
+     *
+     * The mug has been inserted into the tree by the time this method
+     * is called.
```
millerdev: Establishing this invariant makes handling badly written XForms much easier: default itext IDs, which depend on the mug's absolute path, can now be calculated reliably.

src/parser.js
```diff
@@ -338,17 +253,41 @@ define([
...
+        var labelEl = $cEl.children('label'),
+            hintEl = $cEl.children('hint');
+        if (labelEl.length && mug.spec.label.presence !== 'notallowed') {
+            var labelVal = xml.humanize(labelEl);
+            if (labelVal) {
+                mug.p.label = labelVal;
+            }
+        }
+        if (hintEl.length && mug.spec.hintLabel.presence !== 'notallowed') {
+            mug.p.hintLabel = xml.humanize(hintEl);
```
millerdev: Non-itext label parsing is done here, and itext label parsing is done in the javaRosa plugin.

src/widgets.js
```diff
@@ -99,13 +99,13 @@ define([
     };
 
     var normal = function(mug, options) {
-        var path = options.path,
+        var path = options.widgetValuePath || options.path,
```
millerdev: This makes defining virtual properties for the purpose of displaying an alternate widget very simple.

tests/formdesigner.ignoreButRetain.js
```diff
@@ -2,11 +2,17 @@
 define([
     'tests/utils',
     'chai',
-    'jquery'
+    'jquery',
+    'text!static/ignoreButRetain/empty-parent.xml',
+    'text!static/ignoreButRetain/renamed.xml',
+    'text!static/ignoreButRetain/unrenamed.xml'
```
millerdev: No new tests in this file, just moved XML out to files. Makes debugging test failures much easier.

tests/javaRosa.js
```diff
@@ -438,23 +444,22 @@ require([
         });
 
         it("should bulk update multi-line translation", function () {
-            util.loadXML(TEXT_QUESTION_XML);
-            var jr = util.call("getData").javaRosa,
+            var vellum = util.loadXML(TEXT_QUESTION_XML),
```
millerdev: Notice `loadXML` now returns the vellum instance. A convenience.

tests/javaRosa.js
```diff
@@ -522,6 +526,40 @@ require([
...
+
+        it("should not auto-update itext ID when multiple questions point to auto-ish-id", function () {
+            util.loadXML(ITEXT_ITEM_NON_AUTO_ID_XML);
+            var north = util.getMug("north");
+            north.p.nodeID = "west";
+            var xml = call("createXML"),
+                $xml = $(xml);
+            assert.strictEqual($xml.find("text#north-label").length, 2,
+                               "wrong <text> node count\n" + xml);
+        });
```
millerdev: A few new tests to verify that the itext-auto-id bugs are fixed. Some of this code came from https://github.com/dimagi/Vellum/compare/154742-auto-itext-ids, a branch that I had started this javaRosa refactor a while ago.

czue:
"As a bonus, this fixes http://manage.dimagi.com/default.asp?154742"

lol. i'm gonna let @emord field this one.

src/javaRosa.js
```diff
@@ -1198,7 +1040,7 @@ define([
         };
     };
     
-    var parseXLSItext = function (str, Itext) {
+    var parseXLSItext = function (vellum, str, Itext) {
```
emord: Why not just pass the form here?

src/javaRosa.js
emord: This should only be an issue if they added the itext manually and were planning to add appropriate mugs later in vellum, correct?  Seems like a pretty small edge case to me

millerdev: Thinking about this more, I think my "one possible issue" above is not actually an issue because I think there was/is no way to associate a mug itext property with an existing (possibly unreferenced) itext item aside from manually editing the XML, particularly because of the way [deduplication of itext IDs](https://github.com/dimagi/Vellum/pull/360/files#r26888162) is done.

millerdev:
Responded to feedback. Going to have Christy do a limited QA run as a sanity check.

PR #361, opened 2015-03-24
Itemset connector

emord:
Parking this here for now to review the work done so far.

- [x] Break out itemset widget into three different widgets
- [x] add dropdown widget
- [x] cache data sources info so it doesn't murder HQ
- [ ] add autocomplete to columns
- [ ] add advanced editor


src/datasources.js
```diff
@@ -3,7 +3,8 @@ define([
     'underscore',
     'vellum/widgets',
     'tpl!vellum/templates/data_source_editor',
-    'tpl!vellum/templates/select_data_source'
+    'tpl!vellum/templates/select_data_source',
+    'vellum/core'
```
millerdev: why is this here? `core.js` depends on this module.

tests/options.js
```diff
@@ -106,7 +100,7 @@ define(function () {
             },
             objectMap: {}  // todo
         },
-        plugins: ['itemset', 'modeliteration', 'commtrack'],
+        plugins: ['itemset', 'modeliteration', 'commtrack', 'datasources'],
```
millerdev: `datasources` is a regular module, not a plugin

src/datasources.js
```diff
@@ -225,11 +246,50 @@ define([
...
+            super_setValue(val.src || "");
+        }
+
+        widget.getValue = local_getValue;
+        widget.setValue = local_setValue;
+
+        return widget;
+    }
+
+    $.vellum.plugin("datasources", { }, {
```
millerdev: ah, I see you've made it a plugin

millerdev: We probably want to change the `init` function so it's called the same way as other plugins rather than the one-off way it's done by core.js right now.

src/itemset.js
```diff
@@ -201,30 +227,21 @@ define([
...
-        labelRef.onChange(handleChange);
-        valueRef.onChange(handleChange);
+            super_setValue = widget.setValue;
 
         widget.getUIElement = function () {
-            return super_getUIElement()
-                .append(labelRef.element)
-                .append(valueRef.element);
+            return super_getUIElement();
         };
```
millerdev: seems like this function is no longer necessary since it's a straight pass-through now

src/itemset.js
```diff
@@ -201,30 +227,21 @@ define([
...
-            var val = super_getValue();
+            var val = super_getValue(),
+                id = val.src.split(':')[val.src.split(':').length-1];
             return {
-                instance: ($.trim(val.src) ? {id: val.id, src: val.src} : null),
-                nodeset: val.query,
-                labelRef: labelRef.val(),
-                valueRef: valueRef.val()
+                instance: ($.trim(val.src) ? {id: id, src: val.src} : null),
+                nodeset: "instance('" + id + "')",
```
millerdev: isn't there often more to the nodeset? something like `instance('products')/products/product`. maybe the `/products/product` part get added somewhere I haven't seen yet...?

tests/options.js
```diff
@@ -38,36 +38,30 @@ define(function () {
...
-                            selector: "@bar_type='eggs'"
-                        }
-                    ]
-                }
+            sourceUri: "jr://fixture/item-list:some-fixture",
+            columns: [
+                'column1',
+                'column2',
+                'column3'
+            ]
```
millerdev: Is this what a reasonable fixture description will look like, or will this be fleshed out more later? I think we'll need to be able to model somewhat arbitrary XML structure eventually. Something like
```js
{
  sourceUri: "jr://fixture/commtrack:products",
  structure: { // the "structure" key is always present
    products: {
      product: {
        "@id": {},  // attribute
        name: {},   // child element with no attributes and no child nodes
        price: {},
        etc: {
          could: {have: {}},
          more: {structure: {"@in": {}, "@here": {}}}
        }
      }
    }
  }
}
```

Of course there are still XML documents that cannot be modeled with this convention, but I'm not sure we care.

src/itemset.js
emord: Kept this here as it will be changed when I add the advanced editor

src/itemset.js
emord: Yeah. I realized that last night and after looking through a couple of the tests

tests/options.js
emord: Yeah. this is why I decided to upload my progress here. This isn't done by any means. Just to make sure that I'm going along the right path

PR #364, opened 2015-04-01
track add-a-question, click-save actions

sravfeyn:


src/core.js
```diff
@@ -155,6 +155,9 @@ define([
                 }
                 _this.ensureCurrentMugIsSaved(function () {
                     _this.validateAndSaveXForm(forceFullSave);
+                    if (typeof window.analytics !== "undefined") {
```
emord: Should probably use !_.isUndefined(window.analytics)

src/core.js
```diff
@@ -1406,6 +1409,10 @@ define([
                 $firstInput.focus().select();
             }
         });
+
+        if (typeof window.analytics !== "undefined") {
+            window.analytics.track("Added a question");
+        }
```
millerdev: This should be moved up before line 1406

src/core.js
```diff
@@ -155,6 +155,9 @@ define([
                 }
                 _this.ensureCurrentMugIsSaved(function () {
                     _this.validateAndSaveXForm(forceFullSave);
+                    if (typeof window.analytics !== "undefined") {
+                        window.analytics.track("Clicked Save in Formbuilder");
+                    }
```
millerdev: I would move this above the call to `_this.validateAndSaveXForm` in case it raised an exception.

emord:
Just ran into this: https://github.com/dimagi/Vellum/blob/b06a065f06160e11427e2368047158713460989a/src/widgets.js#L219

Should that be changed to use windows.analytics.track?

sravfeyn:
Those are for Google Analytics this PR is related to Segment.com analytics.

millerdev:
I'll push a new version of Vellum to HQ today so this gets deployed.

millerdev:
pushed to HQ. should go live on the next deploy

sravfeyn:
Thanks Daniel!

PR #365, opened 2015-04-01
Itemset connector 2

emord:
Looking over the spec I just realized that the filter doesn't have the nice edit UI, but figured I could go ahead and open this up as it's feature flagged anyways.

selection_020

tooltip_021

The history is a little messy, so commit-by-commit may not be the best way to go through this one.


src/datasources.js
```diff
@@ -31,7 +102,8 @@ define([
                     dataType: 'json',
                     success: function (data) { callback(data); },
                     // TODO error handling
-                    data: {}
+                    data: {},
+                    async: false
```
emord: Made this change so that when itemset widget is loaded it will wait on the data source to be loaded

src/itemset.js
```diff
@@ -203,30 +211,51 @@ define([
...
             super_getUIElement = widget.getUIElement,
             super_getValue = widget.getValue,
             super_setValue = widget.setValue,
-            handleChange = widget.handleChange.bind(widget),
-            labelRef = refSelect("label_ref", "Choice Label", widget.isDisabled()),
-            valueRef = refSelect("value_ref", "Choice Value", widget.isDisabled());
+            super_handleChange = widget.handleChange.bind(widget),
+            labelRef = refSelect("label_ref", "Choice Label", false),
+            valueRef = refSelect("value_ref", "Choice Value", false),
+            filterRef = refSelect("filter_ref", "Choice Filter", false);
```
emord: Will need to change this over to a proper widget to get the edit button (I think)

src/util.js
```diff
@@ -291,7 +291,9 @@ define([
                                     '$1' + xmlns + '"');
         }
         localForm = cleanForDiff(localForm);
-        serverForm = cleanForDiff(serverForm);
+        if (serverForm) {
```
emord: This was an annoyance that the overwrite warning was failing on standalone mode because serverForm was null. Not really related to this PR

amsagoff:
Would be good to have the wording/section headers match the mockup too, unless that was an intentional divergence

src/datasources.js
```diff
@@ -11,14 +11,85 @@ define([
...
     function init(instance) {
         vellum = instance;
         dataSources = vellum.opts().core.dataSources || [];
+        cachedDataSources = {
+            fixture: {}
+        };
+    }
+
+    function cacheFixtures(data) {
+        _.map(data, function(fixture) {
```
millerdev: Should this function return a value? If not, might make more sense to use `_.each`

src/datasources.js
```diff
@@ -11,14 +11,85 @@ define([
...
+                return [key].concat(_.map(generateColumns(value.structure), function(value) {
+                    return key + '/' + value;
+                }));
+            });
+        }
+
+        if (fixture) {
+            return _.flatten(generateColumns(fixture.structure));
+        }
+        return "";
```
millerdev: Would it make sense to return an empty array here rather than a string since the normal case is to return an array?

src/datasources.js
```diff
@@ -225,11 +221,59 @@ define([
         return widget;
     }
 
+    function fixtureWidget(mug, options, labelText) {
```
millerdev: I think this widget will eventually be used for data sources other than fixtures as well. Would it make sense to name it something like `dataSourceWidget` and rename the existing widget with that name to `advancedDataSourceWidget`?

emord: Yeah. I was going to do something like that when I generalized everything to beyond fixtures. I thought may be good to keep it as fixtureWidget until then

src/itemset.js
```diff
@@ -203,30 +211,51 @@ define([
...
-                .append(valueRef.element);
+                .append(valueRef.element)
+                .append(filterRef.element);
         };
 
         widget.getValue = function () {
-            var val = super_getValue();
+            var val = super_getValue(),
+                s = val.query.search(END_FILTER),
+                filter = s === -1 ? '' : val.query.slice(s);
```
millerdev: Could change this to
```js
filter = END_FILTER.test(val.query) ? '' : val.query.slice(s);
```

Is it ever possible for `val.query` be anything other than a string?

src/itemset.js
```diff
@@ -203,30 +211,51 @@ define([
...
             return {
                 instance: ($.trim(val.src) ? {id: val.id, src: val.src} : null),
-                nodeset: val.query,
+                nodeset: val.query ? val.query.replace(END_FILTER, '') : '',
                 labelRef: labelRef.val(),
-                valueRef: valueRef.val()
+                valueRef: valueRef.val(),
+                filterRef: filterRef.val() || filter
```
millerdev: Does this mean the value of `filter` is lost if `filterRef` has a value?

src/widgets.js
```diff
@@ -292,7 +292,72 @@ define([
...
+        var widget = normal(mug, options);
+        widget.input = $("<select />")
+            .attr("name", widget.id)
+            .addClass('input-block-level');
+        widget.isDropdown = true;
+
+        var input = widget.input;
+
+        widget.setValue = function (value, option_name) {
+            var val = equivilentOption(value);
```
millerdev: spelling: equivilentOption -> equivalentOption

src/widgets.js
```diff
@@ -292,7 +292,72 @@ define([
         return widget;
     };
 
-    
+    var dropdown = function (mug, options) {
+        var widget = normal(mug, options);
+        widget.input = $("<select />")
+            .attr("name", widget.id)
+            .addClass('input-block-level');
+        widget.isDropdown = true;
```
millerdev: Is this ever used?

src/datasources.js
millerdev: fair enough

src/itemset.js
emord: This was due to an edge case. As it is now cases are:
1) select from dropdown, no `filterRef` value, only have query with no filtering
2) select from drop down, fill out `filterRef`, `fiilterRef` is tacked on to the end of the query
3) use advanced editor with no filter on the end, fill out `filterRef`, `filterRef` is tacked on to the end of the query
4) use advanced editor with a filter on the end, don't fill out `filterRef`, the filter is put into `filterRef` and the rest of the query is put into the dropdown
5) use advanced editor with a filter on the end, fill out `filterRef`, the `filterRef` overrides the filter you put in the advanced editor

Thinking about it now it seems that maybe that filter should not be erased, but should stay in the dropdown if `filterRef` is already filled out. The other cases seem good to go though

src/widgets.js
emord: Artifact from the abandoned textOrDropdown widget. will remove

emord:
@amsagoff I updated the screenshots. The move away from label and value was in response to @snopoke comment in the spec:

> late to the party here but how about calling these "Value column" and "Label column" or "Value field" since these are not the actual values or labels but the names of columns in the table that contain the value and name.

amsagoff:
Interesting. Curious waht @sheelio thinks - though also something we can save for UAT

src/util.js
millerdev: Is this still needed since my fix got merged?

src/widgets.js
```diff
@@ -292,7 +296,69 @@ define([
...
+            _.forEach(options, function(option) {
+                _this.addOption(option.value, option.text);
+            });
+        };
+
+        widget.getOptions = function () {
+            return _.map(widget.input.find('option'), function(option) {
+                return {
+                    value: option.value,
+                    text: option.text
```
millerdev: Does this work? Given how they were set, I'd expect it to be
```js
value: option.attr("value"),
text: option.text()
```

src/widgets.js
```diff
@@ -292,7 +296,69 @@ define([
...
+                };
+            });
+        };
+
+        function equivalentOption(val) {
+            val = val ? JSON.parse(val) : '';
+            return _.find(widget.getOptions(), function (option) {
+                return _.isEqual(option.value ? JSON.parse(option.value) : '', val);
```
millerdev: Same here with `option.value`

src/widgets.js
emord: Neither of those will work because I'm iterating through each option. `attr("value")` and `text()` work for the first option if  I used `widget.option.text()` but you can't iterate through them.

src/widgets.js
emord: widget.getOptions returns a dictionary of {value: value, text: text} so this is right.

src/util.js
emord: Nope. removed

src/widgets.js
millerdev: Yes, obviously. My bad.

PR #368, opened 2015-04-08
Defer duplicate Question ID error popup until click save

millerdev:
This makes it possible to create questions with conflicting IDs and resolve the warnings as convenient rather than being forced to deal with the conflict on navigating away from the question.

May be easier to review each commit individually.

@emord

### DIFF
--- a/src/jstree-plugins.js
+++ b/src/jstree-plugins.js
@@ -23,7 +23,7 @@ define([
     "use strict";
     $.jstree.defaults.conditionalevents = {
         should_activate: function () { return true; },
-        should_move: function () { return true; },
+        //should_move: function () { return true; },
         redraw_node: function () {
             var args = Array.prototype.slice.call(arguments);
             return this.parent.redraw_node.apply(this.inst, args);
@@ -36,12 +36,12 @@ define([
                 parent.activate_node.apply(this, args);
             }
         };
-        this.move_node = function () {
-            var args = Array.prototype.slice.call(arguments);
-            if(this.settings.conditionalevents.should_move.apply(this, args)) {
-                parent.move_node.apply(this, args);
-            }
-        };
+        //this.move_node = function () {
+        //    var args = Array.prototype.slice.call(arguments);
+        //    if(this.settings.conditionalevents.should_move.apply(this, args)) {
+        //        parent.move_node.apply(this, args);
+        //    }
+        //};
         this.redraw_node = function () {
             var args = Array.prototype.slice.call(arguments),
                 base = {inst: this, parent: parent};
--- a/tests/form.js
+++ b/tests/form.js
@@ -68,6 +68,24 @@ define([
             assert(util.isTreeNodeValid(black), black.getErrors().join("\n"));
         });
 
+        it("should retain conflicted mug ID on move", function () {
+            var form = util.loadXML(""),
+                hid = util.addQuestion("DataBindOnly", "hid"),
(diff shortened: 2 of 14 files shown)


src/jstree-plugins.js
```diff
@@ -23,7 +23,7 @@ define([
     "use strict";
     $.jstree.defaults.conditionalevents = {
         should_activate: function () { return true; },
-        should_move: function () { return true; },
+        //should_move: function () { return true; },
```
millerdev: Keeping this around so we don't need to figure out how to do it next time if we ever want it back. Commented to reduce performance overhead.

tests/formdesigner.lock.js
```diff
@@ -194,7 +194,7 @@ require([
                 btn.click();
                 assert.equal(
                     btn.next().find('li:contains("Cannot Change Question Type")').length,
-                    !bool);
+                    +!bool);
```
millerdev: This was the only change necessary when I switched `assert.equal` and `assert.notEqual` to use strict equality. `+` converts a boolean value to an integer.

tests/utils.js
```diff
@@ -18,7 +18,11 @@ define([
     var assert = chai.assert,
         savedForm = null,
         saveCount = 0;
-    
+
+    // monkey-patch chai to use ===/!== instead of ==/!= in assert.equal
+    assert.equal = assert.strictEqual;
+    assert.notEqual = assert.notStrictEqual;
```
millerdev: I did this because *JavaScript the Good Parts* recommends to never use `==` or `!=`. Why would we want to use them in our tests? IMO Chai should do this by default.

tests/form.js
```diff
@@ -36,6 +36,56 @@ define([
...
+                green = util.addQuestion("Text", "green"),
+                black = util.addQuestion("DataBindOnly", "black");
+            black.p.calculateAttr = "/data/blue + /data/green";
+            green.p.nodeID = "blue";
+            assert.notEqual(green.p.nodeID, "blue");
+            assert(!util.isTreeNodeValid(green), "expected validation error");
+
+            blue.p.nodeID = "orange";
+            assert.equal(green.p.nodeID, "blue");
+            assert.equal(black.p.calculateAttr, "/data/orange + /data/blue");
```
millerdev: Had to go back to my original plan of saving the conflicting ID on the mug rather than allowing duplicate IDs in the tree. Otherwise these two paths would both end up pointing to `/data/orange`.

emord:
@millerdev looks good to me. Seems well tested as well. Did you want to run this through any QA/UAT before merging?

millerdev:
@amsagoff would you like QA/UAT on this? (is there enough human-readable context here for you to tell what it is? if not, let me demo it for you on skype) I'm ambivalent. Would be happy to release as is, but also happy to run through more testing if necessary.

amsagoff:
Hey - would like to put this through full UAT if it's what you demo'd
during the all hands

millerdev:
@amsagoff this is not what I demo'd at the all hands (copy/paste). This is only the part that defers the *'question1' has the same Queston ID as another question in the same group...* modal alert until you click the save button (assuming you have not fixed the error by then).

amsagoff:
yeah that's what I meant - still think full UAT would be good, since this
is an interaction that will come up a lot after copy/paste so we want to
make sure it's all good

millerdev:
This is part of the copy-paste-dev branch

PR #370, opened 2015-04-10
drop checkbox from within formbuilder

orangejenny:
http://manage.dimagi.com/default.asp?142336


orangejenny:
Ought to be deployed with https://github.com/dimagi/commcare-hq/pull/6282

orangejenny:
@emord / @millerdev

PR #372, opened 2015-04-14
Markdown

emord:
This copies the help text to the label xml under "markdown"

More info:

https://docs.google.com/a/dimagi.com/document/d/1FsV7I-08psf5zbAC3Hk8M9IZ4a9JfHOvSZrMjeE9jE4/edit#
https://confluence.dimagi.com/display/TDH/Markdown

Not the most clear code, but didn't see a great way to do this in javarosa.

### DIFF
--- a/src/javaRosa.js
+++ b/src/javaRosa.js
@@ -1785,7 +1785,7 @@ define([
                                     return itext.name === 'markdown';
                                 });
                             if (markdownForms) {
-                                markdownForms.data = JSON.parse(JSON.stringify(helpItextForm.data));
+                                markdownForms.data = _.clone(helpItextForm.data);
                             }
                         });
 
--- a/src/javaRosa.js
+++ b/src/javaRosa.js
@@ -540,6 +540,7 @@ define([
 
         block.getUIElement = function () {
             _.each(block.getForms(), function (form) {
+                if (form === "markdown") { return; }
                 var $formGroup = block.getFormGroupContainer(form);
                 _.each(block.languages, function (lang) {
                     var itextWidget = block.itextWidget(block.mug, lang, form, options);
--- a/src/javaRosa.js
+++ b/src/javaRosa.js
@@ -41,7 +41,8 @@ define([
         RESERVED_ITEXT_CONTENT_TYPES = [
             'default', 'short', 'long', 'audio', 'video', 'image'
         ],
-        _nextItextItemKey = 1;
+        _nextItextItemKey = 1,
+        HELP_MARKDOWN;
 
     function ItextItem(options) {
         this.forms = options.forms || [];
@@ -319,7 +320,9 @@ define([
                 }
                 if (!mug.p.helpItext && mug.spec.helpItext.presence !== "notallowed") {
                     var help = mug.p.helpItext = this.createItem();
-                    help.set("", 'markdown');
+                    if (HELP_MARKDOWN) {
+                        help.addForm('markdown');
+                    }
                 }
             }
             if (!mug.options.isControlOnly) {
@@ -1211,6 +1214,7 @@ define([
             this.data.javaRosa.ItextItem = ItextItem;
             this.data.javaRosa.ItextForm = ItextForm;
             this.data.javaRosa.ICONS = ICONS;
+            HELP_MARKDOWN = this.opts().features.help_markdown;
         },
         insertOutputRef: function (mug, target, path, dateFormat) {
             var output = getOutputRef(path, dateFormat),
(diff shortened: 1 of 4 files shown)


emord:
review buddy @snopoke

snopoke:
Looks good on the surface but don't have enough context to merge. Will let @millerdev merge.

emord:
Will add a feature flag for TDH before releasing

src/javaRosa.js
```diff
@@ -310,8 +310,9 @@ define([
         updateForMug: function (mug, defaultLabelValue) {
             // set default itext id/values
             if (!mug.options.isDataOnly) {
+                var item;
```
millerdev: What do you think of renaming this variable to `label` now that it's scope reaches across the function to the help itext item as well?

src/javaRosa.js
```diff
@@ -319,6 +320,9 @@ define([
                 }
                 if (!mug.p.helpItext && mug.spec.helpItext.presence !== "notallowed") {
                     mug.p.helpItext = this.createItem();
+                    if (item) {
+                        item.set("", 'markdown');
```
millerdev: This seems odd to me, but I may be missing something. It's clearing the content of the "markdown" form for all languages in the label itext item when creating the help itext item. Is that correct?

tests/javaRosa.js
```diff
@@ -785,6 +785,7 @@ require([
                     <translation lang="en" default="">\
                         <text id="question1-label">\
                             <value>question1 en label</value>\
+                            <value form="markdown">question1 en help</value>\
```
millerdev: Why are we putting the help-text markdown content in the label itext item instead of in the help itext item? Where will we put label-text markdown content if at some point in the future we want to allow markdown in the label as well?

emord: I misread the spec. Confirmed with Ro and will. Should be in question1-help now.

src/javaRosa.js
```diff
@@ -1760,13 +1768,28 @@ define([
...
+                                return;
+                            }
+                            var mug = this.mug,
+                                helpItext = mug.p.helpItext,
+                                helpItextForm = helpItext.forms[0],
+                                markdownForms = _.find(helpItext.forms, function(itext) {
+                                    return itext.name === 'markdown';
+                                });
+                            if (markdownForms) {
+                                markdownForms.data = JSON.parse(JSON.stringify(helpItextForm.data));
```
millerdev: Would `_.clone(helpItextForm.data)` be appropriate here? That's how it's done in [`ItextForm.clone()`](https://github.com/dimagi/Vellum/blob/markdown/src/javaRosa.js#L176)

emord: Possibly. This creates a deep copy where as `_.clone` does a shallow copy.

millerdev: Yep. If that matters here, then it probably matters in `ItextForm.clone()` too, but I don't think it does. AFAIK, `data` is a mapping of language codes to itext content strings, so a shallow copy should be fine. Happy to hear if you can think of a reason that it matters.

emord: Yeah. I double checked and shallow copies are ok

tests/static/all_question_types.xml
```diff
@@ -85,6 +85,7 @@
 					</text>
 					<text id="question1-help">
 						<value>question1 en help</value>
+						<value form="markdown">question1 en help</value>
```
millerdev: Would it make sense to omit the markdown itext form if it's identical to the non-markdown version? Would that add a lot of complexity? Seems like wasted bandwidth. Might want to check with @ctsims on this.

emord: I think @wpride worked on this. He can probably comment on it.

wpride: I think that would be fine, the phone behavior would be the same either way. The only edge case I can think of is if you were trying to use markdown but the plain text also had markdown in it making it identical so Vellum eats the markdown version even though you wanted that... but even typing that out that just seems completely ridiculous that you'd want plain text with \n and # riddled all over it so, yeah, fine

If the "markdown" form exists it will always be used instead of the normal version FWIW

emord: The spec says: 
> automatically treat all text in the ‘help’ text field as markdown by generating the xml for the xforms with duplicate <value form=”markdown”></value> fields

So I think that means it's intended for markdown & non-markdown text will be the same in all cases for now. For this release there isn't any pretty UI for making the markdown as it's only for TDH.

If the feature flag is set does that mean that we would only want to write the markdown form? I'm not sure how commcare would react if it was given <value form="markdown"> but not <value>

millerdev: Hmm, so we're duplicating the `<value>` content in `<value form="markdown">` as a flag to say treat this text as markdown. That design seems less than ideal. Is there a way for the phone to access the flag that says "treat all help text as markdown for this form"?

wpride: There is not - we opted against that design

millerdev: Fair enough. @emord are you ready to have this merged?

emord: Yeah. Should be good to go.

PR #373, opened 2015-04-15
Mug messages

millerdev:
Refactor mug messaging system to show messages inline, often beside the widget where the error occurred, although this could be improved in some cases. This should be a huge usability improvement.

screenshot 2015-04-14 23 32 55

Note: merging into copy-paste-dev branch, not master.

@emord @benrudolph

### DIFF
--- a/src/commtrack.js
+++ b/src/commtrack.js
@@ -197,7 +197,7 @@ define([
             } else {
                 error = {
                     key: "commtrack-transfer-src-dest-error",
-                    level: "error",
+                    level: mug.ERROR,
                     message: "Transfer must have at least one of source " +
                              "case and destination case."
                 };
--- a/src/commtrack.js
+++ b/src/commtrack.js
@@ -191,10 +191,19 @@ define([
             }
         },
         transferMugValidation = function (mug) {
-            if (mug.p.dest.value || mug.p.src.value) {
-                return 'pass';
+            var error;
+            if ((mug.p.dest && mug.p.dest.value) || (mug.p.src && mug.p.src.value)) {
+                error = {key: "commtrack-transfer-src-dest-error"};
+            } else {
+                error = {
+                    key: "commtrack-transfer-src-dest-error",
+                    level: "error",
+                    message: "Transfer must have at least one of source " +
+                             "case and destination case."
+                };
             }
-            return 'Transfer must have at least one of source case and destination case.';
+            mug.addMessages({src: [error], dest: [error]});
+            return 'pass';
         },
         transferMugOptions = {
             typeName: 'Transfer',
--- a/src/core.js
+++ b/src/core.js
@@ -1143,6 +1143,9 @@ define([
         }).on('parent-question-type-change', function (e) {
             _this.jstree("set_icon", e.childMug.ufid, e.childMug.getIcon());
         }).on('question-remove', function (e) {
+            if (e.mug) {
+                e.mug.unbind(_this.data.core);
+            }
             var currentMug = _this.getCurrentlySelectedMug();
             if (e.mug && e.mug.parentMug && e.mug.parentMug === currentMug) {
                 _this.displayMugProperties(currentMug);
@@ -1349,6 +1352,10 @@ define([
      * @returns The tree node that was created or `false` if it was not created.
      */
     fn.createQuestion = function (mug, refMug, position) {
+        var _this = this;
+        mug.on("messages-changed", function (event) {
+            _this.setTreeValidationIcon(event.mug);
+        }, null, this.data.core);
         return this.jstree("create_node",
             refMug ? "#" + refMug.ufid : "#",
             {
@@ -1414,7 +1421,7 @@ define([
         this.toggleConstraintItext(mug);
         this.showVisualValidation(mug);
     };
-        
+
     fn.hideQuestionProperties = function() {
         this.disableUI();
     };
@@ -1517,14 +1524,6 @@ define([
         this.data.core.form.clearErrors('form-warning');
       
         if (mug) {
-            this._resetMessages(
-                this.data.core.form.errors.concat(
-                    _.map(this.getErrors(mug), function (error) {
-                        return {
-                            message: error,
-                            level: "form-warning",
-                        };
-                    })));
             this.setTreeValidationIcon(mug);
         }
     };
@@ -1545,8 +1544,13 @@ define([
     };
 
     fn.getErrors = function (mug) {
-        return mug.getErrors().concat(
-            this.data.core.form._logicManager.getErrors(mug));
+        // this should be moved into the logic manager; nodeID is clearly wrong here
+        mug.addMessage("nodeID", {
+            key: "core-logic-manager-error",
+            level: "error",
+            message: this.data.core.form._logicManager.getErrors(mug).join("\n")
+        });
+        return mug.getErrors();
     };
 
     fn._resetMessages = function (errors) {
@@ -1607,9 +1611,6 @@ define([
             $fieldsetContent = $sec.find('.fd-fieldset-content');
         options.properties.map(function (prop) {
             var elemWidget = prop.widget(mug, $.extend(prop.options, {
-                afterChange: function () {
-                    _this.showVisualValidation(mug);
-                },
                 displayXPathEditor: function (options) {
                     _this.data.core.currentlyEditedProperty = prop.options.path;
                     _this.displayXPathEditor(options);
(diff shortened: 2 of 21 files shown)


orangejenny:
nice

czue:
:+1:

src/form.js
```diff
@@ -414,9 +414,13 @@ define([
                 return node.getValue();
             });
         },
-        updateError: function (errObj, options) {
+        /**
+         * Add parsing error to the form
+         *
+         * NOTE these errors are displayed on form load only.
+         */
+        updateError: function (errObj) {
```
millerdev: The old-style error is still used for form parse [errors](https://github.com/dimagi/Vellum/blob/3b17849889117e21264739ee95779fe8af6c29af/src/parser.js#L99-L100) and [warnings](https://github.com/dimagi/Vellum/blob/3b17849889117e21264739ee95779fe8af6c29af/src/parser.js#L171-L183), which are not associated with a mug.

src/core.js
```diff
@@ -1388,10 +1373,16 @@ define([
 
         if (this._propertiesMug) {
             this._propertiesMug.teardownProperties();
+            try {
+                this._propertiesMug.validate();
+            } catch (err) {
+                // ignore error
```
benrudolph: At the very least maybe log it as a warning to the console? Unless this error is completely not important

src/form.js
```diff
@@ -764,11 +741,13 @@ define([
             //    this.fixBrokenReferences(mug);
             //}
             if (mug.options.isODKOnly) {
-                this.updateError({
+                // is this a good candidate for "info" message level?
+                mug.addMessage(null, {
+                    key: 'form-odk-only-warning',
+                    level: 'warning',
```
benrudolph: minor, but would be cool to see an ENUM like struct for the levels. especially if there is only a subset of levels that work. in the event of a typo it will throw an error

src/core.js
millerdev: The only case where there is an error here is when a new form is being loaded and the mug no longer resides in the current form, which can cause errors here (completely not important).

src/mugs.js
```diff
@@ -78,8 +78,116 @@ define([
...
+                // can the user always fix this error?
+                message = label + ' is not allowed.';
+            } else if (spec.validationFunc) {
+                try {
+                    message = spec.validationFunc(mug);
+                } catch (err) {
+                    // this should never happen
+                    message = label + " validation failed\n" + util.formatExc(err);
+                }
+                if (message === "pass") {
```
benrudolph: I remember seeing above that you return "pass" as the success message. If you never actually display the pass message, might consider returning `null` as the message. That way you know it's successful if there are no error messages versus having to check what that message is. Just a thought, don't know the entire context on the use of "pass" as the success string

src/mugs.js
```diff
@@ -94,16 +202,13 @@ define([
...
-                return [{
-                    code: "conflictedNodeId",
-                    message: "'" + dupId + "' has the same Question ID as " +
-                        "another question in the same group. Please choose " +
-                        "a unique Question ID."
-                }];
-            }
-            return [];
+            var warnings = [];
+            this.messages.each(function (msg) {
```
benrudolph: minor: good candidate for underscore chaining. `_.chain(this.messages).filter().map().value()`. can be slightly more readable

src/mugs.js
```diff
@@ -243,36 +340,120 @@ define([
...
-        return retBlock;
-    }
+    };
 
     function MugProperties (options) {
         this.__data = {};
         this.__spec = options.spec;
         this.__mug = options.mug;
-        this.shouldChange = options.shouldChange || function () { return function () {}; };
+        this.shouldChange = function () { return function () {}; };
```
benrudolph: a comment here? this looks really bizarre

src/mugs.js
benrudolph: just realizing that this is probably using it's own implementation of `each` so the comment above may not be relevant

src/mugs.js
```diff
@@ -906,14 +1081,17 @@ define([
                     if (/\s/.test(mug.p.defaultValue)) {
                         return "Whitespace in values is not allowed.";
                     }
-                    var num = 0;
-                    _.each(mug.form.getChildren(mug.parentMug), function(ele, index) {
-                        if (ele.p.defaultValue === mug.p.defaultValue) {
-                            num++;
+                    if (mug.parentMug) {
+                        var num = 0;
+                        _.each(mug.form.getChildren(mug.parentMug), function(ele, index) {
```
benrudolph: another minor, but could do `_.filter().length`

benrudolph:
Looks great, just a few minor comments. The only thing that I might want to see changed in the future is the amount of hardcoded strings. "pass" and the error levels in particular. May be biased, but I've had bad experiences in the past with misspelling constants

src/core.js
benrudolph: :+1:

src/mugs.js
millerdev: Returning `"pass"` is an artifact of the past. I agree it would be nicer to simply return null, but don't want to change all of the places that depend on it in this PR.

src/mugs.js
millerdev: The interface for `shouldChange` is documented on the form object: https://github.com/dimagi/Vellum/blob/mug-messages/src/form.js#L696-L697

src/mugs.js
millerdev: This code existed before. I put it in an if block.

benrudolph:
:white_check_mark: from me. not sure if @emord wants a look, but i'm good to merge once tests pass

src/commtrack.js
```diff
@@ -191,10 +191,19 @@ define([
             }
         },
         transferMugValidation = function (mug) {
-            if (mug.p.dest.value || mug.p.src.value) {
-                return 'pass';
+            var error;
+            if ((mug.p.dest && mug.p.dest.value) || (mug.p.src && mug.p.src.value)) {
+                error = {key: "commtrack-transfer-src-dest-error"};
```
emord: Why do you add the error key even if there is no level/message?

src/form.js
```diff
@@ -764,11 +741,13 @@ define([
             //    this.fixBrokenReferences(mug);
             //}
             if (mug.options.isODKOnly) {
-                this.updateError({
+                // is this a good candidate for "info" message level?
```
emord: +1 to info level

src/mugs.js
```diff
@@ -78,8 +78,119 @@ define([
...
+         * @param attr - The property to validate. All properties will
+         *      be validated if this argument is omitted.
+         * @returns - True if validation messages changed else false.
+         */
+        validate: function (attr) {
+            var mug = this,
+                changed;
+            this.form.updateLogicReferences(this, attr);
+            if (attr) {
+                changed = this._validate(attr);
```
emord: Small nitpick, but it seems slightly more confusing to have `this._validate` here and `mug._validate` later in same function when they are the same thing

src/form.js
millerdev: @benrudolph now that I've done this I'm wondering if it was worth it. The [constants I added](https://github.com/dimagi/Vellum/commit/e14c34c605ae4db9e26924bc5cc84011669f63db) will not protect from misspellings such as `mug.ERORR` or `mug.WRANING` due to JavaScript's liberal attribute access policy. Comparisons against the values returned by those misspellings (which return `undefined`) will not raise an error either.

src/util.js
```diff
@@ -297,7 +301,42 @@ define([
                 "XML " + (opts.not ? "should not be equivalent" : "mismatch"));
         return patch;
     };
-        
+
+    that.markdownlite = function (text) {
```
emord: Nice

src/form.js
emord: I think it's probably a net gain as most editors would autocomplete the ERROR or WARNING, but probably won't autocomplete a string as easily. I would say keep it as it's already in there even though JS isn't perfect.

Could also make the update method check the level against the known good values and have it throw an error so that we learn about a misspelling sooner rather than later

src/commtrack.js
millerdev: A message object without a message (level is not used for anything right now) [will cause a previous error with the same key to be removed](https://github.com/dimagi/Vellum/blob/mug-messages/src/mugs.js#L356-L358). This is important because it causes the error to disappear when the property becomes valid (validation is triggered by property change, among other things).

PR #374, opened 2015-04-15
Save to case

emord:
Originally was based on itemset-connector-2, but I think it survived the rebase to master.

Looks something like this (R&D feature to start with):

image

I think the hope is to support all of this: https://github.com/dimagi/commcare/wiki/casexml20

Doesn't currently have attachment. It can either be added to this PR or done in a separate PR. Figured I'd put it up here before it gets too large.

@ctsims any feedback on this/is this close to what you were expectting?

@millerdev @snopoke


src/saveToCase.js
```diff
@@ -0,0 +1,495 @@
...
+
+    function addSetValue(mug) {
+        var path = mug.absolutePath;
+
+        if (createsCase(mug)) {
+            mug.form.addSetValue('xforms-ready', path + "/case/@case_id", mug.p.case_id);
+        }
+    }
+
+    var saveCasePropWidget = function (mug, options) {
```
emord: Close to https://github.com/dimagi/Vellum/blob/master/src/widgets.js#L228 but based on `normal` instead of `base`

src/saveToCase.js
```diff
@@ -0,0 +1,495 @@
...
+                        return 'pass';
+                    }
+                },
+                "use_close": {
+                    lstring: "Close Case",
+                    visibility: 'visible',
+                    presence: 'optional',
+                    widget: widgets.checkbox
+                },
+                "close_condition": {
```
emord: Closing doesn't have any properties, but can be conditionally closed

src/saveToCase.js
```diff
@@ -0,0 +1,495 @@
...
+                        nodeset: mug.absolutePath + "/case/@date_modified",
+                        calculate: mug.p.date_modified,
+                        type: "xsd:dateTime"
+                    });
+                    ret.push({
+                        nodeset: mug.absolutePath + "/case/@user_id",
+                        calculate: mug.p.user_id
+                    });
+
+                    if (!createsCase(mug)) {
```
emord: If it's being created a set value is generated.

src/saveToCase.js
```diff
@@ -0,0 +1,495 @@
...
+                    ],
+                },
+                {
+                    slug: "create",
+                    displayName: "Create",
+                    properties: [
+                        "use_create",
+                        "create_property",
+                    ],
+                    isCollapsed: function (mug) {
```
emord: Collapse sections when they aren't being used

snopoke:
Would we want to expose the metadata properties to the user? Also at this point the form doesn't even have /data/meta.

@amsagoff  @ctsims This isn't at all what I expected for case management in Vellum. I thought it was going to be attached to individual questions as opposed to a whole new question type. Did I miss a discussion somewhere?

czue:
@snopoke this is an R&D/custom feature that is not what case management in vellum is going to look like. this is to allow for attendance lists for pathways projects

src/saveToCase.js
```diff
@@ -0,0 +1,475 @@
...
+
+    function addSetValue(mug) {
+        var path = mug.absolutePath;
+
+        if (createsCase(mug)) {
+            mug.form.addSetValue('xforms-ready', path + "/case/@case_id", mug.p.case_id);
+        }
+    }
+
+    var propertyWidget = function (mug, options) {
```
millerdev: minor: curious why you switched from `function funcName(arg)` to `var funcName = function (arg)` here? I always prefer the former as it seems like the idiomatic way to define a function. Granted, there are [scoping gotchas](http://javascriptissexy.com/javascript-variable-scope-and-hoisting-explained/) that you should understand, but as long as the function is defined before it is referenced and it's name does not collide with a variable name you should be fine.

src/saveToCase.js
```diff
@@ -0,0 +1,475 @@
...
+            var widget = widgets.normal(mug, options),
+                id = options.id;
+
+            widget.kvInput = $('<div class="control-row" />').attr('name', id);
+
+            widget.getControl = function () {
+                return widget.kvInput;
+            };
```
millerdev: Not sure if you want to future-proof this for the new [mug messages framework](https://github.com/dimagi/Vellum/blob/3b17849889117e21264739ee95779fe8af6c29af/src/widgets.js#L124-L138), or if you want to come back and do that later? I think all you need to do is define `widget.getMessagesContainer` here (and put the messages container div inside the `kvInput` div.

src/saveToCase.js
```diff
@@ -0,0 +1,475 @@
...
+            return widget;
+        },
+        saveCasePropWidget = function (mug, options) {
+            var widget = propertyWidget(mug, options);
+
+            widget.setValue = function (value) {
+                value = _.isUndefined(value) ? {} : value;
+                widget.kvInput.html(widget_update_case({
+                    props: value
+                }));
```
millerdev: `setValue` does not typically rebuild the HTML, but rather just sets input values from HTML that was setup by `widget.getUIElement`. Is there a reason you didn't follow that pattern here?

emord: Ah. my comment was outdated after I uploaded a new commit: 

> Close to https://github.com/dimagi/Vellum/blob/master/src/widgets.js#L228 but based on normal instead of base

src/saveToCase.js
emord: I was following the convention in https://github.com/dimagi/Vellum/blob/master/src/widgets.js for defining widgets which was `var widget = function(Mug, options)`

src/saveToCase.js
```diff
@@ -0,0 +1,475 @@
...
+            spec: {
+                xmlnsAttr: { presence: "optional" },
+                "date_modified": {
+                    lstring: "Date modified",
+                    visibility: 'visible',
+                    presence: 'required',
+                    widget: widgets.xPath,
+                    validationFunc: function (mug) {
+                        if (mug.p.date_modified === "") {
+                            return "Date Modified is required";
```
millerdev: I think this is not necessary (unless you don't like the default error message). Validation for `presence="required"` is done for all mugs in [mugs.js](https://github.com/dimagi/Vellum/blob/save-to-case-on-master/src/mugs.js#L23-L25).

Aside: if the default error message is the problem, I think it would be reasonable to change it globally to use a format like `(spec.lstring || ruleKey) + " is required"`. It's already been changed on the [mug messages PR](https://github.com/dimagi/Vellum/blob/3b17849889117e21264739ee95779fe8af6c29af/src/mugs.js#L121).

src/saveToCase.js
millerdev: Yeah, `baseKeyValue` seems pretty old/outdated, judging by [this comment](https://github.com/dimagi/Vellum/blob/master/src/widgets.js#L21-L23) since it falls into the category of widgets that inherit from `base`. I did a bunch of work to standardize and cleanup the `getValue`/`setValue` implementations on a bunch of the widgets, including `xPath`, but I notably did not touch `baseKeyValue` because it was doing things a really different/crufty way.

src/saveToCase.js
```diff
@@ -0,0 +1,475 @@
...
+                },
+                "case_id": {
+                    lstring: "Case ID",
+                    visibility: 'visible',
+                    presence: 'optional',
+                    widget: widgets.xPath,
+                    validationFunc: function (mug) {
+                        if (mug.p.case_id === "") {
+                            return "Case ID is required";
+                        }
```
millerdev: Wait what? Presence is optional, but validation says it's required?

src/saveToCase.js
```diff
@@ -0,0 +1,475 @@
...
+                "update_property": {
+                    lstring: "Properties To Update",
+                    visibility: 'visible',
+                    presence: 'optional',
+                    widget: saveCasePropWidget,
+                    validationFunc: function (mug) {
+                        if (mug.p.use_update) {
+                            var props = _.without(_.keys(mug.p.update_property), ""),
+                                invalidProps = _.filter(props, function(p) {
+                                    return !/[a-z_]*/.test(p);
```
millerdev: I think this will always return false, `/[a-z_]*/` matches any string including the empty string.

src/saveToCase.js
```diff
@@ -0,0 +1,475 @@
...
+                if (createsCase(mug)) {
+                    actions.push(simpleNode('create', makeColumns(mug.p.create_property)));
+                }
+
+                if (updatesCase(mug)) {
+                    actions.push(simpleNode('update', makeColumns(mug.p.update_property)));
+                }
+
+                if (closesCase(mug)) {
+                    actions.push(simpleNode('close'));
```
millerdev: Should this reference `close_condition` (if non-empty)?

millerdev: nvm. I see it's done with a bind.

src/saveToCase.js
```diff
@@ -0,0 +1,475 @@
...
+            if (!createsCase(mug)) {
+                return;
+            }
+            var values = _.object(_.map(mug.form.getSetValues(), function (value) {
+                    return [value.ref, value];
+                }));
+
+            _.each(values, function(value) {
+                if (/case\/@case_id$/.test(value.ref)) {
+                    mug.p.case_id = value.value;
```
millerdev: What if there is more than one SaveToCase property in the form? Will this end up setting the case_id of this mug to the last setvalue found with a path like `...case/@case_id`?

src/saveToCase.js
```diff
@@ -0,0 +1,475 @@
...
+        parseBindElement: function (form, el, path) {
+            var mug = form.getMugByPath(path);
+            if (!mug) {
+                var basePath = path.replace(/\/case\/[\w_]*\/[\w_]*$/, "");
+                if (path !== basePath) {
+                    mug = form.getMugByPath(basePath);
+                    if (mug.__className === "SaveToCase") {
+                        var matchRet = path.match(/\/case\/(create|update)\/([\w_]*)$/),
+                            action = matchRet[1],
+                            prop = matchRet[2],
```
millerdev: Is it legal for this to be an empty string?

tests/static/saveToCase/close_property.xml
```diff
@@ -0,0 +1,35 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<h:html xmlns:h="http://www.w3.org/1999/xhtml" xmlns:orx="http://openrosa.org/jr/xforms" xmlns="http://www.w3.org/2002/xforms" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:jr="http://openrosa.org/javarosa" xmlns:vellum="http://commcarehq.org/xforms/vellum">
+	<h:head>
+		<h:title>Untitled Form</h:title>
+		<model>
+			<instance>
+				<data xmlns:jrm="http://dev.commcarehq.org/jr/xforms" xmlns="http://openrosa.org/formdesigner/E1B03643-109D-4DC6-9737-815E4401DE79" uiVersion="1" version="1" name="Untitled Form">
+					<name />
+					<save_to_case vellum:role="SaveToCase">
+						<c:case xmlns:c="http://commcarehq.org/case/transaction/v2" case_id="" date_modified="" user_id="">
```
millerdev: Is there any semantic meaning to the `<save_to_case vellum:role="SaveToCase">` element, or is it just a placeholder? I would prefer not to add extra levels to the form hierarchy if we don't need to. I've never seen a hand-written form that has a `<case>` block inside of a container like this, but that doesn't mean it's never done that way.

I would prefer to do this the way the balance and transfer widgets do it. That is, they match a specific tag name (in this case `case`) with a specific `xmlns` attribute value (`http://commcarehq.org/case/transaction/v2`). To do that we either need to find a unique identifier on the case block (would the `case_id` attribute work? ... might need to do some nodeID gymnastics since it appears to allow a blank value) or we need to add an arbitrary vellum-specific attribute like `vellum:id="save_to_case"`.

src/saveToCase.js
```diff
@@ -0,0 +1,475 @@
...
+                    }
+                    return tmpSection;
+                });
+            }
+            return this.__callOld();
+        },
+        parseBindElement: function (form, el, path) {
+            var mug = form.getMugByPath(path);
+            if (!mug) {
+                var basePath = path.replace(/\/case\/[\w_]*\/[\w_]*$/, "");
```
millerdev: The [`\w` character class](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp#character-classes) includes `_`, btw.

millerdev: Seems like you should be able to strip some portion of the path with an RE and know immediately right here if this is a potential `SaveToCase` path. Something like this:
```js
var casePath = /\/case\/(\w+\/\w+|close|@date_modified|@user_id|@case_id)$/;
if (casePath.test(path)) {
    var mug = form.getMugByPath(path.replace(casePath, ""));
    if (mug && mug.__className === "SaveToCase") {
        // do SaveToCase stuff here
        return;
    }
}

File: tests/static/saveToCase/close_property.xml (hunk 14 above)
Comment (emord) [inline tests/static/saveToCase/close_property.xml]: I was using an example from a skype convo with @ctsims 

> [2:35:50 PM] Clayton Sims: One caveat:
[2:35:57 PM] Clayton Sims: The block should be nested in the question id
[2:36:03 PM] Clayton Sims: so if the qid is "test_a"
[2:36:21 PM] Clayton Sims: the output should be 
 \<test_a\>
    <c:case xmlns:c="...
[2:36:28 PM] Clayton Sims: so we can keep things namespaced more or less

So it seems to be required

Comment (millerdev) [inline tests/static/saveToCase/close_property.xml]: ok. interesting

File: src/saveToCase.js (hunk 6 above)
Comment (emord) [inline src/saveToCase.js]: I used the same messages css class you used in `normal` so I don't think I need to define `widget.getMessagesContainer`. Would be nice to change it to show up at the correct property, but for now it's on the mug.

File: src/saveToCase.js (hunk 7 above)
Comment (emord) [inline src/saveToCase.js]: I changed up the way this works a little (mostly just abstracting it to `refreshControl`). The reason it does that on `setValue` is that when the value is set we need to generate x number of properties. So it does that after `getUIElement`

File: src/saveToCase.js [hunk 16]
```diff
@@ -0,0 +1,575 @@
...
+                    validationFunc: function (mug) {
+                        if (mug.p.use_create) {
+                            var props = _.without(_.keys(mug.p.create_property), ""),
+                                required = ["case_type", "case_name"],
+                                optional = ["owner_id"],
+                                legal = _.union(required, optional),
+                                illegalProps = _.difference(props, legal),
+                                requiredProps = _.intersection(props, required),
+                                invalidProps = _.filter(props, function(p) {
+                                    return !INVALID_PROP_REGEX.test(p);
```
millerdev: The negatives here are making it hard for me to decipher the meaning of this code. Are these true statements?

- Any property that contains a lower-case letter or underscore is not valid.
- The set of invalid properties is all the properties that are (not not) valid. ... that can't be right.

Said another way, if I'm doing the logic right, this filter function returns true for any property that does not have at least one lower-case letter or underscore, and returning true means the property is not valid. So `"# %"` and `"AGE"` will return `true` (not valid), while `"age"` and `"#a b c"` will return `false`, meaning they are valid.

Suggestions:
- change the regex to `VALID_PROP_REGEX = /^[a-z_]+$/i`, and keep the `!` operator in the filter function.
- or `INVALID_PROP_REGEX = /[^a-z_]/i`, and drop the `!` operator from the filter function.

Note that I changed the regexps to case-insenstive since case properties are probably allowed to contain upper-case letters. Also note that these regexps do not allow unicode characters such as é, and I'm not sure if that matters or not. Please correct me if I got this all wrong.

src/saveToCase.js
```diff
@@ -0,0 +1,575 @@
...
+                    return tmpSection;
+                });
+            }
+            return this.__callOld();
+        },
+        parseBindElement: function (form, el, path) {
+            var mug = form.getMugByPath(path);
+            if (!mug) {
+                var casePathRegex = /\/case\/((create|update|index)\/(\w+)|close|@date_modified|@user_id|@case_id)$/,
+                    matchRet = _.compact(path.match(casePathRegex));
```
millerdev: Minor: you could avoid `_.compact` and simplify this a bit by using [non-capturing parentheses `(?:x)`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp#grouping-back-references). Not sure if you think this is clearer:
```js
var casePathRegex = /\/case\/((?:create|update|index)\/(\w+)|close|@date_modified|@user_id|@case_id)$/,
    match = path.match(casePathRegex);
if (match) {
    ...
    if (match[2]) {
        // path is .../case/(create|update|index)/(NAME), and match[2] is the NAME
        var prop = match[2], ...
    } else {
        // path is .../case/(close|@date_modified|@user_id|@case_id)
        var attr = match[1], ...
    }
}
```

src/templates/widget_index_case.html
```diff
@@ -0,0 +1,35 @@
...
+      <div class="span1">
+        <a href="#" class="btn fd-remove-property btn-small btn-danger">
+          <i class="icon-remove"></i>
+        </a>
+      </div>
+    </div>
+    <div class="control-row">
+      <div class="span11">
+        <input class="fd-index-property-source input-block-level"
+               value="<%=v['calculate']%>" type="text" placeholder="Calculation" />
```
millerdev: Should `<%=v['calculate']%>` be changed to `<%-v['calculate']%>`? Using `<%-` instead of `<%=` HTML-escapes the value ([docs](http://underscorejs.org/#template)). The same applies to other places as well, but I'm not going to comment on all of them.

src/saveToCase.js
emord: Couldn't quite do it that as I need create/update/index to be caught, but was able to remove use of `_.compact`

src/saveToCase.js
emord: https://github.com/dimagi/Vellum/commit/b256bda667bd4a0340d5eac8cdfeb6635b931809

src/templates/widget_index_case.html
emord: Yep. fixed those.

PR #375, opened 2015-04-17
Hide hints and help for groups

orangejenny:
Mobile ignores both help text and hint text for all kind of groups (base, repeat, and question list). Remove them from the Vellum UI.

Verified that old forms with help/hint text specified will work with the new code. Any old help/hint for groups will not be visible on the UI and will be removed from XML when the form is saved.

### DIFF
--- a/tests/javaRosa.js
+++ b/tests/javaRosa.js
@@ -518,16 +518,6 @@ require([
             assert.equal(hinLabel[0].selectionEnd, 15);
         });
 
-        _.each({group: GROUP_HELP_XML, select: SELECT1_HELP_XML}, function (XML, name) {
-            it("should not create duplicate <help> node on " + name, function () {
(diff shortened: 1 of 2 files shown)


orangejenny:
@nickpell

nickpell:
looks fine to me, I'd be comfortable with @millerdev or @emord merging

tests/javaRosa.js
```diff
@@ -518,16 +518,6 @@ require([
             assert.equal(hinLabel[0].selectionEnd, 15);
         });
 
-        _.each({group: GROUP_HELP_XML, select: SELECT1_HELP_XML}, function (XML, name) {
```
millerdev: You removed two tests here. I think we want to keep the second one (for `select`). Also, static/javaRosa/group-help.xml should probably be removed as well.

PR #376, opened 2015-04-17
Fix tab order

orangejenny:
The edit buttons for display/validation conditions, and the "auto?" checkboxes for various advanced properties, grab focus before their text boxes, which is annoying if you're tabbing through properties. It's easier to change the markup & styling than to futz with global tabindex.

Checked that this looks & behaves the same in Chrome, Firefox, Safari.

### DIFF
--- a/src/javaRosa.js
+++ b/src/javaRosa.js
@@ -464,15 +464,18 @@ define([
         var _getUIElement = widget.getUIElement;
         widget.getUIElement = function () {
             var $uiElem = _getUIElement(),
-                $autoBoxContainer = $('<div />').addClass('pull-right fd-itextID-checkbox-container'),
+                $autoBoxContainer = $('<div />').addClass('fd-itextID-checkbox-container'),
                 $autoBoxLabel = $("<label />").text("auto?").addClass('checkbox');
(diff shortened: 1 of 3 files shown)


src/javaRosa.js
```diff
@@ -464,15 +464,16 @@ define([
         var _getUIElement = widget.getUIElement;
         widget.getUIElement = function () {
             var $uiElem = _getUIElement(),
-                $autoBoxContainer = $('<div />').addClass('pull-right fd-itextID-checkbox-container'),
+                $autoBoxContainer = $('<div />').addClass('fd-itextID-checkbox-container'),
                 $autoBoxLabel = $("<label />").text("auto?").addClass('checkbox');
 
             $autoBoxLabel.prepend($autoBox);
             $autoBoxContainer.append($autoBoxLabel);
+            $uiElem.css('position', 'relative');
```
orangejenny: I don't love having styling in js files, but making .widget.control-group position: relative in LESS affects the layout of the basic fields.

orangejenny:
@nickpell

nickpell:
notifying @millerdev and @emord

millerdev:
I've been annoyed with in the past, but never got around to fixing it. Nice work!

PR #379, opened 2015-04-20
Validation message inputs disappearing

orangejenny:
http://manage.dimagi.com/default.asp?164598

In certain forms, when constraint text changes and fireChangeEvents goes through all mugs to update any that share itext with the changed itext, ids are blank for both widget.getItextItem() and some other mug's labelItext...so a question-label-text-change event gets fired for that other mug, which causes  toggleConstraintItext to run, which updates the visibility of the currently-displayed validation message inputs based on the constraintMsgItext of this other mug, which isn't right.

I'm not clear on when it's expected for an itext id to be blank, and therefore not sure if this is a reasonable fix or a band-aid for a larger problem.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1344,6 +1344,10 @@ define([
 
     fn.toggleConstraintItext = function (mug) {
         // todo: don't handle this one-off in the UI layer
+        var current = this.getCurrentlySelectedMug();
+        if (current && current.ufid !== mug.ufid) {
+	         return;
+        }
         var state = (mug.p.constraintMsgItext &&
                      (!mug.p.constraintMsgItext.isEmpty() ||
                       mug.p.constraintAttr)),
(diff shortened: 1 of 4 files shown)


orangejenny:
needs judgment from @millerdev or @emord 

enchantress @nickpell

millerdev:
Ugh, yeah, a core method name like `toggleConstraintItext` the core UI code clearly indicates that some necessary abstractions are missing (everything itext-related should be done in javaRosa). Note the [rather unhelpful warning on `getCurrentlySelectedMug`](https://github.com/dimagi/Vellum/blob/ebb731fa060aebb883772ff7b9c70dba1b677d21/src/core.js#L1036-L1039)...

This change looks OK for now, but probably some refactoring could be done to flesh out the abstractions and clean up this mess. Is it possible to write a test that replicates the bug?

millerdev:
Continuing from previous comment...

Is it possible to write a test that replicates the bug? I think even if it's dependent on UI code it's worth trying to write a test so we can tell that we're not breaking it going forward, and also to remind us to write tests for the core logic if refactoring happens in the future (which will presumably break the test for this code).

orangejenny:
@millerdev 

Egad, I hadn't noticed the `getCurrentlySelectedMug` comment, that's not good...you're right about adding a test, I'll do that.

orangejenny:
@millerdev test added

PR #380, opened 2015-04-21
Copy/paste

millerdev:
Spec: https://docs.google.com/a/dimagi.com/document/d/1iBIIpwyFhDcu-ccxo720nGs0Dzs-N0zBq646tO64nOM/edit#heading=h.vthcylu009oo

Note this is merging into the copy-paste-dev branch. Feel free to merge once this is reviewed and all comments have been addressed. When QA and UAT is complete, copy-paste-dev will be merged into master.

@emord 
cc @benrudolph


src/commtrack.js
```diff
@@ -136,7 +136,7 @@ define([
                 return {
                     xmlns: LEDGER_XMLNS,
                     type: mug.p.nodeID,
-                    "entity-id": attrs.src || "",
+                    "entity-id": attrs["entity-id"] || "",
```
millerdev: This was a (probably benign) bug in the commtrack module. Unrelated to copy/paste.

src/core.js
```diff
@@ -671,9 +674,22 @@ define([
...
+                if (event.keyCode === 27) {
+                    $modal.modal('hide');
+                }
+            };
+        $body.one("keyup", keyup);
+        $modal.on('hide', function () {
+            $body.off("keyup", keyup);
+        }).one("shown", function () {
+            $modal.find(".btn-default:last").focus();
+        });
```
millerdev: This makes Enter and Escape keys work in dialogs.

src/copy-paste.js
```diff
@@ -0,0 +1,347 @@
...
+
+    function onCut(opts) {
+        var $focus = $(':focus');
+        if ($focus.is('.jstree-anchor')) {
+            var text = opts.cut();
+            if (text) {
+                focusTextarea($focus, text);
+                setTimeout(function () {
+                    unfocusTextarea($focus);
+                }, 10);
```
benrudolph: just curious what the timeout is for

src/copy-paste.js
```diff
@@ -0,0 +1,347 @@
...
+        return string;
+    }
+
+    function headerKey(item) {
+        var rank = "1",
+            itext = /^(.*)Itext(?::([^-]+)-(.*))?$/.exec(item);
+        if (itext) {
+            rank = "0"; // itext before other fields
+            // label before all other itext types
+            if (itext[1] === "label") { itext[1] = "!"; }
```
benrudolph: small note, sometimes i like to label my indices for more readability:
```
var TYPE = 1,
   LANG = 2,
   FORM =3;
itext[TYPE] == 'something'
```  
totally your call

src/copy-paste.js
```diff
@@ -0,0 +1,347 @@
...
+            }
+            node = node.parent;
+        }
+        if (!pos) {
+            pos = {};
+            if (!node.mug) {
+                pos.error = "Cannot insert " +
+                            nameOf(values.type) + " into tree root";
+            } else {
+                pos.error = "Cannot insert $1 into or after $2"
```
benrudolph: cool, never seen string interpolation done like this. more readable for sure

src/copy-paste.js
```diff
@@ -0,0 +1,347 @@
...
+            if (children.length) {
+                row = [row].concat(_.map(children, serialize));
+            }
+            return row;
+        }
+
+        var headings = {id: true, type: true},
+            header = [],
+            form = mugs[0].form,
+            rows = _.filter(_.flatten(_.map(mugs, serialize)), _.identity);
```
benrudolph: :+1:

src/copy-paste.js
```diff
@@ -0,0 +1,347 @@
...
+            values, pos, parent;
+        errors.add = function (message) {
+            errors.update(null, {
+                key: message,
+                level: mugs.ERROR,
+                message: message
+            });
+        };
+        vellum.beforeBulkInsert(form);
+        for (; row; row = next()) {
```
benrudolph: can a `while (row = next())` work here?

src/copy-paste.js
```diff
@@ -0,0 +1,347 @@
...
+            focusTextarea($focus);
+            setTimeout(function () {
+                var pasteValue = unfocusTextarea($focus, true);
+                // on chrome this gets called twice,
+                // the first time with a blank value
+                if (pasteValue) {
+                    var errors = opts.paste(pasteValue);
+                    if (errors.length) {
+                        vellum._resetMessages([{
+                            level: "parse-warning",
```
benrudolph: wasn't sure if you wanted to keep using constants, but thought i'd flag. (another benefit of the constant is if another developer makes modifications, they know what levels to choose from)

src/mugs.js
```diff
@@ -295,6 +295,61 @@ define([
...
+                var value = mug.p[key];
+                if (spec.serialize) {
+                    value = spec.serialize(value, key, mug, data);
+                    if (!_.isUndefined(value)) {
+                        data[key] = value;
+                    }
+                } else if (!(
+                                value === null ||
+                                value === "" ||
+                                _.isUndefined(value) ||
```
benrudolph: does `!value` work here to cover the 3 cases? or can value be `0`

src/core.js
```diff
@@ -1076,23 +1100,6 @@ define([
...
-
-    fn.flipUI = function (state) {
-        var $props = this.$f.find('.fd-question-properties');
-        if (state) {
-            $props.show();
-        } else {
-            $props.hide();
-        }
-    };
-
```
millerdev: These were consolidated into a clearer API: `showQuestionProperties` and `hideQuestionProperties`

src/mugs.js
```diff
@@ -679,7 +783,7 @@ define([
             },
             rawControlAttributes: {
                 presence: 'optional',
-                lstring: "Extra Control Attributes"
+                lstring: "Extra Control Attributes",
```
benrudolph: how bout them trailing commas @orangejenny :)

src/templates/alert_global.html
```diff
@@ -3,7 +3,7 @@
     <h4><% if (messageType.icon) { %><i class="<%=messageType.icon%>"></i> <% } %><%=messageType.title%></h4>
     <ul>
         <% _.each(messages, function (message) { %>
-        <li><%=message%></li>
+        <li><%-message%></li>
```
benrudolph: shouldn't this stay `<%=`

tests/copy-paste.js
```diff
@@ -0,0 +1,780 @@
...
+        (function (undefined) {
+            // is this what we want? does it matter?
+            test(undefined, undefined);
+        })();
+        test("{1} ", "{1} ");
+        test("{1} {2}", '"{1} {2}"');
+        test("[1] [2]", '"[1] [2]"');
+        test("01", '"01"');
+    });
+});
```
benrudolph: :+1:

benrudolph:
nice work! this is a lot of code for me to grok, so i can only really comment on the more minor stuff. none of my comments are blocking

src/copy-paste.js
millerdev: Honestly I'm not sure. It's a pattern that came from @dannyroberts [original contribution of copy/paste code for Vellum](https://github.com/dimagi/Vellum/commit/05f4f59ceeee53a30845cb091bea8def4925ae77#diff-815d962c2df9702e02af486ea3984d4bR70). Handling copy/paste in JS is tricky. I'm guessing it's necessary for some strange edge case.

src/copy-paste.js
millerdev: It could, but that produces lint error (which breaks the build): src/copy-paste.js: line 269, col 28, Expected a conditional expression and instead saw an assignment.

src/copy-paste.js
millerdev: This is the legacy/global message system, and this is the pattern that's used in other places. Constants might be a nice addition, but I'm not doing it as part of this PR, which is already big enough.

src/mugs.js
millerdev: That does appear to work, so I changed it. See also comment on 39052ff929a5ce142fa4845b1ccf0bd5d20d8fcb. I'd prefer to simply use `!_.isEmpty(value)`, but unfortunately `_.isEmpty(true) === true`, which is a deal breaker.

src/mugs.js
millerdev: I’m so conflicted on the best practice now. Is it better to maintain consistency with existing code (and omit trailing commas) or add them because it's more convenient? I think I'm tending to do whatever feels right at the time, which ironically feels wrong when I think about it in a broader sense.

These were added incidentally as the copy/paste plugin was developed (more spec attributes were added here, and then subsequently removed). I probably should have removed the commas as well just to reduce noise in the PR.

src/templates/alert_global.html
millerdev: Nope. [`<%-` HTML-escapes the content](https://cdn.rawgit.com/jashkenas/underscore/1.5.2/index.html#template).

benrudolph: cool sounds good

emord:
@millerdev doesn't merge cleanly atm

millerdev:
@emord fixed

tests/copy-paste.js
```diff
@@ -0,0 +1,780 @@
...
+            eq(mod.copy(), [
+                ["id", "type", "labelItext:en-default", "labelItext:hin-default"],
+                ["/group", "Group", "group", "group"],
+                ["/group/group1", "Group", "group", "group"],
+                ["/group/group2", "Group", "group", "group"],
+                ["/group/group3", "Group", "group", "group"],
+            ]);
+        });
+
+        it("should resolve pasted question with conflicting question ID", function () {
```
emord: When would this actually happen? The only times I can see are:

1) Whenever you copy/paste in one form and then try to paste the same thing in another form 
2) You copy an resolved node. Delete the original node and paste in the copied form.

Is that right?

tests/copy-paste.js
```diff
@@ -0,0 +1,780 @@
...
+                ["/text2", "Text", "non", "non", "text-label"],
+            ]);
+            call("createXML");
+            var text = util.getMug("text"),
+                text2 = util.getMug("text2");
+            // NOTE which one gets auto-renamed is somewhat arbitrary
+            assert.equal(text.p.labelItext.autoId, true);
+            assert.equal(text.p.labelItext.id, "text-label");
+            assert.equal(text2.p.labelItext.autoId, false);
+            assert.equal(text2.p.labelItext.id, "text-label2");
```
emord: Should this test also make sure that the actual text for text2 is still 'non' and not 'auto'?

tests/copy-paste.js
```diff
@@ -0,0 +1,780 @@
...
+            eq(mod.copy(), [
+                ["id", "type", "labelItext:en-default", "labelItext:hin-default"],
+                ["/text", "Text", "Label", "Label"],
+            ]);
+        });
+
+        it("should ignore extra cells in header row", function () {
+            util.loadXML("");
+            paste(tsv.tabDelimit([
+                HEADER.concat(["", ""]),
```
emord: Does this still work if extra cells aren't blank?

tests/copy-paste.js
millerdev: This happens if you copy a question with a conflicted question ID, and then paste that somewhere where the question ID no longer has a conflict. It verifies that the conflict is resolved on paste.

tests/copy-paste.js
millerdev: Not a bad idea.

tests/copy-paste.js
millerdev: I'm sure it does, given how it's [comparing the preamble](https://github.com/dimagi/Vellum/pull/380/files#diff-815d962c2df9702e02af486ea3984d4bR251). The extra cells are ignored. Do you think that's a problem?

emord: No problem. Just wanted to make sure since the test only tested empty cells.

src/copy-paste.js
dannyroberts: Yeah in the commit @millerdev linked to, I'd spent a lot of time coming up with code that works on both firefox and chrome, because they have different exact rules about when copy/paste events are fired that I don't currently recall. My guess is that if you replaced the timeout expression with just `unfocusTextarea($focus);`, it would break on one of chrome or firefox. (Like timeout of 10ms is to wait for the default action to happen and then apply this on top of it, or something.)

czue:
hiyo!

millerdev:
note this is only merged into copy-paste-dev. QA/UAT is still in progress

czue:
yeah, still exciting though.

PR #381, opened 2015-04-21
Have uglify only output ASCII

orangejenny:
My inability to run vellum--to-hq successfully was caused by unicode characters in codemirror.js Still a mystery why this only affected me, but making uglify escape unicode fixes it and doesn't obviously harm anything else.


millerdev:
This is causing errors when I `make`, and the files are not being minified:
```
Uglify2 file: /Users/dimagi/code/vellum/_build/src/exclude.js
Error: Cannot uglify2 file: /Users/dimagi/code/vellum/_build/src/exclude.js. Skipping it. Error is:
Error
Uglify2 file: /Users/dimagi/code/vellum/_build/src/global-deps.js
Error: Cannot uglify2 file: /Users/dimagi/code/vellum/_build/src/global-deps.js. Skipping it. Error is:
Error
Uglify2 file: /Users/dimagi/code/vellum/_build/src/local-deps.js
Error: Cannot uglify2 file: /Users/dimagi/code/vellum/_build/src/local-deps.js. Skipping it. Error is:
Error
Uglify2 file: /Users/dimagi/code/vellum/_build/src/main-components.js
Error: Cannot uglify2 file: /Users/dimagi/code/vellum/_build/src/main-components.js. Skipping it. Error is:
Error
```
Very helpful "Error" there, Uglify2...

orangejenny:
@millerdev Builds successfully now.

millerdev:
It still bothers me a bit that you need this to do a successful build. Would like to get to the bottom of that so we at least know why. I'll merge anyway since this must be a major inconvenience to you.

orangejenny:
Yeah, I don't like it, either. I can build Vellum on my desktop at home, but not on my laptop, and have spent a good deal of time trying to track down what's different about the two environments (guessing some kind of encoding-related OS setting).

PR #384, opened 2015-04-22
Fix markdown

emord:
http://manage.dimagi.com/default.asp?166113#924497

Wasn't adding markdown forms to help itext that already existed

enchantress: @snopoke

### DIFF
--- a/src/javaRosa.js
+++ b/src/javaRosa.js
@@ -101,6 +101,11 @@ define([
                 this.forms.splice(index, 1);
             }
         },
+        cloneForm: function (cloneFrom, cloneTo) {
+            var newForm = this.getOrCreateForm(cloneFrom).clone();
+            newForm.name = cloneTo;
+            this.forms.push(newForm);
+        },
         get: function(form, language) {
             if (_.isUndefined(form) || form === null) {
                 form = "default";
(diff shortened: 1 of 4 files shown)


src/javaRosa.js
```diff
@@ -309,20 +314,36 @@ define([
...
+            function missingMarkdownForm(forms) {
+                return _.filter(forms, function(form) {
+                    return form.name === 'markdown';
+                }).length === 0;
+            }
+
             // set default itext id/values
             if (!mug.options.isDataOnly) {
-                if (!mug.p.labelItext && mug.spec.labelItext.presence !== "notallowed") {
+                if (!mug.p.labelItext && getPresence(mug.spec.labelItext) !== "notallowed") {
```
emord: Itext has a function for presence, so I think this was probably always broken (but didn't have bad side effects)

src/javaRosa.js
```diff
@@ -309,20 +314,36 @@ define([
             return this.updateForMug(mug, mug.getLabelValue());
         },
         updateForMug: function (mug, defaultLabelValue) {
+            function getPresence(itext) {
+                if (_.isFunction(itext.presence)) {
+                    return itext.presence(mug.options);
+                }
+                return itext.presence;
+            }
```
millerdev: I think this should be a method on Mug (`mug.getPresence("labelItext")`), so you wouldn't have to re-implement the protocol here and every other place where it's needed. That can be done on a separate PR though.

PR #385, opened 2015-04-23
Resolve conflicted question ID on dismiss duplicate ID error

millerdev:
screenshot 2015-04-22 22 41 31

Click the X to dismiss the error...

screenshot 2015-04-22 22 41 49

Additionally, I attempted to clean up event unbinding. The API is a little gnarly, but at least it's documented.

Merging into copy-paste-dev

@emord 
cc @benrudolph

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1337,7 +1337,7 @@ define([
         var _this = this;
         mug.on("messages-changed", function (event) {
             _this.setTreeValidationIcon(event.mug);
-        }, null, this.data.core);
+        }, null, null, this.data.core);
         return this.jstree("create_node",
             refMug ? "#" + refMug.ufid : "#",
             {
@@ -1411,10 +1411,7 @@ define([
         function refreshMessages() {
             $messages.empty().append(widgets.getMessages(mug, null));
         }
-        mug.on("messages-changed", refreshMessages, null, $messages);
-        mug.on("teardown-mug-properties", function () {
-            mug.unbind($messages);
-        }, null, $messages);
+        mug.on("messages-changed", refreshMessages, null, "teardown-mug-properties");
         refreshMessages();
 
         $props.show();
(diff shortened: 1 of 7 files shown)


src/core.js
```diff
@@ -1337,7 +1337,7 @@ define([
         var _this = this;
         mug.on("messages-changed", function (event) {
             _this.setTreeValidationIcon(event.mug);
-        }, null, this.data.core);
+        }, null, null, this.data.core);
```
millerdev: This is the ugly one, but this is also the only place that this variant of the API is used.

src/util.js
```diff
@@ -101,39 +102,69 @@ define([
...
                 registry[type].push(handler);
             } else {
                 registry[type] = [handler];
             }
+            if (unbindOn) {
+                if (!unbinders[unbindOn]) {
+                    unbinders[unbindOn] = [];
+                }
+                if (unbinders[unbindOn].indexOf(context) === -1) {
+                    unbinders[unbindOn].push(context);
```
millerdev: This is so only one unbinder is created for a given context. This makes the unbinding more efficient than it was before because now many different event handlers can use the same unbinder without a globally shared context.

src/core.js
benrudolph: interesting, is this a custom event api or are you using some code from somewhere? (looks fine, just wondering)

benrudolph: jk, i should really do a pass on the code before commenting

src/util.js
```diff
@@ -101,39 +102,69 @@ define([
         };
         /**
          * Register an event handler to be called each time an event is fired.
+         *
+         * @param type - Event type string.
+         * @param method - Event handler method.
+         * @param parameters - Parameters to be passed to method. If null
```
benrudolph: no action required, but there may be a time when you need both parameters and the event object

benrudolph:
:white_check_mark:

src/util.js
millerdev: Yeah, `parameters` came with the original code from *JavaScript: The Good Parts*. Personally I wish it wasn't there at all. It's very simple to work around it with a closure. Maybe I'll rip it out someday.

PR #386, opened 2015-04-23
Move code around

millerdev:
- Group error handling functions in core.js
- Move Mug class to top of mugs.js

This was cherry-picked from the copy-paste-dev (e0fef536473f678480d5206d628f3d2e15a78b0a), and it should make merges easier.

@emord @orangejenny 
cc @benrudolph

### DIFF
--- a/src/mugs.js
+++ b/src/mugs.js
@@ -14,6 +14,190 @@ define([
     util,
     undefined
 ) {
+    /**
+     * A question, containing data, bind, and control elements.
+     */
+    function Mug(options, form, baseSpec, copyFromMug) {
+        var properties = null;
+        util.eventuality(this);
+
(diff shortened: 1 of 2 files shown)


src/mugs.js
```diff
@@ -444,176 +601,6 @@ define([
...
-        setItextID: function (val) {
-            var labelItext = this.p.labelItext;
-            if (labelItext) {
-                labelItext.id = val;
-            }
-        },
-        
-        getItext: function () {
-            return this.p.labelItext;
-        },
```
millerdev: The above two (unused) Itext functions were removed during the code move. Should have at least done that as a separate commit...

orangejenny:
ran tests locally

PR #387, opened 2015-04-23
Escape apostrophes in itext

emord:
http://manage.dimagi.com/default.asp?166079#925204

This was already enforced for labelItext, but it isn't checked for auto ids. (and probably shouldn't be. Those should generate valid labels)

On other mugs it is already enforced as you can't have an apostrophe in a node id.


emord:
enchantress @snopoke

PR #389, opened 2015-04-24
Paste into different group

millerdev:
Context: pasting questions with logic references into a group that is different from the source group where they were copied from. Making these pass is non-trivial.

@emord what do you think of these messy edge cases?

@amsagoff, can you tell what's happening in these test cases? If so, do you think they assert the correct behavior?


millerdev:
BTW, I don't think we should block rolling out copy/paste on this, but I'm happy to hear other viewpoints if others disagree.

tests/copy-paste.js
```diff
@@ -630,6 +630,53 @@ require([
                    "double_trouble should not be valid");
         });
 
+        it("should not update broken references on paste into group", function () {
```
emord: Looks like current (and correct) behavior

tests/copy-paste.js
```diff
@@ -630,6 +630,53 @@ require([
...
+            paste([
+                ["id", "type", "calculateAttr"],
+                ["/four", "DataBindOnly", "2 + /data/two"],
+            ]);
+            var four = util.getMug("group/four");
+            assert.equal(four.p.calculateAttr, "2 + /data/two");
+            assert(!util.isTreeNodeValid(four), "four should not be valid");
+        });
+
+        it("should update question references on paste into group", function () {
```
emord: Is your plan for these two tests to treat the root as `/data/`, so `/group` will be transformed to `/data/group`? and resolve that way? That seems like it could get pretty messy.

Seems like something we would want to test out with FMs though. Not sure what they would expect to happen, although I think not changing anything in calculate or similar fields is the easiest/safest way to go.

tests/copy-paste.js
millerdev: Yes, that's right. This test was included for completeness. It is passing.

tests/copy-paste.js
millerdev: > Is your plan for these two tests to treat the root as `/data/`, so `/group` will be transformed to `/data/group`?

I don't really have a plan yet. As you noticed, the copy/paste framework does not include the `/data` path prefix in the `id` column, but that's not too relevant since I'm not really considering doing any path analysis/fixup on the serialized copy/paste data.

One possible solution is to initially put questions into some sort of sandbox area on paste, and then run the logic analyzer on them to link expressions to question paths. Finally "move" them out of the sandbox into the real form, thus invoking the path mutation logic that updates expressions when questions are moved. This sounds messy.

Another idea would be to somehow annotate expressions that reference questions on copy, and then use those annotations to fix up paths on paste. This seems cleaner, but still not trivial.

> Seems like something we would want to test out with FMs though.

Definitely. I wrote these tests to document the issue, but without any specific plan to fix them. I think copy/paste is pretty valuable, even without fixing this issue, so it seems reasonable to get it out the door and then come back and work on this issue as people start to run into it and can inform us of what they expected to happen.

tests/copy-paste.js
emord: Why not include it in copy-paste (just curious)?

millerdev: Heh, good question. I think mainly because I only thought of it in the context of writing the next two tests. Also, currently copy-paste is not making a statement about how this is handled. There are many tests that could be written to confirm implementation details, but they're not necessary since there is no feature or bug that requires them.

PR #392, opened 2015-04-27
Fix node id regex

emord:
screwed up regex here.

https://github.com/dimagi/Vellum/pull/387/files

@millerdev 

Tests will come. but want to deploy soon


czue:
want me to wait for tests before merge?

millerdev:
tests passed locally

PR #394, opened 2015-04-29
Mug refactor

millerdev:
Add `Mug.getPresence(property)` and `Mug.isVisible(property)`

@emord

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -2059,26 +2059,8 @@ define([
     };
 
     function getWidgetClassAndOptions(propPath, mug) {
-        var propDef = mug.p.getDefinition(propPath),
-            propVal = mug.p[propPath];
-
-        // handle properties whose visibility depends on other properties'
-        // visibility (implied visible_if_present on the depending property
-        // takes precedence)
-        if (_.isUndefined(propVal) &&
-            propDef && propDef.visibility &&
-            mug.p.getDefinition(propDef.visibility) &&
-            !getWidgetClassAndOptions(propDef.visibility, mug))
-        {
-            return null;
-        }
-
-        if (!propDef || propDef.visibility === 'hidden' ||
-            (_.isFunction(propDef.visibility) && !propDef.visibility(mug, propDef)) ||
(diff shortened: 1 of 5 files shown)


src/core.js
```diff
@@ -2059,26 +2059,8 @@ define([
...
-        var propDef = mug.p.getDefinition(propPath),
-            propVal = mug.p[propPath];
-
-        // handle properties whose visibility depends on other properties'
-        // visibility (implied visible_if_present on the depending property
-        // takes precedence)
-        if (_.isUndefined(propVal) &&
-            propDef && propDef.visibility &&
-            mug.p.getDefinition(propDef.visibility) &&
-            !getWidgetClassAndOptions(propDef.visibility, mug))
```
millerdev: That `_.isUndefined(propVal)` up there is causing me some consternation. Dependent visibility was ignored unless the property value was undefined... so strange.

src/core.js
```diff
@@ -2059,26 +2059,8 @@ define([
...
-        {
-            return null;
-        }
-
-        if (!propDef || propDef.visibility === 'hidden' ||
-            (_.isFunction(propDef.visibility) && !propDef.visibility(mug, propDef)) ||
-            (_.isUndefined(propVal) &&
-             (propDef.visibility === "visible_if_present" ||
-              propDef.presence === "notallowed")))
```
millerdev: `propDef.presence === "notallowed"` here is also troubling. Doesn't seem to make much sense to have `notallowed` (presence) mean the same thing as `visible_if_present` (visibility).

millerdev: Note both of the above were added in huge commits where lots of things were happening, so I'm inclined to believe that they were mistakes.

src/mugs.js
```diff
@@ -405,7 +472,7 @@ define([
         control: {
             appearance: {
                 deleteOnCopy: true,
-                visibility: 'optional',
+                visibility: 'visible',
```
millerdev: This was a bug.

PR #395, opened 2015-04-29
Fix “Case” property not showing for Balance

millerdev:
http://manage.dimagi.com/default.asp?166979

`_.extend` is tricky. In this case it was mutating the first argument, which meant Balance's section data was getting Transfer's properties.

@orangejenny


emord:
:+1: will wait for tests

PR #396, opened 2015-04-29
Always write value on labels

emord:
http://manage.dimagi.com/default.asp?166694#929816

Functionality changed here https://github.com/dimagi/Vellum/commit/5b0954416ca1a3c6b4ff17c37e5d3197bfd910ed
for no apparent reason.

@millerdev

### DIFF
--- a/tests/javaRosa.js
+++ b/tests/javaRosa.js
@@ -18,6 +18,7 @@ require([
     'text!static/javaRosa/itext-item-rename-group-move.xml',
     'text!static/javaRosa/itext-item-non-auto-id.xml',
     'text!static/javaRosa/select1-help.xml',
+    'text!static/javaRosa/no-label-text-one-lang.xml',
     'text!static/markdown/with-markdown.xml',
     'text!static/markdown/no-markdown.xml'
 ], function (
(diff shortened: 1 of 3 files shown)


tests/javaRosa.js
```diff
@@ -595,6 +597,31 @@ require([
...
+        it("should replace the default form with placeholder when cleared", function(){
+            util.loadXML("");
+            util.addQuestion('Trigger', 'label');
+            util.clickQuestion('label');
+            $('[name=itext-en-label]').val('blah').change();
+            $('.itext-block-label-add-form-image').click();
+            $('[name=itext-en-label]').val('').change();
+            util.assertXmlEqual(call("createXML"), 
+                                NO_LABEL_TEXT_ONE_LANG_XML,
+                                {normalize_xmlns: true});
```
millerdev: When I manually do the steps performed by this test I see a placeholder in the label textarea, but Tools > Edit Source XML shows an empty `<value></value>`. That seems like a bug.

millerdev:
Interesting. I'm merging this because you point out that the `if (val)` block was added recently, and for no apparent reason.

However, it appears that there is another bug somewhere. See [other comment](https://github.com/dimagi/Vellum/pull/396/files#r29368681).

PR #398, opened 2015-04-30
Allow drag/drop question into App Callout value

millerdev:
http://manage.dimagi.com/default.asp?164693
@emord


wpride:
ow ow ow ow ow

emord:
tests pass locally

PR #399, opened 2015-05-04
Ignore retain overhaul

millerdev:
This is a big overhaul of the ignore-but-retain plugin. It makes the plugin smarter about handling ignored things that look like questions. There is a notable UI change: ignored question fragments (data, bind, and control nodes) are now represented by a question-like object in the tree.

screenshot 2015-05-04 16 07 08

Fixes
http://manage.dimagi.com/default.asp?146253 - bind nodes moved to top of form
http://manage.dimagi.com/default.asp?149925 - can't rename questions when ignore/retain is active
http://manage.dimagi.com/default.asp?164398 - case block with nested/ignored update block

cc @wspride

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -2107,6 +2107,10 @@ define([
         return parser.parseBindElement(form, el, path);
     };
 
+    fn.getControlNodeAdaptorFactory = function (tagName) {
+        return this.data.core.controlNodeAdaptorMap[tagName];
+    };
+
     /**
      * Extension point for mug setup during control node parsing
      *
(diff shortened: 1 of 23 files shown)


millerdev:
cc @benrudolph

czue:
baller

millerdev:
Looking at some test failures on Firefox

millerdev:
Fixed

wpride:
I got 99 problems but ignore:retain ain't one

benrudolph:
:+1: from me. nothing jumped out to me, but someone with more context should definitely review

PR #400, opened 2015-05-05
Rename analytics events

NoahCarnahan:
This is part of a conversion of our analytics tracking from Segment to KISSmetrics. This needs to be released at the same time as the corresponding changes to HQ are changed, which is why I put the don't merge tag on it. @millerdev How does the vellum deploy process work exactly? I've seen the vellum-to-hq script, but I'm not sure when exactly that gets run.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -154,8 +154,8 @@ define([
                     return; // abort
                 }
                 _this.ensureCurrentMugIsSaved(function () {
-                    if (!_.isUndefined(window.analytics)) {
-                        window.analytics.track("Clicked Save in Formbuilder");
+                    if (!_.isUndefined(window._kmq)) {
+                        window._kmq.push(["record", "Clicked Save in Formbuilder"]);
                     }
                     _this.validateAndSaveXForm(forceFullSave);
                 });
@@ -1346,8 +1346,8 @@ define([
             if (!foo) {
                 throw new Error("cannot add " + qType + " at the current position");
             }
-            if (!_.isUndefined(window.analytics)) {
-                window.analytics.track("Added a question");
+            if (!_.isUndefined(window._kmq)) {
+                window._kmq.push(["record", "Added a question"]);
             }
             mug = _this.data.core.form.createQuestion(foo.mug, foo.position, qType);
             var $firstInput = _this.$f.find(".fd-question-properties input:text:visible:first");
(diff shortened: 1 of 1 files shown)


millerdev:
@NoahCarnahan `vellum-to-hq master` is used to build and update Vellum in HQ. It should be run prior to deploying HQ (it makes a commit on HQ master). It's important to monitor the `manifest.txt` file in the commit to verify that Vellum dependencies in your build environment were reasonably up to date. To be on the safe side, you can ask one of the Vellum devs to build a new version when you're ready.

src/core.js
```diff
@@ -154,8 +154,8 @@ define([
                     return; // abort
                 }
                 _this.ensureCurrentMugIsSaved(function () {
-                    if (!_.isUndefined(window.analytics)) {
-                        window.analytics.track("Clicked Save in Formbuilder");
+                    if (!_.isUndefined(window._kmq)) {
+                        window._kmq.push(["record", "Clicked Save in Formbuilder"]);
```
millerdev: Can we make a generic API in HQ to register analytics events? Ideally it would be the same API that *segment?* has (`window.analytics.track(EVENT_NAME)`) so we wouldn't have to change things in the form builder every time we decide to switch analytics engines. Also note that [copy-paste-dev](https://github.com/dimagi/Vellum/compare/copy-paste-dev?expand=0) (currently in QA) makes [some](https://github.com/dimagi/Vellum/compare/copy-paste-dev?expand=0#diff-815d962c2df9702e02af486ea3984d4bR188) [calls](https://github.com/dimagi/Vellum/compare/copy-paste-dev?expand=0#diff-815d962c2df9702e02af486ea3984d4bR201) [to](https://github.com/dimagi/Vellum/compare/copy-paste-dev?expand=0#diff-815d962c2df9702e02af486ea3984d4bR248) `window.analytics.track()` as well.

NoahCarnahan: I agree a generic API with the same interface makes sense, I'm going to work on that. Thanks for the heads up about copy-paste-dev. I'll update this PR or close it when needed.

NoahCarnahan:
@millerdev
I've updated the PR to revert the API changes. We want to alter the event names a little though. I'll keep the "don't merge" tag until the hq repo changes are ready.

millerdev:
:+1:

NoahCarnahan:
@millerdev the related analytics changes are ready, so if you could build a new version of vellum with this PR in it for today's deploy that would be great! Thanks.

emord:
@NoahCarnahan it will be deployed today

PR #403, opened 2015-05-06
Fix media capture question type changer

millerdev:
http://manage.dimagi.com/default.asp?160006

cc @NoahCarnahan

### DIFF
--- a/src/mugs.js
+++ b/src/mugs.js
@@ -734,45 +734,23 @@ define([
         tagName: 'upload',
         icon: 'fcc fcc-fd-audio-capture',
         isODKOnly: true,
+        mediaType: "audio/*", /* */
         canOutputValue: false,
         writeCustomXML: function (xmlWriter, mug) {
-            var mediaType = mug.p.mediaType;
-            if (mediaType) {
-                xmlWriter.writeAttributeString("mediatype", mediaType);
-            }
+            xmlWriter.writeAttributeString("mediatype", mug.options.mediaType);
         },
-        init: function (mug, form) {
-            mug.p.mediaType = "audio/*"; /* */
-        },
-        spec: {
-            mediaType: {
-                lstring: 'Media Type',
-                visibility: 'visible',
-                presence: 'required'
-            }
-        }
     });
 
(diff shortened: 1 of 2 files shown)


src/mugs.js
```diff
@@ -734,45 +734,23 @@ define([
...
-        init: function (mug, form) {
-            mug.p.mediaType = "audio/*"; /* */
+            xmlWriter.writeAttributeString("mediatype", mug.options.mediaType);
         },
-        spec: {
-            mediaType: {
-                lstring: 'Media Type',
-                visibility: 'visible',
-                presence: 'required'
-            }
```
millerdev: Strangely, this property was not visible when I tested it, and I'm not sure why. However, the parser [does not support anything other than `audio/*`, `image/*`, and `video/*`](https://github.com/dimagi/Vellum/blob/160006-image-type-change/src/parser.js#L447-L462). It seems fine to move this to mug options (which are static) since setting custom media type values would cause the form to be un-parse-able.

src/mugs.js
```diff
@@ -734,45 +734,23 @@ define([
...
-        }
     });
 
     var Image = util.extend(Audio, {
```
emord: Offtopic: seems weird for Image to inherit from audio

millerdev: Agreed. There should probably be a `Binary` base type that all of the media types inherit from.

src/mugs.js
```diff
@@ -734,45 +734,23 @@ define([
         tagName: 'upload',
         icon: 'fcc fcc-fd-audio-capture',
         isODKOnly: true,
+        mediaType: "audio/*", /* */
```
NoahCarnahan: @millerdev Just curious, do the empty comments (`/* */`) signify something?

millerdev: https://github.com/dimagi/Vellum/blob/031d0d06a0b3c38e1e0d61389fd0c960f51a1368/src/parser.js#L447

NoahCarnahan: Ah ha :)

PR #404, opened 2015-05-06
Support attachments

emord:
This adds attachments to save to a case. Follows the pattern set by the other widgets. Could probably be cleaned up, but I think it would be more reasonable to wait until we  do case management in the form builder.

enchantress @czue

### DIFF
--- a/src/saveToCase.js
+++ b/src/saveToCase.js
@@ -9,6 +9,7 @@ define([
     'vellum/widgets',
     'tpl!vellum/templates/widget_update_case',
     'tpl!vellum/templates/widget_index_case',
+    'tpl!vellum/templates/widget_attachment_case',
     'tpl!vellum/templates/widget_save_to_case',
     'vellum/core'
 ], function (
@@ -22,6 +23,7 @@ define([
     widgets,
     widget_update_case,
     widget_index_case,
+    widget_attach_case,
     widget_save_to_case
 ){
     function createsCase(mug) {
@@ -38,6 +40,10 @@ define([
 
     function indexesCase(mug) {
         return mug ? mug.p.use_index : false;
+    }
+
+    function attachmentCase(mug) {
+        return mug ? mug.p.use_attachment : false;
     }
 
     function addSetValue(mug) {
@@ -130,6 +136,25 @@ define([
                         calculate: $pair.find('.fd-index-property-source').val(),
                         case_type: $pair.find('.fd-index-property-case-type').val(),
                         relationship: $pair.find('.fd-index-property-relationship').val(),
+                    };
+                });
+                return currentValues;
+            };
+
+            return widget;
+        },
+        attachmentCaseWidget = function (mug, options) {
+            options.template = widget_attach_case;
+            var widget = propertyWidget(mug, options);
+
+            widget.getValue = function () {
+                var currentValues = {};
+                _.each(widget.input.find('.fd-attachment-property'), function (kvPair) {
+                    var $pair = $(kvPair);
+                    currentValues[$pair.find('.fd-attachment-property-name').val()] = {
+                        calculate: $pair.find('.fd-attachment-property-source').val(),
+                        from: $pair.find('.fd-attachment-property-from').val(),
+                        name: $pair.find('.fd-attachment-name').val(),
                     };
                 });
                 return currentValues;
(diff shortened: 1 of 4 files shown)


src/saveToCase.js
```diff
@@ -40,6 +42,10 @@ define([
         return mug ? mug.p.use_index : false;
     }
 
+    function attachmentCase(mug) {
+        return mug ? mug.p.use_attachment : false;
```
millerdev: underscore_naming? ... ah, do I spot the precedent four lines above? how did that sneak in there?

src/saveToCase.js
```diff
@@ -40,6 +42,10 @@ define([
         return mug ? mug.p.use_index : false;
     }
 
+    function attachmentCase(mug) {
```
millerdev: Minor: what do you think of renaming this to `isAttachmentCase`?

millerdev: I see now that would be inconsistent with the other functions that do similar boolean checks.

src/saveToCase.js
```diff
@@ -568,6 +687,25 @@ define([
                         return;
                     }
                 }
+                
+                var attachmentRegex = /\/case\/attachment\/(\w+)\/@src/,
```
millerdev: Is this expected to match at the end of the string only? If yes, you probably want to add `$`.

src/saveToCase.js
```diff
@@ -568,6 +687,25 @@ define([
...
+                    attachRet = path.match(attachmentRegex);
+                if (attachRet) {
+                    basePath = path.replace(attachmentRegex, "");
+                    mug = form.getMugByPath(basePath);
+                    if (mug && mug.__className === "SaveToCase") {
+                        var attachProperties = mug.p.attachment_property,
+                            nodeName = attachRet[1];
+                        if (!attachProperties) {
+                            attachProperties = {};
```
millerdev: I think you may also want to set `mug.p.attachment_property` here. Otherwise it seems that you'll get a type error below when you try to set its properties. Might want to add a test for that logic path.

czue:
(i'm gonna assume I have nothing useful to add given that @millerdev has already done a pass, but let me know if you want me to look closer at this)

src/saveToCase.js
emord: Hmm. I'm not sure this is actually a necessary check here. The data tree should have already parsed and initialized `attachmentProperty`. If it isn't in the data tree then a warning should pop up that there was a bind node without a data component right?

src/saveToCase.js
emord: Well that was embarassing. Not sure what came over me.

I'm keeping `date_modified` and similar items as that's how they appear in the case xml. Do you think that is reasonable. I'm only keeping them there where there is a one-to-one mapping from vellum to xml

millerdev: Seems reasonable.

src/saveToCase.js
```diff
@@ -568,6 +692,25 @@ define([
...
+                var attachmentRegex = /\/case\/attachment\/(\w+)\/@src$/,
+                    attachRet = path.match(attachmentRegex);
+                if (attachRet) {
+                    basePath = path.replace(attachmentRegex, "");
+                    mug = form.getMugByPath(basePath);
+                    if (mug && mug.__className === "SaveToCase") {
+                        var attachProperties = mug.p.attachmentProperty,
+                            nodeName = attachRet[1];
+                        if (!attachProperties) {
+                            attachProperties = {};
```
millerdev: This line creates an object that is only bound to a local name, correct? Is the purpose of this to prevent the next `if` statement from throwing a type error? Even if it does, does it matter that `attachProperties[nodeName] = {};` will not modify `mug.p.attachmentProperty` if this branch is evaluated?

tests/saveToCase.js
```diff
@@ -89,6 +91,47 @@ define([
             util.assertXmlEqual(call("createXML"), INDEX_PROPERTY_XML);
         });
 
+        it("should load and save a attachment property", function () {
+            util.loadXML(ATTACHMENT_PROPERTY_XML);
+            var attach = util.getMug("save_to_case");
+            assert.equal(attach.p.useAttachment, true);
+            assert(_.isEqual(attach.p.attachmentProperty, {
```
millerdev: Is this any different than `assert.deepEqual`? (no action needed)

PR #406, opened 2015-05-08
Supply: Receive/Dispense question types

millerdev:
https://trello.com/c/HYmHNRua

cc @kaapstorm

### DIFF
--- a/src/commtrack.js
+++ b/src/commtrack.js
@@ -299,7 +299,7 @@ define([
                         if (mug.p.src.value) {
                             return 'pass';
                         }
-                        return 'Receive must have a Source Case.';
+                        return 'Dispense must have a Source Case.';
                     },
                 },
                 dest: { presence: "notallowed" },
--- a/src/commtrack.js
+++ b/src/commtrack.js
@@ -292,7 +292,7 @@ define([
         }),
         dispenseMugOptions = util.extend(transferMugOptions, {
             typeName: 'Dispense',
-            icon: 'fcc fcc-fd-door',
+            icon: 'icon-signout',
             spec: {
                 src: {
                     validationFunc: function (mug) {
@@ -307,7 +307,7 @@ define([
         }),
         receiveMugOptions = util.extend(transferMugOptions, {
             typeName: 'Receive',
-            icon: 'fcc fcc-fd-inbox',
+            icon: 'icon-signin',
             spec: {
                 src: { presence: "notallowed" },
                 dest: {
--- a/src/commtrack.js
+++ b/src/commtrack.js
@@ -22,6 +22,21 @@ define([
         LEDGER_INSTANCE_ID = "ledger",
         LEDGER_INSTANCE_URI = "jr://instance/ledgerdb",
         nextId = 0,
+        transferValues = [
+            {
+                attr: "entryId",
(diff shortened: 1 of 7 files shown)


src/commtrack.js
```diff
@@ -263,61 +279,85 @@ define([
...
+        dispenseMugOptions = util.extend(transferMugOptions, {
+            typeName: 'Dispense',
+            icon: 'fcc fcc-fd-door',
+            spec: {
+                src: {
+                    validationFunc: function (mug) {
+                        if (mug.p.src.value) {
+                            return 'pass';
+                        }
+                        return 'Receive must have a Source Case.';
```
czue: I think this should say "dispense"

src/commtrack.js
```diff
@@ -263,61 +279,85 @@ define([
...
+                    },
+                },
+                dest: { presence: "notallowed" },
```
czue: assume this hides it from the UI?

src/commtrack.js
```diff
@@ -263,61 +279,85 @@ define([
...
+            types.normal.Transfer = transferMugOptions;
+            types.normal.Dispense = dispenseMugOptions;
+            types.normal.Receive = receiveMugOptions;
             return types;
         },
+        parseDataElement: function (form, el, parentMug, role) {
+            var tag = el.nodeName;
+            if (!role && (tag === "transfer" || tag === "balance") &&
+                    $(el).attr("xmlns") === LEDGER_XMLNS) {
+                if (tag === "transfer") {
```
czue: out of curiosity is possible to change between these question types or you would have to manually migrate all the properties over if you wanted to add a src/dest to a single-sided transaction?

czue:
looked good overall but I think @emord should probably do a full review

czue:
thanks for banging this out!

src/commtrack.js
millerdev: correct

src/commtrack.js
millerdev: you can. I added that in d57ea083c026656b2b9e3ac0ee1c07449840c772

millerdev: there are even tests for it [here](https://github.com/dimagi/Vellum/commit/d57ea083c026656b2b9e3ac0ee1c07449840c772#diff-ff401b6488cc00a539c9523227c94c15R314) and [here](https://github.com/dimagi/Vellum/commit/d57ea083c026656b2b9e3ac0ee1c07449840c772#diff-905dcd5f60ed0d9e12b350401d356a0aR416)

millerdev:
Possibly adding/changing icons yet

src/commtrack.js
```diff
@@ -89,21 +100,11 @@ define([
...
-                    ],
-                }),
-                logicSection
-            ]
+            Transfer: [basicSection, logicSection],
+            Dispense: [basicSection, logicSection],
+            Receive: [basicSection, logicSection],
         },
-        baseTransactionOptions = {
+        baseTransactionOptions = util.extend(mugs.defaultOptions, {
```
emord: Side note: Maybe we could change this to `util.deepExtend` so as to not confuse it with `_.extend`

millerdev: good idea

PR #409, opened 2015-05-11
Replace label elements with divs

orangejenny:
The label elements in vellum aren't really labels in the HTML sense, which is a minor semantic quibble, but being labels makes them appear clickable, and then clicking them does nothing. Swapped them for divs.

### DIFF
--- a/src/templates/control_group.html
+++ b/src/templates/control_group.html
@@ -1,4 +1,4 @@
 <div class="control-group">
-    <label class="control-label"><%=label%></label>
+    <div class="control-label"><%=label%></div>
     <div class="controls"></div>
 </div>
--- a/src/templates/control_group_stdInput.html
+++ b/src/templates/control_group_stdInput.html
@@ -1,8 +1,8 @@
 <div class="control-group">
-    <label class="control-label">
+    <div class="control-label">
         <%=label%>
         <div class="controls">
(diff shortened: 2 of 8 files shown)


emord:
Good with me. How much work would it take to make them real labels that work with the click?

orangejenny:
@emord Not a lot: the labels would need `for` attributes, and the checkboxes would need ids (looks like we only use names). I was thinking it might be weird to use labels, because they're only relevant to checkboxes, so usually that left-hand text would be inert and then once in a while it'd be useful. But that might be laziness talking. Thoughts?

emord:
Oh I thought you said there may be accessibility gains using labels correctly.

Looking at horizontal forms: http://getbootstrap.com/2.3.2/base-css.html#forms it does look like clicking the label selects the appropriate input.

If you think it's not a big gain, I'm good with merging now.

orangejenny:
@emord It is better for accessibility, I was being lazy because Vellum has so few checkboxes. But I just realized that HQ does the same pattern (labels for left-hand side of property lists, regardless of input type), and HQ has way more checkboxes, so the clicking behavior is more useful there (although it interferes with making help bubbles pop up on click instead of on hover...). Closing this PR until I figure out what to do with HQ.

PR #407, opened 2015-05-08
Show help on click and link at bottom of text

emord:
@orangejenny I'm pulling this from the markdown work we did yesterday into this PR. I think you talked to @amsagoff already correct?

This changes the help bubbles in vellum to have a see more link at the end of the help text and the button now shows up once you click it instead of hover. 

cc @czue

PR #410, opened 2015-05-11
Markdown

emord:
Paired with @orangejenny on this.

Should no longer have any of the changes of #407 in it.

Should put this through a UAT round.

cc: @czue


src/templates/markdown_help.html
```diff
@@ -0,0 +1,7 @@
+<a href="#" class="turn-markdown-off markdown-trigger">turn off markdown</a>
+<a href="#" class="turn-markdown-on markdown-trigger">turn on markdown</a>
+<a href="#" class="fd-help"
+            data-title=<%-title%>
+            data-content="CommCare supports a subset of the Markdown annotation language <br><br><a target='_blank' href='https://confluence.dimagi.com/display/TDH/Markdown'> See more</a>"
```
emord: Link will need to change

src/javaRosa.js
```diff
@@ -1055,6 +1054,85 @@ define([
...
+            parent.toggleClass("has-markdown");
+        };
+
+        markdownOutput = $('<div>').addClass("controls well markdown-output");
+
+        widget.handleChange = function() {
+            super_handleChange();
+            var val = widget.getValue(),
+                item = this.getItextItem();
+            if (/[~*#[\]]+|^-/m.test(val)) {
```
czue: maybe extract this test to a function (i think the intent is `looksLikeMarkdown(val)`?) so it's clear what the intent is.

czue: guessing the regex might get expanded in the future.

src/javaRosa.js
```diff
@@ -1055,6 +1054,85 @@ define([
...
+                val = widget.getValue();
+
+            elem.detach('.markdown-output');
+            elem.append(markdownOutput);
+            elem.find('.control-label').append(markdown_help({title:options.lstring }));
+
+            markdownOff = elem.find('.turn-markdown-off').click(function() {
+                wantsMarkdown = false;
+                var item = widget.getItextItem();
+                item.hasMarkdown = false;
```
czue: stylistically it seems cleaner to just combine these two items and get rid of the variable.

tests/markdown.js
```diff
@@ -0,0 +1,150 @@
+/*jshint multistr: true */
```
czue: :+1:

czue:
took a cursory pass - generally looks good to me! will let @millerdev or @orangejenny do a full review and merge.

src/javaRosa.js
```diff
@@ -1055,6 +1054,85 @@ define([
...
+                return false;
+            });
+
+            if (widget.getItextItem().hasMarkdown) {
+                parent.addClass("has-markdown");
+            }
+            else {
+                parent.addClass("markdown-ignorant");
+            }
+            if (/[-~*#[\]]+/.test(val)) {
```
millerdev: Is it intentional that this regex is different than the one above (currently line 1077)? I like @czue's idea of extracting it to a function.

src/javaRosa.js
```diff
@@ -1613,6 +1694,13 @@ define([
                             xmlWriter.writeXML(xml.normalize(val));
                             xmlWriter.writeEndElement();
                         }
+                        if (writeMarkdown) {
```
millerdev: Would it work to change this to `if (item.hasMarkdown)` and eliminate the local variable?

src/javaRosa.js
```diff
@@ -1613,6 +1694,13 @@ define([
...
+                            val = forms[0].getValueOrDefault(lang);
```
millerdev: `forms[0].getValueOrDefault(lang)` doesn't look right because you don't know which form `form[0]` is, but I could be mistaken. Are you looking for the default form? In that case you probably want `itext.get('default', lang)`, right?

src/util.js
```diff
@@ -300,7 +304,11 @@ define([
                 "XML " + (opts.not ? "should not be equivalent" : "mismatch"));
         return patch;
     };
-        
+
+    that.markdownFull = function (text) {
```
millerdev: What do you think of renaming this to simply `markdown`?

tests/markdown.js
millerdev: wait, why is this needed? typically this was only necessary in old-style tests with inline XML strings.

tests/markdown.js
```diff
@@ -0,0 +1,150 @@
...
+    _,
+    form,
+    MARKDOWN_TEST_XML,
+    SIMPLE_MARKDOWN_XML,
+    SIMPLE_MARKDOWN_NO_CHARS_XML,
+    NO_MARKDOWN_XML,
+    NO_MARKDOWN_STARS_XML
+) {
+
+    // see note about controlling time in formdesigner.lock.js
```
millerdev: is this comment relevant? I don't see any timing related code in this file.

tests/markdown.js
```diff
@@ -0,0 +1,150 @@
...
+            it("should not write any markdown if markdown characters are not input", function() {
+                util.loadXML("");
+                util.addQuestion("Text", 'markdown_question');
+                $('[name=itext-en-label]').val("no markdown").change();
+                util.assertXmlEqual(call('createXML'), NO_MARKDOWN_XML, {normalize_xmlns: true});
+            });
+        });
+
+        describe("when a user explicitly wants no markdown", function() {
+            function beforeFn(done) {
```
millerdev: Could use an anonymous function like `beforeEach(function (done) { ...`

tests/static/markdown/markdown-test.xml
```diff
@@ -0,0 +1,68 @@
...
+			</instance>
+			<bind nodeset="/data/markdown_question" type="xsd:string" />
+			<bind nodeset="/data/no_markdown" type="xsd:string" />
+			<bind nodeset="/data/markdown_but_not" type="xsd:string" />
+			<bind nodeset="/data/escaped_markdown" type="xsd:string" />
+			<itext>
+				<translation lang="en" default="">
+					<text id="markdown_question-label">
+						<value>some markdown</value>
+                      	<value form="markdown">**some markdown**</value>
```
millerdev: inconsistent use of tabs and spaces for indentation

src/javaRosa.js
emord: :+1: expanded and documented much better now

tests/markdown.js
emord: yeah. I just copied this file from javaRosa so I didn't need to do the boilerplate. cleaned this up

src/util.js
emord: was originally based on copy-paste-dev so this was to differentiate with your markdownlite. I suppose that function could be thrown away on copy-paste once these are merged together.

src/javaRosa.js
```diff
@@ -1055,6 +1054,95 @@ define([
...
+        var parent = options.parent;
+        var widget = itextLabelWidget(mug, language, form, options),
+            super_setValue = widget.setValue,
+            super_getUIElement = widget.getUIElement,
+            super_handleChange = widget.handleChange,
+            wantsMarkdown = true,
+            markdownOff, markdownOn, markdownOutput;
+
+        function looksLikeMarkdown(val) {
+            /* Regex checks (in order):
```
emord: Pulled these from: https://confluence.dimagi.com/display/TDH/Markdown

src/util.js
millerdev: Yep, fine to remove `markdownlite` now that we have markdown-it.

millerdev:
:+1:

emord:
This passed QA, but I want to make sure the migration for REC is good to go before releasing

millerdev:
Looks good. Ping when you're ready to merge.

emord:
@millerdev I think it's good to merge now. Added a commit to always accept markdown as the true value and overwrite the default form. I confirmed with a/rec that always accepting markdown works for them

PR #411, opened 2015-05-12
Pre UAT copy paste changes

millerdev:
Three changes, three commits. Merging into copy-paste-dev

cc @kaapstorm

### DIFF
--- a/src/copy-paste.js
+++ b/src/copy-paste.js
@@ -200,7 +200,7 @@ define([
         if (!_.isUndefined(window.analytics)) {
             window.analytics.track("Copy questions in form builder");
         }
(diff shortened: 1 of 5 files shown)


src/copy-paste.js
```diff
@@ -330,7 +330,8 @@ define([
             $(document).on('cut copy paste keydown', function (e) {
                 if (e.type === 'cut' ||
                     e.metaKey && String.fromCharCode(e.keyCode) === 'X') {
-                    onCut(opts);
+                    // Disable cut until undo feature is implemented
+                    if (false) { onCut(opts); }
```
millerdev: Used a never-true conditional because otherwise the `onCut` function caused a jshint warning: 'onCut' is defined but never used.

PR #414, opened 2015-05-13
Setvalue

emord:
@millerdev 

https://docs.google.com/document/d/1B5-fPgW-as-Cfo4vQDMnvLfGlCHkt2GgovT711-10Xc/edit#

Milestone 1 of case management in form builder

set values are now treated like binds, but still uses the old array for set values that aren't associated with a mug/have weird event types

cc @czue

### DIFF
--- a/src/commtrack.js
+++ b/src/commtrack.js
@@ -177,7 +177,7 @@ define([
             },
             getSetValues: function (mug) {
                 var path = mug.absolutePath,
-                    event = mug.isInRepeat(mug) ? "jr-insert" : "xforms-ready",
+                    event = mug.isInRepeat() ? "jr-insert" : "xforms-ready",
                     ret = [];
 
                 _.each(setvalueData[mug.__className], function (data) {
@@ -340,7 +340,7 @@ define([
         setValuePaths = _.chain(setvalueData)
             .map(function (mugSetValues, mugClass) {
                 return _.map(mugSetValues, function(attrs) {
-                    return attrs.path;
+                    return RegExp.escape(attrs.path);
                 });
             }).flatten().uniq().value(),
         setValueDataRegex = new RegExp("/(" + setValuePaths.join('|') + ")$");
@@ -405,10 +405,10 @@ define([
                 var basePath = path.replace(setValueDataRegex, "");
                 if (path !== basePath) {
                     mug = form.getMugByPath(basePath);
-                    var setValue = _.find(setvalueData[mug.__className], function(sv) {
-                        return sv.path === path.match(setValueDataRegex)[1];
-                    }).attr;
                     if (isTransaction(mug)) {
+                        var setValue = _.find(setvalueData[mug.__className], function(sv) {
+                            return sv.path === path.match(setValueDataRegex)[1];
+                        }).attr;
                         mug.p[setValue] = el.attr("value");
                         return;
                     }
@@ -424,9 +424,6 @@ define([
             }
             return this.__callOld();
         },
-        loadXML: function () {
-            this.__callOld();
-        }
     });
 
     function isTransaction(mug) {
--- a/src/commtrack.js
+++ b/src/commtrack.js
@@ -155,7 +155,7 @@ define([
                     lstring: 'Product',
                     visibility: 'visible',
                     presence: 'optional',
-                    widget: setValueWidget,
+                    widget: widgets.xPath,
                     xpathType: "generic",
                     help: 'A reference to a product ID, e.g., "/data/products/current_product"',
                 },
@@ -175,13 +175,30 @@ define([
                     lstring: 'Display Condition'
                 },
             },
+            getSetValues: function (mug) {
+                var path = mug.absolutePath,
+                    event = mug.isInRepeat(mug) ? "jr-insert" : "xforms-ready",
+                    ret = [];
+
+                _.each(setvalueData[mug.__className], function (data) {
+                    var value = mug.p[data.attr],
+                        ref = path + "/" + data.path;
+                    if (value) {
+                        ret.push({
+                            event: event,
+                            ref: ref,
+                            value: value
+                        });
+                    }
+                });
+
+                return ret;
+            },
         }),
         balanceMugOptions = util.extend(baseTransactionOptions, {
             typeName: 'Balance',
             getTagName: function () { return "balance"; },
             getExtraDataAttributes: function (mug) {
-                // HACK must happen before <setvalue> and "other" <instance> elements are written
-                prepareForWrite(mug);
                 var attrs = mug.p.rawDataAttributes || {};
                 return {
                     xmlns: LEDGER_XMLNS,
@@ -193,11 +210,11 @@ define([
             },
             icon: 'fcc fcc-fd-hash',
             init: function (mug, form) {
-                mug.p.entityId = {value: ""};
+                mug.p.entityId = "";
                 mug.p.sectionId = "";
-                mug.p.entryId = {value: ""};
+                mug.p.entryId = "";
                 mug.p.quantity = "";
-                mug.p.date = {value: "today()"};
+                mug.p.date = "today()";
                 addLedgerDBInstance(mug, form);
             },
             spec: {
@@ -206,7 +223,7 @@ define([
                     lstring: 'Case',
                     visibility: 'visible',
                     presence: 'optional',
-                    widget: setValueWidget,
+                    widget: widgets.xPath,
                     xpathType: "generic",
                     help: 'XPath expression for the case ID associated with this balance.',
                 },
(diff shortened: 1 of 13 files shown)


src/commtrack.js
```diff
@@ -377,20 +392,23 @@ define([
...
             this.__callOld();
         },
-        handleMugParseFinish: function (mug) {
-            this.__callOld();
-            if (!isTransaction(mug)) {
-                return;
+        parseSetValue: function (form, el, path) {
+            var mug = form.getMugByPath(path);
+            if (!mug) {
+                var regex = /\/(entry\/@id|@entity-id|@date|@src|@dest)$/,
```
emord: I should autogenerate this regex from `setValueData`

amsagoff: OMG FIRST CASE MANAGEMENT PR

src/commtrack.js
```diff
@@ -282,7 +297,7 @@ define([
                     lstring: 'Destination Case',
                     visibility: 'visible',
                     presence: 'required',
-                    widget: setValueWidget,
+                    widget: widgets.xPath,
                     xpathType: "generic",
                     help: 'XPath expression for the case ID receiving the transaction. Leave blank if unknown or not applicable.',
```
millerdev: Not related to this PR, but I just realized the "Leave blank if unknown..." part here should be removed since these are now required for the Transfer question type. Same above.

src/commtrack.js
```diff
@@ -314,14 +329,21 @@ define([
...
                         return 'Receive must have a Destination Case.';
                     },
                 },
             }
-        });
+        }),
+        setValuePaths = _.chain(setvalueData)
+            .map(function (mugSetValues, mugClass) {
+                return _.map(mugSetValues, function(attrs) {
+                    return attrs.path;
```
millerdev: Not sure it matters, but might want to do `RegExp.escape(attrs.path)` to be safe.

src/commtrack.js
```diff
@@ -377,20 +399,22 @@ define([
...
-                return;
+        parseSetValue: function (form, el, path) {
+            var mug = form.getMugByPath(path);
+            if (!mug) {
+                var basePath = path.replace(setValueDataRegex, "");
+                if (path !== basePath) {
+                    mug = form.getMugByPath(basePath);
+                    var setValue = _.find(setvalueData[mug.__className], function(sv) {
+                        return sv.path === path.match(setValueDataRegex)[1];
+                    }).attr;
```
millerdev: Could avoid the find for non-transaction widgets with paths that happen to match `setValueDataRegex` by moving the above 3 lines into the `if` block below.

src/commtrack.js
```diff
@@ -402,9 +426,6 @@ define([
         },
         loadXML: function () {
             this.__callOld();
-            this.data.core.form.on("question-remove", function (event) {
-                dropSetValues(event.mug);
-            });
         }
```
millerdev: Looks like you can remove this entire method since it's only doing a super call now.

src/commtrack.js
```diff
@@ -175,13 +175,30 @@ define([
                     lstring: 'Display Condition'
                 },
             },
+            getSetValues: function (mug) {
+                var path = mug.absolutePath,
+                    event = mug.isInRepeat(mug) ? "jr-insert" : "xforms-ready",
```
millerdev: Could `mug.isInRepeat(mug)` be changed to simply `mug.isInRepeat()`?

src/mugs.js
```diff
@@ -649,6 +662,20 @@ define([
             return attrs.nodeset ? [attrs] : [];
         },
 
+        getSetValues: function (mug) {
+            var ret = [];
+
+            if (mug.p.setValue) {
+                ret = [{
+                    value: mug.p.setValue,
+                    event: mug.isInRepeat() ? 'jr-insert' : 'xforms-ready',
```
millerdev: Could this change the event if, for example, the form originally had `<setvalue event="jr-insert" .../>` for this mug but it was not in a repeat?

src/parser.js
```diff
@@ -70,13 +70,8 @@ define([
...
             instances = _getInstances(xml),
-            data = $(instances[0]).children();
-
-        xml.find('setvalue').each(function () {
-            var $this = $(this);
-            form.addSetValue(
-                $this.attr('event'), $this.attr('ref'), $this.attr('value'));
-        });
+            data = $(instances[0]).children(),
+            setValues = xml.find('setvalue');
```
millerdev: I know you have not changed it here, but this `find` always felt way too liberal to me. Do you think this might be a reasonable time to change it to, for example, only find setvalue nodes that are immediate children of the `<model>` node? The same applies to finding `bind` nodes above. It seems that these two would incorrectly parse the form if anyone ever had a question ID of `bind` or `setvalue`.

Actually, let's not change that now. No need to possibly introduce more instability in a large-ish changeset like this. It's a good thing to keep in mind though. Curious to hear what @ctsims thinks too.

src/parser.js
```diff
@@ -221,7 +218,31 @@ define([
...
+        var mug = form.getMugByPath(path),
+            event = el.attr('event'),
+            ref = el.attr('ref'),
+            value = el.attr('value');
+
+        // HACK: hardcoding these as that's what setValue will support for now
+        if (!mug || (event !== 'xforms-ready' && event !== 'jr-insert')) {
+            form.addSetValue(event, ref, value);
+        } else {
+            mug.p.setValue = value;
```
millerdev: For some mug types setValue is presence is notallowed. This could cause a validation error that the user cannot fix. Instead of setting the spec to `{presence: 'notallowed'}`, I suggest setting it to `{presence: 'optional', visibility: 'hidden'}`. I'm 70% confident that should not cause problems elsewhere...

src/writer.js
```diff
@@ -185,6 +179,29 @@ define([
...
+        });
+
+        dataTree.walk(function (mug, nodeID, processChildren) {
+            if(mug && mug.options.getSetValues) {
+                _.each(mug.options.getSetValues(mug), function(setValue) {
+                    writeSetValue(setValue);
+                });
+            }
+            processChildren();
+        });
```
millerdev: What you're doing with setvalues in this PR is looking really sweet, but there's a problem here. Setvalue nodes are evaluated in document order. The parser is not keeping track of, and the writer is not preserving the order in which setvalue nodes were originally parsed from the form. That means opening a form and saving it could break it because setvalue nodes would be written out in a different order.

I don't see an easy solution to this problem with the approach you've taken here. I wonder what @ctsims thinks about writing setvalues in a different order than they were parsed? Is it a major issue?

tests/static/setvalue/set-value-special.xml
```diff
@@ -0,0 +1,27 @@
...
+	<h:head>
+		<h:title>Untitled Form</h:title>
+		<model>
+			<instance>
+				<data xmlns:jrm="http://dev.commcarehq.org/jr/xforms" xmlns="http://openrosa.org/formdesigner/85BAD5DB-3ECE-447C-A29B-6A92C074B1E2" uiVersion="1" version="1" name="Untitled Form">
+					<text />
+				</data>
+			</instance>
+			<bind nodeset="/data/text" type="xsd:string" />
+			<setvalue event="special-snowflake" ref="/data/text" value="blah" />
```
millerdev: your setvalue is not a special snowflake ... just had to say that

czue:
i'm not even going to bother trying to review this. keep up the good work @emord  and @millerdev! you're both special snowflakes.

src/mugs.js
emord: [This](http://dimagi.github.io/xform-spec/#action-events) makes me think that jr-inserts are only valid for that repeats, so maybe we shouldn't care?

src/parser.js
emord: Hmm. I think it would be a good idea to change it to find immediate children, but can wait until we hear from @ctsims 

Maybe we could warn on certain node ids like bind or setvalue (I think you had a recent PR that did that)?

src/writer.js
emord: Ah. From an earlier conversation I thought we didn't care about ordering at the moment.

Easy solution would be to use `vellum:ignore="retain"` if you really care.

We could add another input that had a priority associated with it.  The user could specify priority and we sort on that order. That sounds messy and hard to keep track of though, so I am against that. How common are problems relating  to the ordering of setvalues

src/parser.js
emord: Fixed this in those three places

src/parser.js
millerdev: -1 for warning. Let's do correct form parsing instead. I did have a PR that did something like that, it's implemented on copy-paste-dev, but it's a dirty hack workaround for poor use of namespaces.

src/writer.js
millerdev: `vellum:ignore="retain"` will probably not work either since it now creates `Ignored` mugs for data, bind, and control elements. I suppose it should probably be extended to handle setvalue nodes to conform with this PR.

The previous implementation (before this PR) preserved the order of setvalue nodes, so I don't believe we've had many problems (caused by vellum) related to ordering in the past.

emord:
Closing for now. Going to rebase on staging

PR #416, opened 2015-05-15
Fix no placeholder

emord:
https://github.com/dimagi/Vellum/pull/396#issuecomment-97542601

elf: @czue

### DIFF
--- a/src/javaRosa.js
+++ b/src/javaRosa.js
@@ -892,6 +892,9 @@ define([
             var itextItem = widget.getItextItem();
             if (itextItem) {
                 if (widget.isDefaultLang) {
+                    if (!value) {
+                        value = widget.getPlaceholder();
+                    }
                     widget.mug.fire({
                         type: 'defaultLanguage-itext-changed',
                         form: widget.form,
--- a/src/javaRosa.js
+++ b/src/javaRosa.js
@@ -1609,12 +1609,14 @@ define([
                         for (var k = 0; k < forms.length; k++) {
                             form = forms[k];
                             val = form.getValueOrDefault(lang);
-                            xmlWriter.writeStartElement("value");
-                            if(form.name !== "default") {
-                                xmlWriter.writeAttributeString('form', form.name);
+                            if (val) {
+                                xmlWriter.writeStartElement("value");
+                                if(form.name !== "default") {
+                                    xmlWriter.writeAttributeString('form', form.name);
+                                }
+                                xmlWriter.writeXML(xml.normalize(val));
+                                xmlWriter.writeEndElement();
                             }
-                            xmlWriter.writeXML(xml.normalize(val));
-                            xmlWriter.writeEndElement();
                         }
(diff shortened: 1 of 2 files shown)


czue:
does this mean that if I load a form with an empty label and then save it it will change it to the placeholder? that could be confusing if people wanted blankness.

i'll let @millerdev review and merge but :+1: on code changes, just hard for me to parse behavior or know wider implications.

src/javaRosa.js
```diff
@@ -1606,12 +1609,14 @@ define([
                         for (var k = 0; k < forms.length; k++) {
                             form = forms[k];
                             val = form.getValueOrDefault(lang);
-                            xmlWriter.writeStartElement("value");
-                            if(form.name !== "default") {
-                                xmlWriter.writeAttributeString('form', form.name);
+                            if (val) {
```
millerdev: Is there ever a case where `val` evaluates to false, and therefore the `<value></value>` tag is not written? That was the bug we hit [last time](https://github.com/dimagi/Vellum/commit/ccd3475d298337f954034c4e1d8d37ce8e04dfbd), right?

src/javaRosa.js
```diff
@@ -892,6 +892,9 @@ define([
             var itextItem = widget.getItextItem();
             if (itextItem) {
                 if (widget.isDefaultLang) {
+                    if (!value) {
+                        value = widget.getPlaceholder();
+                    }
```
millerdev: Having this inside the `isDefaultLang` block seems a little wierd since it will change `value` for https://github.com/dimagi/Vellum/blob/fix-no-placeholder/src/javaRosa.js#L906, which is outside of the if-block. Does that seems like a problem to you?

millerdev:
Re: "if people wanted blankness"

I've been thinking about that use case, and it doesn't seem to be very well supported. The placeholder always appears when there is no text in the field, so it will be misleading at best if we allow the user to clear the field to get a blank value. I'd recommend something different than placeholders if we want to support that.

Currently you can enter a single space in the label to make it appear blank, but that's a hack.

src/javaRosa.js
emord: That should change. The itext form is where the writer will look when writing that value: https://github.com/dimagi/Vellum/blob/fix-no-placeholder/src/javaRosa.js#L1610-L1619

This is a bit of a corner case though, as this only happens whenever we have one language.

src/javaRosa.js
emord: Yeah. The test that I changed here was the same test I added in that PR. The only time it should evaluate to false is when it is `""`. I haven't seen a case where an itext form would be false, undefined, or null. Cases where a form shouldn't be written are handled by removing it from the list of forms.

czue:
cool. agree it sounds like it's not a supported feature yet, so seems fine to not worry about it for now

millerdev:
Tests passed locally.

PR #417, opened 2015-05-19
Revive advanced itemsets

emord:
Revived this from before itemset-connector-2 was merged. Uses the same tests as well

@millerdev elf: @czue

### DIFF
--- a/src/itemset.js
+++ b/src/itemset.js
@@ -45,7 +45,7 @@ define([
     debug
 ) {
     var mugTypes = mugs.baseMugTypes.normal,
-        Itemset,
+        Itemset, AdvancedItemset,
         END_FILTER = /\[[^\[]*\]$/;
 
     Itemset = util.extend(mugs.defaultOptions, {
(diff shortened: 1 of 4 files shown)


tests/options.js
```diff
@@ -102,6 +102,7 @@ define(function () {
         features: {
             'group_in_field_list': true,
             'help_markdown': true
+            // 'advanced_itemsets': true
```
emord: Left in here to know what to expect

src/itemset.js
```diff
@@ -180,7 +222,8 @@ define([
         },
         updateControlNodeAdaptorMap: function (map) {
             this.__callOld();
-            var adaptItemset = parser.makeControlOnlyMugAdaptor('Itemset');
+            var adaptItemset = parser.makeControlOnlyMugAdaptor('Itemset'),
+                use_advanced_itemsets = this.opts().features.advanced_itemsets;
```
emord: Annoyingly the places where I need to query this value are called before the init function: https://github.com/dimagi/Vellum/blob/master/src/base.js#L127-L149 so couldn't use a global

millerdev: `updateControlNodeAdaptorMap` is called by `parser.init()`, which is called by `core.postInit`, and that should be called after all plugins have been initialized. So I'm not sure why this particular one is a problem.

On the other hand, `getMugTypes` is called by `core.init()`, which is called before any other plugins are initialized. That one is a problem for you here. It may be worth investigating if we can replace [these lines](https://github.com/dimagi/Vellum/blob/f93eb0f304c088e976437dde702e953a519e278c/src/base.js#L144-L149) with simply `instance.init()`, which would init all other plugins before core (in the reverse order from how they are currently initialized)—this seems pretty elegant to me and may be worth some effort. Another option would be to move the `getMugTypes` call from `core.init()` to `core.postInit()`.

millerdev:
I think this is a reasonable stop-gap fix, but it makes the nice new itemsets selector unavailable to any domain that has the advanced itemsets feature flag enabled. I'm going to see if I can spike something quick so we don't have to take such a big step backward when that flag is enabled.

PR #418, opened 2015-05-19
Add advanced editor to fixture widget (flagged)

millerdev:
screenshot 2015-05-19 16 30 59

Spike on https://github.com/dimagi/Vellum/pull/417

@emord 
cc @kaapstorm

### DIFF
--- a/src/datasources.js
+++ b/src/datasources.js
@@ -232,6 +232,11 @@ define([
         return widget;
     }
 
+    /**
+     * @param options - Optionally pass in:
+     *      getSource(mug) - Initializes the source for the advanced editor
+     *      setSource(source, mug) - Saves the source from the advanced editor
+     */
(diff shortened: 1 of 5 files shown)


emord:
Looks good to me. What do you think about disabling the filter widget whenever the lookup table isn't found? It seems really nasty that the last filter will be taken out of the query and inserted there, which is part of the reason I went for the stop gap first.

millerdev:
Yeah, was kinda vacillating on that one. Not sure if it's worth the extra complexity only to accommodate advanced use cases. Were you thinking we'd make it tack that filter back on the query when you drop into advanced mode?

emord:
@millerdev Updated to handle the filter nicely. Still doesn't support multiple filters

emord:
g2m after tests

PR #419, opened 2015-05-20
Copy paste QA fixes

millerdev:
- [Fix bug in ordering of pasted questions](http://manage.dimagi.com/default.asp?168666)
- [Warn on paste missing multimedia](http://manage.dimagi.com/default.asp?168667)

cc @sravfeyn

### DIFF
--- a/src/javaRosa.js
+++ b/src/javaRosa.js
@@ -1665,6 +1665,7 @@ define([
         },
         getMugSpec: function () {
             var spec = this.__callOld(),
+                that = this,
                 databind = spec.databind,
                 control = spec.control;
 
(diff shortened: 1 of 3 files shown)


src/javaRosa.js
```diff
@@ -1743,6 +1747,17 @@ define([
...
                                 if (value) {
                                     item.set(value, form, lang);
+                                    if (mmForms.hasOwnProperty(form) &&
+                                            !objectMap.hasOwnProperty(value)) {
+                                        // TODO make itext widgets display messages
+                                        // and then mug.addMessage(name, ...
+                                        mug.addMessage(null, {
+                                            key: "missing-multimedia-warning",
+                                            level: mug.WARNING,
+                                            message: "MultiMedia was not copied; " +
```
sravfeyn: Translation? (may be excluded as there isn't a neat way to mark translations in vellum code?)

millerdev: yeah, we don't do translations in Vellum yet.

PR #420, opened 2015-05-20
Fix infinite loop in bulk translation updater

millerdev:
http://manage.dimagi.com/default.asp?168791

This bug was found during copy/paste QA, but applies to master as well.

cc @sravfeyn

### DIFF
--- a/src/javaRosa.js
+++ b/src/javaRosa.js
@@ -1117,9 +1117,7 @@ define([
 
         var items = getItextItemsFromMugs(form, true);
         items = _.object(_.map(items, function (item) { return [item.id, item]; }));
-        cells = nextRow();
-        while (cells) {
-            // what's the point of creating items here?
+        for (cells = nextRow(); cells; cells = nextRow()) {
             item = items[cells[0]];
             if (!item) {
                 // TODO alert user that row was skipped
@@ -1135,7 +1133,6 @@ define([
                     }
                 }
             }
-            cells = nextRow();
         }
         Itext.fire("change");
     };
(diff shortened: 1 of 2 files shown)


src/javaRosa.js
```diff
@@ -1117,9 +1117,7 @@ define([
 
         var items = getItextItemsFromMugs(form, true);
         items = _.object(_.map(items, function (item) { return [item.id, item]; }));
-        cells = nextRow();
-        while (cells) {
-            // what's the point of creating items here?
+        for (cells = nextRow(); cells; cells = nextRow()) {
```
orangejenny: curious what the bug is here, old & new code look effectively identical to me

emord: I think it's if it hits https://github.com/dimagi/Vellum/pull/420/files#diff-e5b3eca8407b45b132e0147b8392192cR1124 then it won't increment the row.

g2m when tests are done

millerdev: yeah, the `continue` skipped `cells = nextRow()`, which was previously at the end of the loop.

czue: huh, i did not know that javascript assignment operators returned the value.

millerdev: Interesting comment, @czue. I do not believe I am using that feature here, am I? A more concise way to write this is
```js
while (cells = nextRow()) {
```
but our lint tool doesn't like it (and I think that statement looks pretty misleading), so I opted for the `for` loop instead.

czue: Unless I'm missing something I think you're using the feature. The statement is an assignment, which in python would evaluate to nothing, but in javascript evaluates to the right side of the assignment. Therefore I think both constructs rely on that feature.

millerdev: Not understanding... This is now I understand it:
```
for (<assignment, result ignored>; <test, result used>; <increment, result ignored>) {
```
What am I missing?

czue: sorry you're right. i was confused about the ordering

PR #422, opened 2015-05-21
Document datasources

emord:
@millerdev Did I miss anything? I don't think this changes too much in your generalization branch

@czue A little context from the earlier email

enchantress: @esoergel

### DIFF
--- a/src/datasources.js
+++ b/src/datasources.js
@@ -17,28 +17,32 @@
  *          defaultId: string (used in instance definition)
  *          intialQuery: string (used in nodeset)
  *          name: string (text in the dropdown)
- *          structure: nested dictionary of attributes
+ *          structure: nested dictionary of elements and attributes
  *          {
- *              attribute1: {
+ *              element: {
  *                  structure: {
- *                      inner-attribute1: {
+ *                      inner-element: {
  *                          no_option: boolean (does not show up in the dropdown)
  *                      }
  *                  }
- *              }
+ *              },
+ *              @attribute: { }
  *          }
  *      }
  * ]
  *
+ * elements can be nested indefinitely with structure keys describing inner 
+ *     elements and attributes
+ *
  * The result of that would be (if used in an itemset):
- * <instance src=sourceUri id=defaultId>
- * ...
- * <itemset nodeset=initialQuery />
+ *     <instance src=sourceUri id=defaultId>
+ *     ...
+ *     <itemset nodeset=initialQuery />
  *
  *
  * The dropdown would have options:
- * name (would have output as above)
- * name - attribute1 (would change nodeset to initialQuery/attribute1
+ *     name (would have output as above)
+ *     name - element (would change nodeset to initialQuery/element)
  */
 define([
     'jquery',
--- a/src/datasources.js
+++ b/src/datasources.js
@@ -21,9 +21,7 @@
  *          {
  *              element: {
  *                  structure: {
- *                      inner-element: {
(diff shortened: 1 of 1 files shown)


src/datasources.js
```diff
@@ -1,3 +1,45 @@
...
+ * ]
+ *
+ * The endpoint function (or URL) should return the following format:
+ * [
+ *      {
+ *          sourceUri: string (used in the instance definition)
+ *          defaultId: string (used in instance definition)
+ *          intialQuery: string (used in nodeset)
+ *          name: string (text in the dropdown)
+ *          structure: nested dictionary of attributes
```
millerdev: Since this is describing XML, and in XML "attributes" are different from "elements", it might make more sense to say "nested dictionary describing the data source"

src/datasources.js
```diff
@@ -1,3 +1,45 @@
...
+ *          {
+ *              attribute1: {
```
millerdev: this should say `element` or `element1` since an XML attribute cannot have nested items inside it (that I know of).

src/datasources.js
```diff
@@ -1,3 +1,45 @@
...
+ *                  structure: {
+ *                      inner-attribute1: {
+ *                          no_option: boolean (does not show up in the dropdown)
+ *                      }
```
millerdev: might be clearer to have two examples here:
```
inner-element: {},
@attribute: {}
```

src/datasources.js
```diff
@@ -1,3 +1,45 @@
...
+ *                  }
+ *              }
+ *          }
+ *      }
+ * ]
```
millerdev: Down here below the example, some further documentation would be nice:
- The value of an element may contain `structure: {...}` describing nested elements and/or attributes.

Also, can we drop the `no_option` thing and instead use the absence of a `structure: {...}` key/value pair to mean the same as `no_option: true`?

emord: Hmm. We could drop no_option: true and use the structure test as the default. I think we should still allow it to be overridden

millerdev: I can't think of a time you'd want to use `no_option: true` to override the default (that is, when there is a `structure` present). Do you have an example of when we would use both `structure: ...` and `no_option: ...` on the same element? If not, I feel like `no_option` is complicating the spec without adding value.

emord: Yeah. I can't really think of a great reason to keep it at the moment. My original thought was that we are using a dropdown and if we added cases/locations to that dropdown there would be a lot of noise.

However, before we add those two we should probably change the dropdown to a combobox so that the noise wouldn't matter as much.

I think I just tried to over engineer it.

src/datasources.js
```diff
@@ -1,3 +1,49 @@
...
+ *          sourceUri: string (used in the instance definition)
+ *          defaultId: string (used in instance definition)
+ *          intialQuery: string (used in nodeset)
+ *          name: string (text in the dropdown)
+ *          structure: nested dictionary of elements and attributes
+ *          {
+ *              element: {
+ *                  structure: {
+ *                      inner-element: {
+ *                          no_option: boolean (does not show up in the dropdown)
```
millerdev: Are you keeping this line? I think no need to (`no_option` can be an undocumented feature until we remove it). However, if you are then I suggest changing it to
```
no_option: true (does not show up in the dropdown)
```
since only a value of `true` will cause it to not show up in the dropdown.

emord: Removed until fixed

PR #426, opened 2015-05-26
Ignore filters in itemset values and labels

emord:
http://manage.dimagi.com/default.asp?168793

@esoergel

### DIFF
--- a/src/itemset.js
+++ b/src/itemset.js
@@ -115,11 +115,13 @@ define([
                     var possibleSrcs = _.map(datasources.getPossibleFixtures(), 
                                              function(val) { return val.src; }),
                         notCustom = _.contains(possibleSrcs, itemsetData.instance.src),
-                        choices = datasources.autocompleteChoices(itemsetData.instance.src);
+                        choices = datasources.autocompleteChoices(itemsetData.instance.src),
+                        strippedValue = itemsetData.valueRef.replace(/(\[.+])+/, ""),
+                        strippedLabel = itemsetData.labelRef.replace(/(\[.+])+/, "");
 
-                    if (notCustom && !_.contains(choices, itemsetData.valueRef)) {
+                    if (notCustom && !_.contains(choices, strippedValue)) {
                             return itemsetData.valueRef + " was not found in the lookup table";
-                    } else if (notCustom && !_.contains(choices, itemsetData.labelRef)) {
+                    } else if (notCustom && !_.contains(choices, strippedLabel)) {
                             return itemsetData.labelRef + " was not found in the lookup table";
                     }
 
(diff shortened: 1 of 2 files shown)


src/itemset.js
```diff
@@ -115,11 +115,13 @@ define([
                     var possibleSrcs = _.map(datasources.getPossibleFixtures(), 
                                              function(val) { return val.src; }),
                         notCustom = _.contains(possibleSrcs, itemsetData.instance.src),
-                        choices = datasources.autocompleteChoices(itemsetData.instance.src);
+                        choices = datasources.autocompleteChoices(itemsetData.instance.src),
+                        strippedValue = itemsetData.valueRef.replace(/(\[.+])+/, ""),
+                        strippedLabel = itemsetData.labelRef.replace(/(\[.+])+/, "");
```
millerdev: Not sure if it matters since the case I'm thinking of is probably for advance users only, but will this work correctly with a src like `instance("foo")/foo/bar[a=1]/baz[b=2]`? Does it matter?

emord: Nope. That won't work. It only works for the end filters.

I don't think we care about that as users that do that sort of thing should know what they are doing

orangejenny: To a future dev reading this code, though, that's going to look like a possible bug (that `instance("foo")/foo/bar[a=1]/baz[b=2]` becomes `instance("foo")/foo/bar`). How about `/(\[^\]+])+/` to make it more obvious that you only want to replace filters?

emord: Err crap I think I responded to @millerdev incorrectly anyways.

This is only for the value and label in itemsets, so it doesn't handle instances. https://github.com/dimagi/Vellum/blob/master/tests/static/itemset/dropdown-fixture.xml#L26-L27

So the form there should be attribute[filter][filter]. Is there another use case of value/label i should be thinking of?

millerdev: I think I'm satisfied that this should typically be applied to some relative path like `attribute[filter][filter]`, and not `instance("foo")/foo/bar[a=1]/baz[b=2]`.

The regexp could be improved, although I'm not sure it matters.

Examples with with `/(\[.+])+/`
- `bar[a=1][b=2]` becomes `bar`
- `bar[a=1]/foo[b=2]` becomes `bar`
- `bar[a=1]/foo` becomes `bar/foo`

I think the first two are fine. Possibly more could be done on the second to validate that `bar` contains a `foo`, but probably not important right now. But the third one seems worse, if a bit far out.

Possible alternate: `/\[.+\].*/`
Matches from the first filter to the end of the string. Tries to match at least one well-formed filter, but there are malformed expressions that this will match too. Probably none of these crazy edge cases are worth caring about, although maybe we should decide how to err: do we want to strip possibly malformed filters, or do we want to try to strip well-formed filters only?

emord: I changed the regex to `/\[[^\[]+]/g`

Using your examples:
* `bar[a=1][b=2]` becomes `bar`
* `bar[a=1]/foo[b=2]` becomes `bar/foo`
* `bar[a=1]/foo` becomes `bar/foo`

I think this is desired, because we will want to make sure that bar/foo is an valid attribute in the structure.

PR #427, opened 2015-05-28
At js

emord:
Adding a generic feature flag for experimental UI enhancements that are being done as part of case management.

@millerdev At.js looks like a pretty sweet jquery plugin that can do the things we need for phase 3 (demo here: http://ichord.github.io/At.js/)

output

@esoergel

### DIFF
--- a/src/javaRosa.js
+++ b/src/javaRosa.js
@@ -42,7 +42,8 @@ define([
             'default', 'short', 'long', 'audio', 'video', 'image'
         ],
         _nextItextItemKey = 1,
-        HELP_MARKDOWN;
+        HELP_MARKDOWN,
+        EXPERIMENTAL_UI;
 
     function ItextItem(options) {
         this.forms = options.forms || [];
@@ -837,6 +838,34 @@ define([
             });
 
         if (options.path === 'labelItext') {
+            if (EXPERIMENTAL_UI) {
+                $input.atwho({
+                    at: "#form",
+                    data: _.chain(mug.form.getMugList())
+                           .map(function(mug) {
+                                return {
+                                    id: mug.ufid,
+                                    name: mug.form.getAbsolutePath(mug, true),
+                                    path: mug.absolutePath
+                                };
+                            })
+                            .filter(function(choice) { return choice.name; })
+                            .value(),
+                    displayTpl: '<li>${name}</li>',
+                    insertTpl: '<output value="${path}" />',
+                    limit: 100,
+                    maxLen: 30,
+                    callbacks: {
+                        matcher: function(flag, subtext) {
+                            var match, regexp;
+                            regexp = new RegExp('(\\s+|^)' + flag + '([\\w_/]*)$', 'gi');
+                            match = regexp.exec(subtext);
+                            return match ? match[2] : null;
+                        }
+                    }
+                });
+            }
+
             $input.addClass('jstree-drop');
             $input.keydown(function (e) {
                 // deletion of entire output ref in one go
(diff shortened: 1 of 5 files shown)


esoergel:
:sunglasses: 
What'd you use for the gif screencast?

czue:
very slick

orangejenny:
fancy

src/javaRosa.js
```diff
@@ -837,6 +838,34 @@ define([
...
+                                    id: mug.ufid,
+                                    name: mug.form.getAbsolutePath(mug, true),
+                                    path: mug.absolutePath
+                                };
+                            })
+                            .filter(function(choice) { return choice.name; })
+                            .value(),
+                    displayTpl: '<li>${name}</li>',
+                    insertTpl: '<output value="${path}" />',
+                    limit: 100,
```
orangejenny: this could be more like 10, maybe 20, don't think scrolling beyond that many options is likely

src/javaRosa.js
```diff
@@ -837,6 +838,34 @@ define([
...
+                    maxLen: 30,
+                    callbacks: {
+                        matcher: function(flag, subtext) {
```
orangejenny: what's this doing? (i did read the docs but don't quite follow them) ...stripping leading whitespace and the "#form" part?

also, nitpicky, but \w includes underscores

src/javaRosa.js
emord: Yeah. I changed it to 10 (default is 5).

src/javaRosa.js
emord: That's understandable I had to dig around some of their code to figure it out exactly.

The first part comes from here: https://github.com/ichord/At.js/wiki/Base-Document#custom-query-matcher

subtext is the entire string before the cursor. It then looks for either the start of file or leading whitespace then the flag. Then it returns whatever was filter the user input.

> also, nitpicky, but \w includes underscores
I did that as it makes the regex slightly clearer to me. It'd also be cool if there was some fuzzy searching that allowed _ to be changed to -. Might look into this as a nice to have feature, but thought it would be nice to get some early user feedback and put it under an experimental flag

emord:
@esoergel Used gtk-recordmydesktop and convert to convert the ogv to gif. Now that i think about it I think ogv is supported in browsers, so probably didn't need the convert.

orangejenny:
looks good once tests pass

emord:
Should be merged with: https://github.com/dimagi/commcare-hq/pull/6877

src/javaRosa.js
```diff
@@ -837,6 +838,49 @@ define([
...
+                $input.atwho(_.extend(at_opts, {
+                    at: "#form",
+                    data: validMugs,
+                    insertTpl: '<output value="${path}" />',
+                })).atwho(_.extend(at_opts, {
+                    at: "#date",
+                    data: dateMugs,
+                    insertTpl: '<output value="format-date(date(${path}), \'%d/%n/%y\')" />',
+                })).atwho(_.extend(at_opts, {
+                    at: "#fulldate",
```
millerdev: I'm not sure I like adding extra tags like this for things special formatting. It seems the namespace could quickly become cluttered with hundreds of hard-to-remember tags. I think a better idea would be to add formatting extensions once we can represent `<output .../>` as an inline object rather than raw text.

emord: Yeah. That seems like something good to UAT before releasing to a wide audience, could quickly get unmanagable.

@amsagoff and I had thought that this was a good solution. I'll let her comment on if I should remove those.

millerdev:
Tested locally. Automated test pass, but what's *#formnull*?

screenshot 2015-05-28 14 42 00

millerdev:
Tests passed locally.

PR #428, opened 2015-05-28
Setvalue

emord:
resurrecting https://github.com/dimagi/Vellum/pull/414 after copy-paste was merged

What has changed since then:
* ignore retain now handles setvalue
* warn when a setvalue references a node
* fix setvalue parsing in parser
* only show the old default data value when it's present (that inserted the data directly into the data tree)

@millerdev 

@esoergel

### DIFF
--- a/src/form.js
+++ b/src/form.js
@@ -719,31 +719,6 @@ define([
                 this.fireChange(mug);
             }.bind(this);
         },
-        handleMugPropertyChange: function (mug, e) {
-            if (e.property === 'nodeID') {
-                var currentPath = this.getAbsolutePath(mug),
-                    valid = true,
-                    parsed;
-                try {
-                    parsed = xpath.parse(currentPath);
-                    if (_.isUndefined(parsed.steps)) {
-                        valid = false;
-                    }
-                } catch (err) {
-                    valid = false;
-                }
-                if (valid) {
-                    parsed.steps[parsed.steps.length - 1].name = e.previous;
-                    var oldPath = parsed.toXPath();
-                    this.vellum.handleMugRename(this, mug, e.val, e.previous, currentPath, oldPath);
-                }
-            } else {
-                if (mug.p.getDefinition(e.property).widget === widgets.xPath ||
-                    mug.p.getDefinition(e.property).widget === widgets.droppableText) {
-                    this.updateAllLogicReferences(mug);
-                }
-            }
-        },
         createQuestion: function (refMug, position, newMugType, isInternal) {
             var mug = this.mugTypes.make(newMugType, this);
             if (!mug.options.isControlOnly) {
--- a/src/form.js
+++ b/src/form.js
@@ -719,6 +719,32 @@ define([
                 this.fireChange(mug);
             }.bind(this);
         },
+        handleMugPropertyChange: function (mug, e) {
+            if (e.property === 'nodeID') {
+                var currentPath = this.getAbsolutePath(mug),
+                    valid = true,
+                    parsed;
+                try {
+                    parsed = xpath.parse(currentPath);
+                    if (_.isUndefined(parsed.steps)) {
+                        valid = false;
+                    }
+                } catch (err) {
+                    valid = false;
+                }
+                if (valid) {
+                    parsed.steps[parsed.steps.length - 1].name = e.previous;
+                    var oldPath = parsed.toXPath();
+                    this.vellum.handleMugRename(this, mug, e.val, e.previous, currentPath, oldPath);
+                }
+            } else {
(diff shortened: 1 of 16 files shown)


src/form.js
```diff
@@ -719,6 +719,31 @@ define([
                 this.fireChange(mug);
             }.bind(this);
         },
+        handleMugPropertyChange: function (mug, e) {
```
millerdev: This looks like an unused function. Am I missing something?

src/ignoreButRetain.js
```diff
@@ -308,6 +345,11 @@ define([
                     return attrs;
                 });
             },
+            getSetValues: function(mug) {
+                return _.map(mug.p.setValues, function(setValue) {
+                    return setValue;
+                });
```
millerdev: Could this be changed to `return mug.p.setValues;`? It doesn't look like the `map` is doing anything.

src/mugs.js
```diff
@@ -856,8 +863,25 @@ define([
             // could use a key-value widget for this in the future
             rawBindAttributes: {
                 presence: 'optional',
-                lstring: 'Extra Bind Attributes',
-            }
+                lstring: 'Extra Bind Attributes'
+            },
+            setValue: {
```
millerdev: Could we change the name of this to `defaultValue`?

src/parser.js
```diff
@@ -70,13 +70,8 @@ define([
...
             instances = _getInstances(xml),
-            data = $(instances[0]).children();
-
-        xml.find('setvalue').each(function () {
-            var $this = $(this);
-            form.addSetValue(
-                $this.attr('event'), $this.attr('ref'), $this.attr('value'));
-        });
+            data = $(instances[0]).children(),
+            setValues = xml.find('model > setvalue');
```
millerdev: Will this capture things in the data tree if there is a group named `model` with a child named `setvalue`?

src/widgets.js
```diff
@@ -426,6 +426,20 @@ define([
         return widget;
     };
 
+    var setValue = function (mug, options) {
```
millerdev: Can't find where this is used. Can it be removed?

src/writer.js
```diff
@@ -185,6 +179,29 @@ define([
...
+            xmlWriter.writeStartElement('setvalue');
+            xmlWriter.writeAttributeString('event', setValue.event);
+            xmlWriter.writeAttributeString('ref', setValue.ref);
+            xmlWriter.writeAttributeString('value', setValue.value);
+            xmlWriter.writeEndElement();
+        }
+
+        _.each(form.getSetValues(), function (setValue) {
+            writeSetValue(setValue);
+        });
```
millerdev: Could be shortened to `_.each(form.getSetValues(), writeSetValue);`

src/writer.js
```diff
@@ -185,6 +179,29 @@ define([
...
+
+        dataTree.walk(function (mug, nodeID, processChildren) {
+            if(mug && mug.options.getSetValues) {
+                _.each(mug.options.getSetValues(mug), function(setValue) {
+                    writeSetValue(setValue);
+                });
```
millerdev: Same here

tests/commtrack.js
```diff
@@ -33,10 +33,10 @@ define([
...
             assert.equal(trans.p.sectionId, "stock");
             assert.equal(trans.p.quantity, "/data/amount_received");
-            assert.equal(trans.p.entryId.value, "instance('commcaresession')/session/data/product_id");
-            assert.equal(trans.p.src.value, "instance('commcaresession')/session/data/case_id");
-            assert.equal(trans.p.dest.value, "instance('casedb')/casedb/case[@case_id=instance('commcaresession')/session/data/case_id]/index/parent");
-            assert.equal(trans.p.date.value, "today()");
+            assert.equal(trans.p.entryId, "instance('commcaresession')/session/data/product_id");
+            assert.equal(trans.p.src, "instance('commcaresession')/session/data/case_id");
+            assert.equal(trans.p.dest, "instance('casedb')/casedb/case[@case_id=instance('commcaresession')/session/data/case_id]/index/parent");
+            assert.equal(trans.p.date, "today()");
```
millerdev: This is great! Widget change handling will work better because of this.

src/mugs.js
emord: I was wavering on doing this since it's called setvalue in the xml, but it does get messy with those references and widget.setValue, so I changed it.

src/form.js
emord: I think this snuck in after the rebase from master to copy-paste-dev

PR #430, opened 2015-05-28
Async data source improvements

millerdev:
Also clean up window manager. All of this work is pre-req for case management.

@emord 
cc @sravfeyn

### DIFF
--- a/src/datasources.js
+++ b/src/datasources.js
@@ -56,30 +56,30 @@ define([
     edit_source,
     select_source
 ) {
-    var BLANK_FIXTURE = {
-        sourceUri: "",
-        defaultId: "No Lookup Table Found",
-        initialQuery: "",
-        name: '',
-        structure: {}
+    var BLANKS = {
+        fixture: {
+            sourceUri: "",
+            defaultId: "No Lookup Table Found",
+            initialQuery: "",
+            name: '',
+            structure: {}
+        }
     };
 
-    var vellum, dataSources, cachedDataSources;
+    var vellum, dataSources, dataCache = {};
 
     function init(instance) {
         vellum = instance;
         dataSources = vellum.opts().core.dataSources || [];
-        cachedDataSources = {
-            fixture: {}
-        };
     }
 
-    function cacheFixtures(data) {
+    function cacheData(data, type) {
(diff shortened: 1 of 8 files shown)


src/datasources.js
```diff
@@ -30,142 +30,116 @@
...
         }
 
+        function finish(data) {
+            dataCache[type] = {};
+            if (data.length === 0) {
+                dataCache[type][""] = {
+                    sourceUri: "",
+                    defaultId: "",
+                    initialQuery: "",
+                    name: "Not Found",
```
millerdev: I think we could improve this a bit by changing this to something like `source[type].name + " Not Found"`, but I don't think HQ gives its data sources a name right now.

emord: https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/app_manager/templates/app_manager/form_designer.html#L25 ?

Could change that to lookup tables, but that would also label products as lookup tables right now

millerdev: Oh, I was wrong. Yeah, should definitely change that to "Lookup Tables".

src/datasources.js
```diff
@@ -30,142 +30,116 @@
...
                 $.ajax({
                     type: 'GET',
                     url: source.endpoint,
                     dataType: 'json',
-                    success: function (data) { callback(data); },
-                    // TODO error handling
+                    success: finish,
+                    error: function (jqXHR, errorType, exc) {
+                        finish([]);
+                        window.console.log(util.formatExc(exc || errorType));
```
emord: Perhaps this could add an error message under the dropdown? (or whatever is affected in the future)

millerdev: Hmm, interesting idea. Problem is, this will be used in more places than just for the dropdown widget. Maybe we want to do something like the *Not Found* (when there is no data), except in this case indicate that the data could not be loaded because of a network error. Might also be nice for the user to be able to somehow retry in this case.

emord: We could keep dataErrors dictionary somewhere and if a mug/case list/whatever is affected by those errors the error could be displayed  to retry. So on refresh it calls some helper method here to return any errors with a filter function?

PR #431, opened 2015-05-29
Enable copy/paste for non mac firefox users

emord:
http://manage.dimagi.com/default.asp?169570#949638

@millerdev @esoergel

### DIFF
--- a/src/copy-paste.js
+++ b/src/copy-paste.js
@@ -19,7 +19,8 @@ define([
             position: 'absolute',
             width: 0,
             height: 0
-        }).css(offScreen).appendTo('body');
+        }).css(offScreen).appendTo('body'),
+        isMac;
 
     function focusTextarea($focus, value) {
         if ($focus.length === 0) {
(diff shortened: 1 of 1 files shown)


esoergel:
:+1:

src/copy-paste.js
```diff
@@ -329,27 +330,39 @@ define([
...
             var opts = this.opts().copyPaste;
             vellum = this;
+            isMac = /Mac/.test(navigator.platform);
+
+            function copyModifierKeyPressed(event) {
+                if (isMac && event.metaKey) {
+                    return true;
+                }
+
+                return event.ctrlKey;
```
millerdev: good place for a one-liner? `return (isMac && event.metaKey) || event.ctrlKey`

src/copy-paste.js
```diff
@@ -329,27 +330,39 @@ define([
         init: function () {
             var opts = this.opts().copyPaste;
             vellum = this;
+            isMac = /Mac/.test(navigator.platform);
```
millerdev: This could be moved up to where `isMac` is declared on line 23.

millerdev:
Looks good. Feel free to self-merge unless you want to address my non-blocker comments first.

src/copy-paste.js
emord: Oops. I think that was a bug anyways. that would have returned true if it a was a mac and ctrl(command?) key was pressed.

millerdev: oh, good catch

millerdev: although what does that do to Chrome on Mac?

millerdev: tested. works on both FF and Chrome on Mac

millerdev:
tests passed locally

PR #432, opened 2015-06-01
Editable div

emord:
This changes the inputs from being `input` or `textarea` into `<div contenteditable>`

- [x] change text area to divs
- [ ] change input to divs
  - [x] change text widget to div
  - [ ] change xpath widgets to div
- [ ] fix css properly
- [x] get divs to handle text content

Things that will probably wait until a future pr
- [ ] get divs to handle html content
- [ ] change question references to pretty labels
- [ ] have an editable menu for question references

I've probably missed some todo items on this list.

@millerdev @esoergel

### DIFF
--- a/src/javaRosa.js
+++ b/src/javaRosa.js
@@ -1022,7 +1022,7 @@ define([
         };
 
         widget.setPlaceholder = function (val) {
-            $input.attr("data-placeholder", val);
+            $input.attr("placeholder", val);
         };
 
         widget.getValue = function () {
@@ -1030,7 +1030,7 @@ define([
         };
 
         widget.getPlaceholder = function () {
-            return $input.attr('data-placeholder');
+            return $input.attr('placeholder');
         };
 
         widget.getDefaultValue = function () {
--- a/src/javaRosa.js
+++ b/src/javaRosa.js
@@ -827,25 +827,11 @@ define([
         }
         options.id = id;
 
-        var widget = widgets.base(mug, options);
-        var $input = $("<div contenteditable>")
-            .attr("name", widget.id)
-            .css('min-height', '3em')
-            .addClass('fake-textarea input-block-level itext-widget-input')
-            .on('change input', function (e) { widget.handleChange(); })
-            .focus(function () {
-                util.setCaretPosition(this, 0, widget.getValue().length);
-            })
-            .keyup(function (e) {
-                // workaround for webkit: http://stackoverflow.com/a/12114908
-                if (e.which === 9) {
-                    this.focus();
-                }
-            });
+        var widget = widgets.multilineText(mug, options);
 
         if (options.path === 'labelItext') {
             if (EXPERIMENTAL_UI) {
-                $input.atwho({
+                widget.input.atwho({
                     at: "#",
                     data: _.chain(mug.form.getMugList())
                            .map(function(mug) {
@@ -876,8 +862,8 @@ define([
                 });
             }
 
-            $input.addClass('jstree-drop');
-            $input.keydown(function (e) {
+            widget.input.addClass('jstree-drop');
+            widget.input.keydown(function (e) {
                 // deletion of entire output ref in one go
                 if (e && e.which === 8 || e.which === 46) {
                     var control = widget.getControl()[0],
@@ -926,10 +912,6 @@ define([
         widget.isDefaultLang = widget.language === widget.defaultLang;
         widget.isSyncedWithDefaultLang = false;
         widget.hasNodeIdPlaceholder = options.path === 'labelItext';
-
-        widget.getControl = function () {
-            return $input;
-        };
 
         widget.getItextItem = function () {
             // Make sure the real itextItem is being updated at all times, not a stale one.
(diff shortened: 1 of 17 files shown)


src/javaRosa.js
```diff
@@ -824,16 +824,17 @@ define([
         options.id = id;
 
         var widget = widgets.base(mug, options);
-        var $input = $("<textarea></textarea>")
+        var $input = $("<div contenteditable>")
             .attr("name", widget.id)
-            .attr("rows", "2")
-            .addClass('input-block-level itext-widget-input')
+            .addClass('fake-textarea input-block-level itext-widget-input')
```
emord: Need to figure out a good way to specify # of rows in a div. May just need to go with height though.

src/util.js
```diff
@@ -280,6 +280,8 @@ define([
             pos = sel.text.length;
         } else if (typeof ctrl.selectionStart !== 'undefined') {
             pos = ctrl.selectionStart;
+        } else if (_.contains(ctrl.classList, "fake-textarea")) {
+            return $(ctrl).caret('pos');
```
emord: I think caret will work for more than just this. Haven't tested it out though

src/javaRosa.js
orangejenny: I think min-height is as good as it gets.

src/less-style/question-props.less
```diff
@@ -157,3 +157,17 @@
...
+    -webkit-appearance: textarea;
+    border: 1px solid #cccccc;
+    border-radius: 4px;
+    font: medium -moz-fixed;
+    font: -webkit-small-control;
+    height: 28px;
+    overflow: auto;
+    padding: 2px;
+    resize: both;
+    width: 400px;
```
orangejenny: Setting all pseudo-textareas to a pixel width seems suspicious. Is this temporary?

src/less-style/question-props.less
```diff
@@ -157,3 +157,17 @@
...
 }
+
+.fake-textarea {
+    -moz-appearance: textfield-multiline;
+    -webkit-appearance: textarea;
+    border: 1px solid #cccccc;
+    border-radius: 4px;
+    font: medium -moz-fixed;
+    font: -webkit-small-control;
+    height: 28px;
```
orangejenny: Is this going to prevent the div from growing as someone types in a lot of text?

src/less-style/question-props.less
```diff
@@ -157,3 +157,17 @@
   padding: 0 2em;
   margin-bottom: 2em;
 }
+
+.fake-textarea {
```
orangejenny: Is it on your list to add :focus styles? Not important right now, but later.

src/templates/xpath.html
```diff
@@ -16,8 +16,8 @@
                 XPath Expression:
             </label>
             <div class="controls">
-                <textarea rows="5"
-                          class="input-block-level fd-xpath-editor-text jstree-drop"></textarea>
+                <div contenteditable rows="5"
```
orangejenny: might as well drop the rows attribute

tests/javaRosa.js
```diff
@@ -362,40 +358,36 @@ require([
             });
         });
 
-        it("drag question into label makes output ref in correct position", function (done) {
```
orangejenny: nice bit of test cleanup

src/less-style/question-props.less
orangejenny: or maybe that's what's happening with the hqstyle-core.css changes...?

emord: Yeah. I'm going to fix that properly later. Probably going to do it in HQ so that I can be sure it inherits all the correct styles and still looks like textarea/input

src/javaRosa.js
```diff
@@ -824,16 +824,17 @@ define([
         options.id = id;
 
         var widget = widgets.base(mug, options);
-        var $input = $("<textarea></textarea>")
+        var $input = $("<div contenteditable>")
```
millerdev: I wonder how much jQuery manipulation we do with these inputs. What do you think of making a wrapper object for the div that gives it the same interface as a regular input? Then you wouldn't have to go around an change every piece of code that does `$input.val()`, for example.

src/less-style/question-props.less
millerdev: I think we should be using font-based dimensions rather than pixels. So this should be something like `1em`.

src/javaRosa.js
emord: https://github.com/dimagi/Vellum/pull/432/files#diff-0c6adbb645e08e26dfcc96c660406ee4R1

I really like this solution. It will get a little more complicated once I add in support for HTML elements, but it's a good place to keep it.

I'm going to keep `val` `setValue` and `getValue` the same API wise. i.e you will still need to use <output value> when interacting with these. I'll probably create something like `getHTML` for tests and anything else that may need the HTML content, but I don't imagine that will be much.

src/less-style/question-props.less
emord: Yeah. I don't actually intend on keeping this here. I'll probably try to add it to HQ style so that we can be sure the divs will capture the HQ styling consistently

src/jquery-extensions.js
```diff
@@ -0,0 +1,12 @@
+(function ($) {
+    var originalVal = $.fn.val;
+    $.fn.val = function(){
+        if (this.hasClass('fake-textarea') || this.hasClass('fake-input')) {
+            return $.fn.text.apply(this, arguments);
+        }
+        var result = originalVal.apply(this,arguments);
+        if(arguments.length>0)
+            $(this).change();
```
millerdev: Why fire change here? There are times when you want to set the value programmatically but not fire the change event.

emord: You caught this right as I did. It's already updated and won't be there. (copied from SO)

emord:
Closing while I rebase this to make it a little more readable

PR #434, opened 2015-06-04
Editable div

emord:
Resurrecting https://github.com/dimagi/Vellum/pull/432

Going to QA this to make sure I didn't miss anything, but should offer the same functionality as the previous inputs, but with content editable divs

cc: @gcapalbo


src/less-style/question-props.less
```diff
@@ -157,3 +157,52 @@
   padding: 0 2em;
   margin-bottom: 2em;
 }
+
+.fake-input {
```
emord: Originally wanted this to inherit from styles in HQ. however that proved to be a little more involved than I thought. Would require editing the bootstrap files as we're on less 1.3 (was going to use: http://lesscss.org/features/#extend-feature) and compilation breaks when upgraded to 1.4

src/templates/xpath_expression.html
```diff
@@ -10,7 +10,7 @@
     </div>
     <div class="control-row xpath-expression-row">
         <div class="span4">
-            <input class="input-block-level left-question xpath-edit-node jstree-drop"
+            <div contenteditable=true class="fake-input input-block-level left-question xpath-edit-node jstree-drop"
```
millerdev: `contenteditable="true"`

src/less-style/question-props.less
millerdev: What do you think of not using the word "fake" in the name since it's not fake? Maybe `input` or `fd-input`? Same with "fake-textarea" below.

src/widgets.js
```diff
@@ -186,6 +190,62 @@ define([
         return widget;
     };
 
+    var multilineText = function (mug, options) {
+        // TODO: make this inherit from normal
+        // normal adds property- to all ids (make javarosa accept that)
```
millerdev: it would also improve support for mug messages.

gcapalbo:
looks good to me, but will let @millerdev sign off and merge

PR #433, opened 2015-06-03
Dialog jQuery UI => bootstrap

orangejenny:
some ui refactoring, in the hope that we can get off jQuery UI altogether soon

Old:
screen shot 2015-06-03 at 6 19 47 pm

New:
screen shot 2015-06-03 at 6 19 53 pm

PR #435, opened 2015-06-04
Another dialog

orangejenny:
another https://github.com/dimagi/Vellum/pull/433

@proteusvacuum

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -697,16 +697,8 @@ define([
                 title: title,
                 closeButtonTitle: closeButtonTitle
             })),
-            $body = $("body"),
-            keyup = function (event) {
-                if (event.keyCode === 27) { // escape
-                    $modal.modal('hide');
-                }
-            };
-        $body.on("keyup", keyup);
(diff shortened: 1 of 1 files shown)


src/core.js
```diff
@@ -696,17 +696,8 @@ define([
...
-            })),
-            $body = $("body"),
-            keyup = function (event) {
-                if (event.keyCode === 27) { // escape
-                    $modal.modal('hide');
-                }
-            };
-        $body.on("keyup", keyup);
-        $modal.on('hide', function () {
-            $body.off("keyup", keyup);
```
millerdev: Why did you remove this keyup code? This makes the escape key dismiss the dialog.

orangejenny: That's how bootstrap modals behave by default (the `keyboard` parameter defaults to true) and having this extra handler prevented me from turning off that behavior.

millerdev: > That's how bootstrap modals behave by default...

Oh, cool. Didn't realize that.

PR #437, opened 2015-06-05
Last dialogs

orangejenny:
get rid of the last of the jQuery UI dialogs

@proteusvacuum

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -459,7 +459,7 @@ define([
                         _this.closeModal();
                     }
                 }
-            ], false);
+            ], false, "icon-warning-sign");
             var content = "There are validation errors in the form.  Do you want to continue anyway?";
             content += "<br><br>WARNING: The form will not be valid and likely not perform correctly on your device!";
             $modal.find(".modal-body").html(content);
@@ -582,7 +582,7 @@ define([
                     $modal.find('.btn-info').attr('disabled', 'disabled');
                 }
             }
-        ], "Cancel");
+        ], "Cancel", "icon-warning-sign");
 
         var diff = util.xmlDiff(formText, serverForm);
 
@@ -652,7 +652,7 @@ define([
         this.$f.find('.fd-modal-generic-container .modal').modal('hide');
     };
     
-    fn.generateNewModal = function (title, buttons, closeButtonTitle) {
+    fn.generateNewModal = function (title, buttons, closeButtonTitle, headerIcon) {
         if (typeof closeButtonTitle === "undefined") {
             closeButtonTitle = "Close";
         }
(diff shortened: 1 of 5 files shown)


src/core.js
```diff
@@ -1091,42 +1051,15 @@ define([
                     msg = "Parsing Error. Please check that your form is valid XML.";
                 }
 
-                // this button doesn't seem to actually exist
-                // todo: fix
```
orangejenny: this is a Mike White leftover, happy to put it back if someone else knows what it's talking about...

src/core.js
```diff
@@ -1848,10 +1789,10 @@ define([
         var url = saveType === 'patch' ?  opts.patchUrl : opts.saveUrl;
 
         $(document).ajaxStart(function () {
-            _this.showWaitingDialog();
+            _this.showWaitingModal();
         });
         $(document).ajaxStop(function () {
-            _this._hideConfirmDialog();
+            _this.$f.find('.fd-modal-generic-container .modal').modal("hide");
```
millerdev: Why not `_this.closeModal()`?

src/javaRosa.js
```diff
@@ -2102,12 +2102,12 @@ define([
...
                     name: "Edit Bulk Translations",
                     action: function (done) {
-                        _this.showItextDialog(done);
+                        _this.showItextModal(done);
                     }
                 }
             ]);
         },
-        showItextDialog: function (done) {
+        showItextModal: function (done) {
```
millerdev: Curious why you decided to rename from "Dialog" to "Modal". *Modal* is short for *modal dialog*, right?

src/core.js
```diff
@@ -30,7 +30,7 @@ define([
     'jquery.jstree',
     'jquery.bootstrap',
     'jquery.fancybox',  // only thing we use fancybox for is its spinner, no actual display of anything
-    'jquery-ui',  // used for buttons in Edit Source XML, and dialogs
+    'jquery-ui',  // used for autocomplete
```
millerdev: is it going to be possible to substitute the autocomplete functionality as well, or are we stuck with jquery-ui?

orangejenny: Hopefully we can replace autocomplete with bootstrap's "typeahead" - haven't looked into it yet, but planning on it.

src/javaRosa.js
orangejenny: No good reason. For the most part, we've ended up using "dialog" for functions that use the jQuery UI widget and "modal" for the bootstrap widget, so my todo list for this PR was basically "grep -i dialog", and I ended up changing this.

src/core.js
orangejenny: good point, will fix

src/core.js
emord: This works with inputs, but unfortunately won't work with content editable divs.

Took a quick look and it seems like there are plugins that handle that, but I'm not sure how well those will work.

emord:
Looks good to me, but can't merge.

PR #438, opened 2015-06-05
Question shortcut

emord:
Release the  question shortcut. The key is `/data/`

Analytics are:

Category | Action | Label | Description
-----------|---------|----------|------------
Question Reference |	Drag n Drop vs. Autocomplete	|display condition vs. validation condition|	Question reference created via either autocomplete or drag n drop (not direct writing)
Output Value|	Drag n Drop vs. Autocomplete|	date format|	Output value created via autocomplete or drag n drop

cc @gcapalbo

### DIFF
--- a/src/javaRosa.js
+++ b/src/javaRosa.js
@@ -837,7 +837,7 @@ define([
             });
 
         if (options.path === 'labelItext') {
-            util.questionAutoComplete($input, mug.form, {
+            util.questionAutocomplete($input, mug.form, {
                 category: "Output Value",
                 insertTpl: '<output value="${name}" />',
                 property: "labelItext",
--- a/src/javaRosa.js
+++ b/src/javaRosa.js
@@ -840,7 +840,9 @@ define([
         if (options.path === 'labelItext') {
             if (EXPERIMENTAL_UI) {
                 util.questionAutoComplete($input, mug.form, {
-                    insertTpl: '<output value="${name}" />'
+                    category: "Output Value",
+                    insertTpl: '<output value="${name}" />',
+                    property: "labelItext",
                 });
             }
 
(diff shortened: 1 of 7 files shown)


src/javaRosa.js
```diff
@@ -838,37 +837,11 @@ define([
...
-                        matcher: function(flag, subtext) {
-                            var match, regexp;
-                            regexp = new RegExp('(\\s+|^)' + flag + '([\\w_/]*)$', 'gi');
-                            match = regexp.exec(subtext);
-                            return match ? match[2] : null;
-                        }
-                    }
-                });
-            }
+            util.questionAutoComplete($input, mug.form, {
```
orangejenny: picky: can autocomplete be one word, not two?

emord: done

gcapalbo:
:+1: 
Does @millerdev typically sign off on Vellum new feature PRs or is this good to merge?

src/util.js
```diff
@@ -373,6 +373,47 @@ define([
...
+                    .filter(function(choice) { return choice.name; })
+                    .value(),
+            displayTpl: '<li>${name}</li>',
+            insertTpl: options.insertTpl,
+            limit: 10,
+            maxLen: 30,
+            callbacks: {
+                matcher: function(flag, subtext) {
+                    var match, regexp;
+                    regexp = new RegExp('(\\s+|^)' + flag + '([\\w_/]*)$', 'gi');
```
millerdev: Would it make sense to `RegExp.escape(flag)` before adding it here?

emord: This isn't really necessary right now as it's always `/data/` but went ahead and escaped it because I assume it will get generalized at some point and will probably cause me a headache of a bug.

src/widgets.js
```diff
@@ -284,7 +284,9 @@ define([
                             super_setValue(val);
                             widget.handleChange();
                         }
-                    }
+                    },
+                    form: mug.form,
+                    questionType: options.path,
```
millerdev: What is this? I don't see where it's used.

millerdev:
Sorry for the slowness of this review. Except for my final comment I think it looks good. Feel free to merge this after addressing the comment by either removing those two lines (assuming they're not needed) or keeping them if they're necessary for something I didn't notice.

src/widgets.js
emord: Oops. that snuck in from some other work I was doing on case management. form is used here: https://github.com/dimagi/Vellum/pull/438/files#diff-3514e86c408245666c24881d2a721031R84

PR #439, opened 2015-06-05
Remove unused template

emord:
@millerdev think this was part of an old model iteration PR. isn't used anymore

code buddy: @gcapalbo


millerdev:
:+1:

PR #440, opened 2015-06-05
replace itemset autocomplete with at.js

orangejenny:
@proteusvacuum / @emord

### DIFF
--- a/src/itemset.js
+++ b/src/itemset.js
@@ -413,7 +413,7 @@ define([
                     callbacks: {
                         filter: function(query, data, searchKey) {
                             return _.filter(data, function(item) {
-                                return item.name.indexOf(query) != -1 && item.name !== query;
+                                return item.name.indexOf(query) !== -1 && item.name !== query;
                             });
                         },
                         matcher: function(flag, subtext, should_startWithSpace) {
--- a/src/itemset.js
+++ b/src/itemset.js
@@ -406,6 +406,7 @@ define([
         input.attr("name", name);
         return {
             addAutoComplete: function(sources, changeFunction) {
+                var selectedValue = "";
                 input.atwho({
                     at: "",
                     data: sources,
@@ -420,7 +421,13 @@ define([
                         matcher: function(flag, subtext, should_startWithSpace) {
                             return input.val();
                         },
+                        beforeInsert: function(value, $li) {
+                            selectedValue = value;
+                        },
                     }
+                });
+                input.on("inserted.atwho", function(event, $li, otherEvent) {
+                    input.val(selectedValue);
                 });
                 input.on("blur change", function() {
                     if (_.isFunction(changeFunction)) {
--- a/src/itemset.js
+++ b/src/itemset.js
@@ -410,6 +410,7 @@ define([
                     at: "",
                     data: sources,
                     maxLen: Infinity,
+                    suffix: "",
                     callbacks: {
                         filter: function(query, data, searchKey) {
                             return _.filter(data, function(item) {
@@ -422,7 +423,6 @@ define([
                     }
                 });
                 input.on("blur change", function() {
-                    input.val(input.val().trim());
                     if (_.isFunction(changeFunction)) {
                         changeFunction();
                     }
--- a/src/itemset.js
+++ b/src/itemset.js
@@ -406,14 +406,26 @@ define([
         input.attr("name", name);
         return {
             addAutoComplete: function(sources, changeFunction) {
-                input.autocomplete({
-                    source: sources,
-                    minLength: 0,
-                    change: changeFunction,
-                    close: changeFunction
-                }).focus(function (e) {
-                    // populate the list
-                    $(this).autocomplete('search', $(this).val());
+                input.atwho({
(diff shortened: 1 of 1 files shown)


src/itemset.js
```diff
@@ -406,14 +406,26 @@ define([
...
+                            return _.filter(data, function(item) {
+                                return item.name.indexOf(query) !== -1 && item.name !== query;
+                            });
+                        },
+                        matcher: function(flag, subtext, should_startWithSpace) {
+                            return input.val();
+                        },
+                    }
+                });
+                input.on("blur change", function() {
```
orangejenny: This gets called twice when you select an item, but I think that would have happened with the previous autocomplete as well (calling close, then change).

src/itemset.js
```diff
@@ -406,14 +406,26 @@ define([
...
-                    // populate the list
-                    $(this).autocomplete('search', $(this).val());
+                input.atwho({
+                    at: "",
+                    data: sources,
+                    maxLen: Infinity,
+                    callbacks: {
+                        filter: function(query, data, searchKey) {
+                            return _.filter(data, function(item) {
+                                return item.name.indexOf(query) !== -1 && item.name !== query;
```
orangejenny: `item.name !== query` prevents the single-item menu from appearing on exact match, because I found this annoying while testing, but it's a change from the current behavior. Open to changing it back if anyone else feels strongly.

emord:
I like using only one library for autocomplete, but realized this will have some weird behavior with using tab.

Whenever there are nested attributes tab can autocomplete and still have options available, so you can't actually tab to the next input. You can try it out by going to itemsets and using inner_attribute hit tab and then inner_attribute/extra-inner-attribute will pop up and you can't tab to the next input.

I don't think this is a problem right now with products or lookup tables, but I think it will be with locations and cases.

I think to fix it, it's possible to insert a space afterwards and change the validation func to trim the value before validating it. That should close the popup and allow you to tab to the next question without a validation error popping up. There might be an easier way to fix this though

proteusvacuum:
This looks fine to merge to me if @emord's comments aren't blocking.

orangejenny:
I'd like to get the tab behavior resolved before merging this.

@emord: "I think to fix it, it's possible to insert a space afterwards and change the validation func to trim the value before validating it. That should close the popup and allow you to tab to the next question without a validation error popping up."

So, at.js does automatically insert a space, which is why I put a `trim` in the blur/change handler...this is probably why that space is added. Leaving the space in the input, and then trimming it on validation, sounds reasonable & quick.

emord:
> So, at.js does automatically insert a space,

I can get this behavior with div contenteditable, but not with input or textarea. (at least on chrome/linux)

millerdev:
> Leaving the space in the input, and then trimming it on validation, sounds reasonable & quick.

I'm -1 on that. Validation should not modify the value. I'm also generally -1 on automatically trimming spaces unless we can be sure we only ever trim spaces that were automatically added (i.e., not trimming spaces added purposely by the user).

src/itemset.js
millerdev: How does this compare to the way github autocomplete works?

orangejenny: GitHub shows the single-item match, as does FogBugz, and jQuery UI itself...point taken, I'll change it back.

orangejenny:
@emord @millerdev 

So, I tweaked it, but it's not quite perfect. Two things are weird:
- The tabbing, as described above. I don't see a way around this without turning off tab completion in at.js (which could be fine; I don't think the tab completion adds much value, just generally opposed to modifying third-party libraries)
- I reinstated the behavior where the menu doesn't appear if it only has one element. This is a little unusual, as autocomplete behavior goes, but otherwise, you can't tab out of the field at all (it just repeatedly selects the single item displayed in the menu).

On the other hand...I'm convinced that dropping jQuery UI would be a good thing.

Thoughts?

emord:
@orangejenny was looking through source code for other reason. would probably go here: https://github.com/ichord/At.js/blob/ee088630480ebd0681b02296d0be30ae0c27ff9f/src/app.coffee#L137 

Should be a pretty easy PR for it if he's into accepting that sort of thing.

orangejenny:
@emord Good thought, submitted a PR

millerdev:
Briefly tested locally, and it looks good to me. @emord, you can merge when you're happy with this.

> Good thought, submitted a PR

I took the liberty to edit your comment to add a link.

> I'm convinced that dropping jQuery UI would be a good thing.

Me too! Don't give up!

emord:
I really hate breaking tabbing through the form, but I'm ok with merging this for now as it doesn't really affect anything for fixtures and products.

Will probably want to revisit when including cases and locations though/upgrade at.js to use your PR.

Can't merge at the moment. Also do you want to remove jquery-ui with this PR? I thought you mentioned this was the last thing keeping that in here.

orangejenny:
It's only a little bit broken...but revisiting when cases/locations come in makes sense.

There's another autocomplete in expressEditor.js, that's the last jQuery UI widget.

Mergeable now.

PR #441, opened 2015-06-08
Don't write the placeholder in groups if no label exists

emord:
http://manage.dimagi.com/default.asp?169124#952225

I'm not sure that you could ever get to this state through the UI, but if there isn't a label for groups then we won't insert them.

@millerdev has context

code buddy @gcapalbo

### DIFF
--- a/src/javaRosa.js
+++ b/src/javaRosa.js
@@ -941,7 +941,7 @@ define([
         widget.setItextValue = function (value) {
             var itextItem = widget.getItextItem();
             if (itextItem) {
-                if (!value) {
+                if (!value && !mug.options.canHaveBlankLabel) {
                     value = widget.getPlaceholder();
                 }
                 if (widget.isDefaultLang) {
@@ -1657,14 +1657,12 @@ define([
                         for (var k = 0; k < forms.length; k++) {
                             form = forms[k];
                             val = form.getValueOrDefault(lang);
-                            if (val) {
-                                xmlWriter.writeStartElement("value");
-                                if(form.name !== "default") {
-                                    xmlWriter.writeAttributeString('form', form.name);
-                                }
-                                xmlWriter.writeXML(xml.normalize(val));
-                                xmlWriter.writeEndElement();
+                            xmlWriter.writeStartElement("value");
+                            if(form.name !== "default") {
+                                xmlWriter.writeAttributeString('form', form.name);
                             }
+                            xmlWriter.writeXML(xml.normalize(val));
+                            xmlWriter.writeEndElement();
                         }
                         xmlWriter.writeEndElement();
                     }
--- a/src/mugs.js
+++ b/src/mugs.js
@@ -1379,6 +1379,7 @@ define([
         isSpecialGroup: true,
         isNestableGroup: true,
         isTypeChangeable: false,
+        canHaveBlankLabel: true,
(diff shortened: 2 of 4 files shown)


src/javaRosa.js
```diff
@@ -941,7 +941,7 @@ define([
         widget.setItextValue = function (value) {
             var itextItem = widget.getItextItem();
             if (itextItem) {
-                if (!value) {
+                if (!value && !mug.options.canHaveBlankLabel) {
                     value = widget.getPlaceholder();
```
millerdev: Does this mean that all group question types (Group, Repeat, Field List) will not use the question ID as default label?

millerdev:
After reviewing the FB case, I cannot tell which bread-crumb is appearing incorrectly (Derek did not specify). There are no bread-crumbs in vellum that I know of, so the bug may not even be a vellum issue. Do we know that this change fixes the bread-crumb issue (has it been tested)?

src/javaRosa.js
emord: This is reverting to old behavior for groups, which means that if there is a blank label, then we will not automatically update it with the placeholder. I think if you create a group the placeholder will still be there unless you manually blank it out. This was the previous behavior before the bug was logged

emord:
I believe that he meant the breadcrumb was showing up on the phone when he wrote the form as the change caused the placeholders to be written when it used to be `<value></value>`.

Right now he has whitespace as the label, which on the phone looks the same, but in vellum when you select the language instead of question id for display it shows up as whitespace.

Old behavior (pre my change that caused the bug):
1. load a form with a group `<value></value>`
2. select a language to display question tree in
3. group shows the placeholder, other questions show label text
4. `<value></value>` causes no breadcrumb on mobile

After my change that caused the bug:
1. load a form with a group `<value></value>`
2. when writing the form it is replaced with `<value>placeholder</value>`
3. on mobile breadcrumb is placeholder

Workaround:
1. load a form with a group `<value></value>`
2. Change group to be `<value>   </value>`
3. On hq question tree shows whitespace
3. on mobile breadcrumb is blank which is desired

So this change brings back old behavior for groups, but keeps new behavior for text and other questions, which I believe is desired. Hopefully that makes things more clear

millerdev:
Do you think it's a problem that the group label displays a placeholder even though it writes no itext (i.e., a blank label) for the group? Here's a screenshot after manually performing the steps in [the test](https://github.com/dimagi/Vellum/pull/441/files#diff-b249980439a6e6297a2387c58fe527beR632):

screenshot 2015-06-09 19 55 23

And the resulting itext portion of the XForm:
```xml
<translation lang="en" default="">
	<text id="label-label">
		<value>label</value>
		<value form="image">jr://file/commcare/image/data/label.png</value>
	</text>
	<text id="group/text-label">
		<value>text</value>
	</text>
</translation>
```

millerdev:
BTW, I was not able to clear the group label itext using these steps in a form with multiple languages. It always writes itext for the group, using the placeholder as the itext value.

ctsims:
Seems awkward since there's no way to "bring" the 'group' text from the
background to the foreground. I love the gray "this is what this will be by
default" affordance, but it definitely feels like it will muddy up the
expectation if it doesn't actually write that.

emord:
Closing after conversation with @millerdev

PR #442, opened 2015-06-08
Convert == to =

emord:
http://manage.dimagi.com/default.asp?170516#955370

Technically it only works in chrome because whenever you do val('blah') to a select it just defaults to the first option.

I think the real bug is here: https://github.com/dimagi/js-xpath/blob/master/models.js#L341 as I don't think == is a valid xpath operator.

This: https://github.com/dimagi/js-xpath/blob/master/models.js#L359 makes me think that whoever wrote this knows that. @czue I think?

Anyways, I tested this in firefox and chrome on linux and it worked on both.

code buddy: @gcapalbo

### DIFF
--- a/src/expressionEditor.js
+++ b/src/expressionEditor.js
@@ -48,9 +48,11 @@ define([
         }
     };
     function addOp(expr, value, label) {
+        value = xpathmodels.expressionTypeEnumToXPathLiteral(value);
         simpleExpressions[value] = expr;
         operationOpts.push([label, value]);
     }
+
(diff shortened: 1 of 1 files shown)


czue:
confirming that the xpath equals should be `=`

czue:
I definitely don't remember enough about that xpath code to know whether that's a bug. i thought those constants were just for its internal representation, though. it seems to be converting to single `=` here: https://github.com/dimagi/js-xpath/blob/master/models.js#L358

czue:
whoops see you already pointed that out.

PR #443, opened 2015-06-09
Add alternate copy/paste method for Safari

millerdev:
http://manage.dimagi.com/default.asp?170523

Safari is a real pain when it comes to copy/paste, so I implemented a stupid workaround that should work in any situation where the keydown-copy-paste-event-hidden-textarea dance does not work, which seems potentially useful.

I ran out of energy when it came to implementing the UI. Ideas for improvements are welcome.

screenshot 2015-06-09 15 46 26

Clicking this new button reveals a text box containing the serialized text for the selected questions.

screenshot 2015-06-09 15 46 47

Finally, an *Insert Questions* button at the bottom to deserialize the text and insert questions into the form.

### DIFF
--- a/src/copy-paste.js
+++ b/src/copy-paste.js
@@ -389,7 +389,7 @@ define([
                 });
             }).keyup(function (e) {
                 // workaround for webkit: http://stackoverflow.com/a/12114908
-                if(e.which === 9) {
+                if(e.which === 9) { // tab
                     copyPasteArea.select();
                 }
             });
--- a/src/copy-paste.js
+++ b/src/copy-paste.js
@@ -357,7 +357,19 @@ define([
         },
         displayMultipleSelectionView: function () {
             this.__callOld();
-            var html = $(copy_paste_help({"metachar": (isMac ? "\u2318" : "Ctrl+")}));
+            var html = $(copy_paste_help({"metachar": (isMac ? "\u2318" : "Ctrl+")})),
+                showCopyArea = html.find(".show-copy-area"),
+                copyArea = html.find(".copy-area"),
+                insertQuestions = html.find(".insert-questions");
+            showCopyArea.click(function () {
+                showCopyArea.hide();
+                copyArea.parent().removeClass("hide");
+                copyArea.val(copy()).focus().select();
+                insertQuestions.removeClass("hide").click(function () {
+                    paste(copyArea.val());
+                });
+                return false;
+            });
             this.$f.find(".fd-props-content").html(html);
         }
     });
--- a/src/copy-paste.js
+++ b/src/copy-paste.js
@@ -372,13 +372,9 @@ define([
                 // HACK show textarea for copy/paste because the hidden
                 // textarea dance doesn't work in Safari
                 showCopyPasteBox();
-            }
-            html.find(".insert-questions").click(function () {
-                paste(copyPasteArea.val());
-            });
-            this.$f.find(".fd-props-content").html(html);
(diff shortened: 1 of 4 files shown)


millerdev:
cc @biyeun

orangejenny:
> in any situation where the keydown-copy-paste-event-hidden-textarea dance does not work

Any known problematic scenarios other than "using Safari"? If not, might be worth detecting browser and customizing the help UI based on that.

millerdev:
Thanks for the push, @orangejenny. Here's what the Safari copy/paste screen looks like:
screenshot 2015-06-09 21 45 51

emord:
Looks good to me.

Curious if this has been tested on IE? Although it'd probably be best to just ignore that until we get an issue about it not working

millerdev:
> Curious if this has been tested on IE?

Does Vellum even load in IE? I certainly did not test it there.

orangejenny:
> Does Vellum even load in IE? I certainly did not test it there.

This is probably bad, but I find it kind of charming that none of us know this.

src/copy-paste.js
```diff
@@ -20,7 +20,9 @@ define([
             width: 0,
             height: 0
         }).css(offScreen).appendTo('body'),
-        isMac = /Mac/.test(navigator.platform);
+        isMac = /Mac/.test(navigator.platform),
+        isChrome = /Chrome/.test(navigator.userAgent),
+        isSafari = /Safari/.test(navigator.userAgent) && !isChrome;
```
orangejenny: So these are userAgent strings that contain both "Safari" and "Chrome", and they're really Chrome?

src/copy-paste.js
```diff
@@ -357,7 +359,43 @@ define([
...
+                });
+            }
+            copyPasteArea.focus(function () {
+                copyPasteArea.select().mouseup(function() {
+                    copyPasteArea.off('mouseup');
+                    return false;
+                });
+            }).keyup(function (e) {
+                // workaround for webkit: http://stackoverflow.com/a/12114908
+                if(e.which === 9) {
```
orangejenny: Not blocking, but it'd be nice to have a set of constants so we could use something like KeyCode.TAB instead of the numbers.

src/copy-paste.js
emord: "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/43.0.2357.81 Safari/537.36"

user agents are awful things

millerdev: http://stackoverflow.com/a/7768006/10840

src/templates/copy_paste_help.html
```diff
@@ -1,20 +1,31 @@
...
+            <div class="span3">
+                <h1><%=metachar%>V</h1>
+                <span>for paste</span>
+            </div>
         </div>
     </div>
+    <div class="copy-paste-box hide">
+        <p>
+            Copy from or Paste into this text box:
+            <textarea style="width: 100%; height: 100px;"></textarea>
```
orangejenny: should go in LESS

src/copy-paste.js
orangejenny: ha, that's awful

orangejenny:
Browser-specific UI is much simpler - nice!

millerdev:
Updated styles and added comment about tab key code.

orangejenny:
ran tests locally

PR #444, opened 2015-06-09
Cleanup atwho containers

emord:
This cleans up the atwho-containers that will show up at the bottom of the body tag.

A nicer solution would be to PR something to at.js that cleans this stuff up automatically, but I would need to go a little more in depth with the library and how it keeps references (and brush up on coffeescript)

cc: @gcapalbo

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -825,6 +825,7 @@ define([
             data.node.icon = mug.getIcon();
             _this.refreshCurrentMug();
         }).bind("deselect_all.jstree deselect_node.jstree", function (e, data) {
+            $('.atwho-container').remove();
             _this.resetQuestionTypeGroups();
         }).bind('model.jstree', function (e, data) {
             // Dynamically update node icons. This is unnecessary for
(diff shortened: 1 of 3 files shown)


src/core.js
```diff
@@ -825,6 +825,7 @@ define([
             data.node.icon = mug.getIcon();
             _this.refreshCurrentMug();
         }).bind("deselect_all.jstree deselect_node.jstree", function (e, data) {
+            $('.atwho-container').remove();
```
emord: This would actually cause issues if atwho was ever implemented outside of a question type, but I don't see that happening anytime soon.

src/javaRosa.js
```diff
@@ -826,6 +826,7 @@ define([
         var widget = widgets.base(mug, options);
         var $input = $("<textarea></textarea>")
             .attr("name", widget.id)
+            .attr("id", widget.id)
```
emord: These id additions are not necessarily required, but it helps with debugging: https://github.com/ichord/At.js/blob/ee088630480ebd0681b02296d0be30ae0c27ff9f/dist/js/jquery.atwho.js#L261

It'll either be the id or some unreadable id that you can't associate with a widget easily.

src/core.js
millerdev: Could also do something like this in widgets that use autocomplete:
```js
mug.on("teardown-mug-properties", function () {
    $('.atwho-container').remove();
}, null, "teardown-mug-properties");
```

PR #446, opened 2015-06-10
Cleanup atwho containers

emord:
https://github.com/dimagi/Vellum/pull/444#discussion_r32084248

jquery ui autocomplete in the expressoin editor for itemsets also has the same issue, but those will probably get replaced with at who soon

cc: @gcapalbo


millerdev:
tests passed locally

PR #447, opened 2015-06-10
Fix fieldset overflow

emord:
Found this while working on changing to editable divs, but it also applies to master.

Before:
selection_011

After:
selection_012

It's probably a small edge case. Could also use word-wrap: break-word on the alert box, but I think that's unnecessary.

code buddy: @gcapalbo

### DIFF
--- a/src/less-style/question-props.less
+++ b/src/less-style/question-props.less
@@ -39,6 +39,8 @@
 .fd-question-fieldset {
   border: none;
   position: relative;
+  min-width: 0px; /* for chrome */
+  display: table-column; /* for firefox */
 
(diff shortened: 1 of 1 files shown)


millerdev:
Interesting, `min-width` and `display: table-column` are not styles I would think of to solve that problem (not that I have any better ideas).

emord:
Yeah. I'm not even entirely sure why display:table-column works, but I found it here: http://stackoverflow.com/questions/1673346/fieldset-firefox-overflow-css-fix

  (min-width works because of a chrome user agent styling)

This one was definitely a PITA to figure out.

PR #451, opened 2015-06-16
Help on focus

orangejenny:
http://manage.dimagi.com/default.asp?170957

Should wait on the [corresponding HQ PR](https://github.com/dimagi/commcare-hq/pull/7126).

@sravfeyn / @emord 

cc @amsagoff

### DIFF
--- a/src/commtrack.js
+++ b/src/commtrack.js
@@ -68,8 +68,7 @@ define([
             help: {
                 title: "Basic",
                 text: "<p>The <strong>Question ID</strong> is an internal identifier for a question. " +
-                    "It does not appear on the phone. It is the name of the question in data exports.</p>" +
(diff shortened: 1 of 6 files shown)


sravfeyn:
:+1:

millerdev:
Is it possible to make the popup disappear when clicking on the question mark a second time? For example:

- Click question mark (do not move the mouse).
- Help popup is displayed.
- Click question mark (again).
- Help popup is hidden.

orangejenny:
No. The options are
- hover
- click, where the popup is only closed if you click the icon again
- focus, where the popup is closed if you click anywhere other than the icon

I experimented with using both focus and click, hoping that would behave as you're describing, but it doesn't work.

emord:
:+1: from me whenever hq gets merged

PR #453, opened 2015-06-17
Convert last autocomplete to use at.js

orangejenny:
Will actually drop jQuery UI in another PR, not enough time to finish that right now.

@sravfeyn / @emord

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -13,7 +13,7 @@ define([
     RegExp.escape = function(s) {
         return s.replace(/[-\/\\^$*+?.()|[\]{}]/g, '\\$&');
     };
-    
+
     $.fn.stopLink = function() {
         // stops anchor tags from clicking through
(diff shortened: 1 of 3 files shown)


sravfeyn:
:+1: from me, leave it to @emord

src/util.js
```diff
@@ -373,6 +373,31 @@ define([
         return div.html();
     };
 
+    that.dropdownAutocomplete = function (input, sources) {
```
millerdev: Might be nice to add docstrings to `dropdownAutocomplete` and `questionAutocomplete` to make it easier to tell the purpose of each since their names are so similar.

millerdev:
Reviewed. Looks good. Do you want to add docstrings in this PR? If not, feel free to self merge.

orangejenny:
@millerdev good point, added docstrings

PR #455, opened 2015-06-19
Don't register ajax start

emord:
@orangejenny 

This was the bug Noah ran into:

1. load form
2. make change to form
3. save
4. click lookup table data
5. ajaxstart calls show waiting modal and there's nothing to close it.

I'm reluctant to reintroduce the waiting dialog to datasources since it's going to be refactored heavily soon by @millerdev 

cc: @biyeun

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1799,9 +1799,7 @@ define([
 
         var url = saveType === 'patch' ?  opts.patchUrl : opts.saveUrl;
 
-        $(document).ajaxStart(function () {
-            _this.showWaitingModal();
-        });
+        _this.showWaitingModal();
 
         if (saveType === 'patch') {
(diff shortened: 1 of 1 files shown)


src/core.js
```diff
@@ -1837,7 +1835,6 @@ define([
                             // unconditionally overwrite if no xform to compare
                             _this.send(formText, 'full');
                         } else {
-                            _this.closeModal();
```
orangejenny: second thoughts after merging: why remove this?

emord: clean up. https://github.com/dimagi/Vellum/blob/master/src/core.js#L670

I would like to clean up all the random modals in src/core and put them into their own module. That would make it a lot easier to use these in other areas (like adding a waiting dialog to any future ajax calls)

PR #456, opened 2015-06-20
drop jquery ui

orangejenny:


orangejenny:
Hmm, this built successfully locally, and test output doesn't reference jQuery UI. Will look into test failure.

Makefile
```diff
@@ -28,16 +28,13 @@ _rjs:
...
-	   bower_components/jquery-ui/themes/redmond/images/*.gif \
-	   _build/bower_components/jquery-ui/themes/redmond/images
 # combine CSS files (and adjust location for relative image paths)
 	# TODO do we need a blank line between the files? doesn't seem like it after initial test
 	cat _build/src/local-deps.css _build/src/main-components.css > _build/style.css
 	rm _build/src/local-deps.css _build/src/main-components.css
 	# for some reason relative image paths are wrong, so move stuff around
-	mv _build/src/global-deps.css _build/src/images _build/
+	# necessary iff global-deps includes any CSS
+	#mv _build/src/global-deps.css _build/src/images _build/
```
emord: Maybe we should remove these two lines as global-deps doesn't have any output anymore?

build.js
```diff
@@ -64,11 +64,7 @@
             name: 'global-deps',
```
emord: should create be false as it no longer has any output? (I'm not super familiar with r.js build system so I'm not 100%)

emord:
Thoughts on removing: jquery.bootstrap-better-typeahead Looks like it was included by Mike White a long time ago in the first itemsets implementation and isn't used.

build.js
orangejenny: ? it's still outputting jquery & jquery.bootstrap, isn't it?

orangejenny:
sure, I'm pro-removing just about anything we aren't using

build.js
emord: d'oh forgot it also has a js component even though the css doesn't have output anymore

bower.json
```diff
@@ -39,7 +38,6 @@
         "XMLWriter": "dimagi/XMLWriter#master",
         "MediaUploader": "dimagi/MediaUploader#master",
         "xpath": "dimagi/js-xpath",
-        "bootstrap-better-typeahead": "ptnplanet/Bootstrap-Better-Typeahead",
```
millerdev: is this being replaced by At.js? if not, is it unused? (doesn't seem related to jquery-ui, so unclear why it's being removed in this PR)

millerdev: heh, just realized @emord already commented/answered

PR #457, opened 2015-06-20
drop fancybox

orangejenny:
Don't really need to load an entire library for the sake of one spinner image.

Replaced the spinner with a "Loading" modal like the one used while saving a form - less pretty, but it blocks user from interacting with screen while page loads.


src/core.js
```diff
@@ -1071,7 +1058,7 @@ define([
                 _this.data.core.formLoadingFailed = true;
                 _this.data.core.failedLoadXML = formString;
 
-                $.fancybox.hideActivity();
+                _this.closeModal();
```
emord: This is going to close the modal that pops up telling the user they have an error.

I think the best solution may be to modify closeModal to conditionally take in an argument that will close a specific modal. You may need to add unique ids to each modal though.

orangejenny: Indeed, it will. Fixed.

I agree that the modal code could use refactoring. Any idea why we currently have an fd-modal-container and then another container for multimedia, with one modal for each media type inside that? I'd think we'd be fine with only ever having one modal on the page, appending it to the body (or $f), and having it remove itself when hidden (with the code that created it also responsible for removing it).

millerdev:
> Replaced the spinner with a "Loading" modal like the one used while saving a form

Hmm, I'm not convinced this is an improvement. It feels worse UX-wise.

What do you think about putting a spinner image in our repo and rolling our own spinner/overlay? Alternately we could use something like [spin.js](http://fgnass.github.io/spin.js/). Spin.js is 4K minified; couldn't easily find the minified size of fancybox.

Also, can't merge.

orangejenny:
@millerdev 

Sure, I can do a spinner - it is a bit weird when the modal pops up with text, but so briefly you can't read it. I'm inclined to also replace the "Processing..." dialog on save, which seems like a similar action, do you agree?

If it's just this one use case (single spinner + modal over full screen) I'll probably just steal HQ's spinner image & use bootstrap for the overlay. Agree that spin.js is a good library if we use spinners more.

millerdev:
> I'm inclined to also replace the "Processing..." dialog on save, which seems like a similar action, do you agree?

Great idea!

> I'll probably just steal HQ's spinner image

Fine with me.

orangejenny:
@millerdev 

Changed to spinner, tested (page load, saving, editing XML, overwrite warning) on staging

src/core.js
```diff
@@ -689,15 +687,14 @@ define([
...
-            showCloseButton : true,
-            onClosed: function() {}
-        });
+    fn._showPageSpinner = function() {
+        var spinner = $("<div><div><div></div></div></div>");
+        spinner.addClass("fd-form-saving");
+        $('body').append(spinner);
+    };
+
+    fn._hidePageSpinner = function() {
```
millerdev: Any particular reason you made these "private" (leading underscore)? I generally avoid using that pattern for making things private in Vellum (never quite understood the need for this kind of [fake privacy](https://github.com/dimagi/Vellum/blob/b7903bf/src/base.js#L34-L44)). Instead, if it really needs to be private, define them in the closure scope (don't assign to `fn`).

Incidentally, this does seem like something a plugin might want to override, maybe? Although an argument could be made for keeping it private until we know we need that.

orangejenny: I do prefer to keep things private until they're needed publicly. But you're right, real privacy is better than fake privacy. Updated.

PR #458, opened 2015-06-22
No placeholder

emord:
http://manage.dimagi.com/default.asp?169124#959318

@millerdev 

@biyeun

### DIFF
--- a/src/javaRosa.js
+++ b/src/javaRosa.js
@@ -891,7 +891,7 @@ define([
         widget.defaultLang = Itext.getDefaultLanguage();
         widget.isDefaultLang = widget.language === widget.defaultLang;
         widget.isSyncedWithDefaultLang = false;
-        widget.hasNodeIdPlaceholder = options.path === 'labelItext';
+        widget.hasNodeIdAsDefault = options.path === 'labelItext';
 
         widget.getControl = function () {
             return $input;
@@ -954,7 +954,7 @@ define([
                     if (!widget.isDefaultLang) {
                         value = widget.getItextValue(widget.defaultLang);
                     } else {
-                        value = widget.hasNodeIdPlaceholder ? widget.mug.p.nodeID : "";
+                        value = widget.hasNodeIdAsDefault ? widget.mug.p.nodeID : "";
                     }
                 }
 
@@ -995,7 +995,7 @@ define([
             return null;
         };
 
-        if (widget.hasNodeIdPlaceholder && widget.isDefaultLang) {
+        if (widget.hasNodeIdAsDefault && widget.isDefaultLang) {
             widget.mug.on('property-changed', function (e) {
                 if (e.property === "nodeID") {
                     if (widget.getItextValue() === e.previous || !widget.getValue()) {
@@ -1010,7 +1010,7 @@ define([
             widget.mug.on('defaultLanguage-itext-changed', function (e) {
                 if (e.form === widget.form && e.itextType === widget.itextType) {
                     var placeholder = e.value;
-                    if (!_.isString(placeholder) && widget.hasNodeIdPlaceholder) {
+                    if (!_.isString(placeholder) && widget.hasNodeIdAsDefault) {
                         placeholder = widget.mug.p.nodeID;
                     }
                     if (widget.getItextValue() === e.prevValue) {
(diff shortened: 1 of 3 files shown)


src/javaRosa.js
```diff
@@ -951,20 +948,18 @@ define([
                 widget.setValue(defaultValue);
                 widget.handleChange();
             } else {
-                var itextItem = widget.getItextItem();
+                var value = widget.getItextValue();
 
-                if (!itextItem) {
-                    widget.setValue("");
-                    return;
+                if (!_.isString(value)) {
```
millerdev: Is this catching any non-string values other than `null` and `undefined`?

src/javaRosa.js
```diff
@@ -992,29 +987,20 @@ define([
...
         widget.getDefaultValue = function () {
             return null;
         };
 
-        if (widget.hasNodeIdPlaceholder && widget.isDefaultLang) {
+        if (widget.hasNodeIdAsDefault && widget.isDefaultLang) {
             widget.mug.on('property-changed', function (e) {
                 if (e.property === "nodeID") {
-                    widget.setPlaceholder(e.val);
                     if (widget.getItextValue() === e.previous || !widget.getValue()) {
```
millerdev: I think `|| !widget.getValue()` should be removed.

- Create text question
- Clear default language label
- Change question ID

Expected: label remains empty
Observed: label changes to question ID

millerdev: Please add a test for this too.

src/javaRosa.js
```diff
@@ -1023,15 +1009,14 @@ define([
         if (!widget.isDefaultLang) {
             widget.mug.on('defaultLanguage-itext-changed', function (e) {
                 if (e.form === widget.form && e.itextType === widget.itextType) {
-                    var placeholder = e.value;
-                    if (!placeholder && widget.hasNodeIdPlaceholder) {
-                        placeholder = widget.mug.p.nodeID;
+                    var value = e.value;
+                    if (!_.isString(value) && widget.hasNodeIdAsDefault) {
```
millerdev: When is `e.value` not a string? Seems like this branch might never execute.

src/javaRosa.js
emord: It only catches `undefined`. I originally used `!_.isUndefined` but like that this makes it obvious we only expect strings

src/javaRosa.js
emord: This gets hit in phantomjs, but not chrome/firefox. So I'm leaving it in here, but added a comment about why it's there.

src/javaRosa.js
emord: Nice catch. thanks!

src/javaRosa.js
millerdev: Interesting. I have the opposite reaction to `isString` vs `isUndefined` since the latter is more specific and will not discard the value if, for example, it happens to be an integer. Not sure how that would happen, and it would probably be a bug anyway, so not a big deal.

However, this does seem a little fragile since it is on the critical path for setting the default value of the widget, and that depends on the Itext item's default value being `undefined`. Seems like we may want to do this default itext value initialization somewhere else. Maybe in `ItextModel.updateForNewMug`? ... hmm, looks like it's already being done there. Is this code even necessary?

src/javaRosa.js
millerdev: All tests pass on my machine with this `if` block removed. What am I missing?

src/javaRosa.js
emord: It's undefined for constraint, hint and help messages for a new mug. https://github.com/dimagi/Vellum/blob/bec373ab8ac1ceb586ed6c05a377a11289de8ead/src/javaRosa.js#L329-L342.

emord: I did some more digging. In a test case like this: https://github.com/dimagi/Vellum/blob/bec373ab8ac1ceb586ed6c05a377a11289de8ead/tests/static/javaRosa/test-xml-1.xml#L25-L34 there isn't an itext for question1 so this will create the itext for hin with the same value as en. This could probably use some sort of refactor, but I think it may be out of scope for this particular fix.

PR #461, opened 2015-06-24
Fix draggable divider since jquery-ui removal

millerdev:
This did not break prod or staging because jquery-ui is loaded by HQ.

@orangejenny 
cc @proteusvacuum

### DIFF
--- a/src/jquery-extensions.js
+++ b/src/jquery-extensions.js
@@ -1,0 +1,23 @@
+define(["jquery"], function ($) {
+    if (!$.fn.disableSelection) {
+        // stolen from jquery-ui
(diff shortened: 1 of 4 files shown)


emord:
I started jquery-extensions file in one of my other branches (not on master) https://github.com/dimagi/Vellum/blob/f30408761c57884ae76f479057f696fd0040cfbd/src/jquery-extensions.js

Seems like a nice place for any jquery overrides we may have. (I think there may be a couple more in src/util.js)

millerdev:
@emord updated

orangejenny:
thanks for fixing, Daniel

PR #463, opened 2015-06-25
Fix no placeholder

emord:
The bug:
The non-default language is given in the langs array first
There is no constraint message (or if there is, there is no hint message or help message, etc)
The `constraintMsg` will have an `ItextItem`, but the value will be `undefined`.
If it's the default language it will be set to `nodeID || ""`.
If it's not the default language it will be set to `ItextValue` which is `undefined`
This then trickles down to markdown which expects a string and not undefined.

https://github.com/dimagi/Vellum/blob/bec373a/src/javaRosa.js#L953-L978

In the old javarosa before #458 itext the value was guaranteed to be set to a string. 

cc: @biyeun

### DIFF
--- a/src/javaRosa.js
+++ b/src/javaRosa.js
@@ -961,7 +961,7 @@ define([
 
                 if (!_.isString(value)) {
                     if (!widget.isDefaultLang) {
-                        value = widget.getItextValue(widget.defaultLang);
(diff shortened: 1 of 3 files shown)


orangejenny:
looks good once tests pass

PR #464, opened 2015-06-25
Validation media

emord:
Add media to validation messages

@millerdev 

cc: @biyeun

### DIFF
--- a/src/javaRosa.js
+++ b/src/javaRosa.js
@@ -792,7 +792,8 @@ define([
     };
 
     var itextMediaBlock = function (mug, options) {
-        var block = itextConfigurableBlock(mug, options);
+        var block = itextConfigurableBlock(mug, options),
+            pathPrefix = options.pathPrefix || "";
 
         block.getForms = function () {
             return _.intersection(block.activeForms, block.forms);
@@ -802,31 +803,7 @@ define([
         // only hand-made forms will ever end up with a different
         // ID (the ability to set it in the UI has been broken for
         // a while), it seemed ok to make it just 'data'
-        block.itextWidget = itextMediaWidget(mug.form.getBasePath());
-
-        return block;
-    };
-
-    var itextMediaHelpBlock = function (mug, options) {
-        var block = itextConfigurableBlock(mug, options);
-
-        block.getForms = function () {
-            return _.intersection(block.activeForms, block.forms);
-        };
-
-        block.itextWidget = itextMediaWidget('/help' + mug.form.getBasePath());
-
-        return block;
-    };
-
-    var itextMediaConstraintBlock = function (mug, options) {
-        var block = itextConfigurableBlock(mug, options);
-
-        block.getForms = function () {
-            return _.intersection(block.activeForms, block.forms);
-        };
-
-        block.itextWidget = itextMediaWidget('/constraint' + mug.form.getBasePath());
+        block.itextWidget = itextMediaWidget(pathPrefix + mug.form.getBasePath());
 
         return block;
     };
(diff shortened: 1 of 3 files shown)


src/javaRosa.js
```diff
@@ -792,7 +792,12 @@ define([
     };
 
     var itextMediaBlock = function (mug, options) {
-        var block = itextConfigurableBlock(mug, options);
+        var block = itextConfigurableBlock(mug, options),
+            pathPrefix = options.pathPrefix;
+
+        if(!_.isString(options.pathPrefix)) {
```
millerdev: Is this another case of detecting null or undefined? I'm finding this pattern of using `_.isString` instead of `_.isUndefined` or `=== null` to raise more questions than in answers, but maybe I'm being silly?

src/javaRosa.js
```diff
@@ -1980,6 +1973,24 @@ define([
                 widget: iTextIDWidget,
                 widgetValuePath: "constraintMsgItext"
             };
+            databind.constraintMediaIText = function (mugOptions) {
+                return mugOptions.isSpecialGroup ? undefined : {
```
millerdev: it's a little dangerous to reference `undefined` by name like this because it can be globally reassigned (is not guaranteed to actually have an undefined value). A very unhandy feature of JavaScript. Probably better to use `null`.

millerdev: Ignore my previous comment. It seems safe to say all browsers we target are ES5 compliant, and therefore `undefined` is immutable. I guess that means we can stop using `_.isUndefined(x)` in favor of `x === undefined`, which I like much better. http://stackoverflow.com/a/8783528/10840

emord: Yeah. We also use it in other places (which is where I got this convention from) https://github.com/dimagi/Vellum/blob/97ebfd6860969dc891362dab7f9a1a092408a224/src/javaRosa.js#L2071

src/javaRosa.js
emord: Eh. I suppose that technically concatenating strings with booleans and numbers is legal in javascript, but since this is getting appended in a uri a string would be best and more clearly states intent

millerdev: Trying to pin down why this is problematic. A non-string value other than `null` or `undefined` is probably a bug, and using `_.isString` means it will be silently discarded/overwritten. The alternate check for `null` and/or `undefined` is more strict; other (likely buggy) non-string values will pass on through to a place where they will either cause an error (best outcome) or hopefully be displayed in a place where a user can see it and report a bug.

src/javaRosa.js
millerdev: heh, who wrote that code? blame says mwhite, but I didn't dig deeper to see if he copied it from somewhere. It's probably just my personal sense of style, but I don't like to explicitly return `undefined` from a function that is intended to return a value (that's what `null` is for). I can't really think of a valid reason other than something like:
```js
function getValue(arg) {
    return arg < 0 ? undefined :  arg + 1;
}

var props = {value: getValue(-1)}

// is this confusing?
props.value === undefined       // true
props.val === undefined         // true
props.hasOwnProperty("value")   // true
props.hasOwnProperty("val")     // false
```
You can ignore this comment too. It's just me ranting about silly differences between `null` and `undefined`.

src/javaRosa.js
emord: Ha. I can see how that is an issue.

When I thought about it, I looked at it the other way. 

If `options.pathPrefix = true` then the upload path is `jr://blah/en/true/data/questionid`, which is a completely valid path and since a basic user would not look at any uri they would have no idea there was a bug (since it isn't really one), then later when we see a bug report we have no idea where `true` is coming from.

millerdev: I wouldn't ever expect a basic user to detect any of the bugs we're talking about. However, many other values (like an object or array) would be glaringly obvious to a developer.

PR #465, opened 2015-06-25
upload xml file in standalone

emord:
image

cc: @biyeun

### DIFF
--- a/index.html
+++ b/index.html
@@ -42,7 +42,7 @@
             <h4>Reload using last saved form</h4>
           </button>
           <div>
-            <input type="file" class="btn" id="file" />
+            <input type="file" id="file" />
           </div>
         </div>
         <div class="span10">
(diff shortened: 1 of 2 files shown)


millerdev:
Is that a button in a button?

What's the difference between this and pasting the file content into Tools > Edit Source XML?

emord:
Removed btn class

image

There's not really a difference, but it's a lot faster to just upload a file than open it in a text editor and copy/paste it into there

millerdev:
lol, what text editor do you use?

emord:
vim, but until I can integrate vim into the web browser and use ggyG p to paste a file it's a lot easier to choose file -> select file than use mouse to copy xml -> click Edit source XML -> select all -> paste -> click update source

(can't always use reload last saved form because vellum can change ordering in setvalues, binds, etc. especially if you're changing thing in src and need to reload the page)

millerdev:
>> lol, what text editor do you use?
>
> vim...

No worries, I was just kidding around. Good to merge when tests pass.

PR #466, opened 2015-06-25
Use _ in export translations

emord:
http://manage.dimagi.com/default.asp?173827

Seems silly that we don't use the same format as HQ. 

cc: @biyeun

### DIFF
--- a/src/javaRosa.js
+++ b/src/javaRosa.js
@@ -1251,7 +1251,7 @@ define([
 
         function makeHeadings(languages, forms) {
             return rowify("label", languages, forms, function (language, form) {
-                return form + '-' + language;
+                return form + '_' + language;
             });
         }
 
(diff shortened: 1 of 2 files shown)


src/javaRosa.js
```diff
@@ -1251,7 +1251,7 @@ define([
 
         function makeHeadings(languages, forms) {
             return rowify("label", languages, forms, function (language, form) {
-                return form + '-' + language;
+                return form + '_' + language;
```
millerdev: I think this will break imports. See [line 1205 in `parseXLSItext`](https://github.com/dimagi/Vellum/blob/export-trans/src/javaRosa.js#L1205) above.

Also, should probably make `parseXLSItext` handle both _ and - so that previously exported forms will still be importable. It seems to me that HQ's format was conceived after the form builder. Maybe HQ's import/export code would be a better place to fix this? (and I personally prefer - to _).

emord: I updated this to split on both

PR #467, opened 2015-06-26
Refactor drag/drop API to be more generic

millerdev:
Delegate handle drop to the tree node so we can drag from different types of trees (in the future). This refactor is preparation for case management.

cc @proteusvacuum

### DIFF
--- a/src/javaRosa.js
+++ b/src/javaRosa.js
@@ -1296,6 +1296,16 @@ define([
         }
     }
 
+    function insertOutputRef(vellum, target, path, mug, dateFormat) {
+        var output = getOutputRef(path, dateFormat),
+            form = vellum.data.core.form;
+        util.insertTextAtCursor(target, output, true);
+        if (mug) {
+            vellum.warnOnCircularReference('label', form, mug, path, 'output value');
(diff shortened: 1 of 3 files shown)


src/javaRosa.js
```diff
@@ -1296,6 +1296,16 @@ define([
         }
     }
 
+    function insertOutputRef(vellum, target, path, mug, dateFormat) {
+        var output = getOutputRef(path, dateFormat),
+            form = vellum.data.core.form;
+        util.insertTextAtCursor(target, output, true);
+        if (mug) {
+            vellum.warnOnCircularReference('label', form, mug, path, 'output value');
```
emord: `warnOnCircularReference` technically does not need `form`. Could refactor that as well

src/javaRosa.js
```diff
@@ -1296,6 +1296,16 @@ define([
...
+            warnOnNonOutputableValue(form, mug, path);
```
emord: I think this one could be reworked into `dnd_stop`, but that might be out of scope for this PR

emord:
Merging this one. My comments are non blocking, feel free to refactor in another PR.

PR #468, opened 2015-06-26
Refactor async data sources

millerdev:
Clean up async code and update to new data sources API in preparation for case management.

Depends on https://github.com/dimagi/commcare-hq/pull/7254

cc @proteusvacuum

### DIFF
--- a/src/form.js
+++ b/src/form.js
@@ -44,8 +44,7 @@ define([
     };
     
     function processInstance(instance) {
-        instance.defaultId = instance.defaultId || convertToId(instance.sourceUri);
-        instance.id = instance.defaultId;
+        instance.id = instance.id || convertToId(instance.uri);
         _.each(instance.levels, function (level) {
             var i = 1,
                 mappedSubsets = {};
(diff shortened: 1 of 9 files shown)


src/form.js
```diff
@@ -188,7 +187,7 @@ define([
         }
         return that;
     };
-    var INSTANCE_REGEXP = /^instance\((['"])([^'"]+)\1\)/i;
+    var INSTANCE_REGEXP = /(^)instance\((['"])([^'"]+)\2\)/i;
```
emord: Why not use (?:) ?

millerdev: This group is functionally unnecessary in this regexp. I added it to make the relative position and number of groups match the [regexp below on line 341](https://github.com/dimagi/Vellum/pull/468/files#diff-62acef2590804164cb77b4b5aced00ebR341), allowing either to be used in `updateInstanceQuery`. Notice the replacement string on [line 344](https://github.com/dimagi/Vellum/pull/468/files#diff-62acef2590804164cb77b4b5aced00ebR344) references it.

PR #469, opened 2015-06-26
Case management in form builder - phase 1 milestone 2

millerdev:
Add data browser plugin, which displays a tree of session properties (currently only the case). Properties can be dragged into logic expressions. The plugin is disabled by default. See also [requirements](https://docs.google.com/document/d/1B5-fPgW-as-Cfo4vQDMnvLfGlCHkt2GgovT711-10Xc/edit)

Merging into "databrowser-dev" branch, which will subsequently be merged into master as soon as the PRs it [depends](https://github.com/dimagi/Vellum/pull/467) [on](https://github.com/dimagi/Vellum/pull/468) are merged.

cc @proteusvacuum


emord:
Haven't reviewed yet, just noticed this:

image

from @ctsims on skype:

[9:20:03 AM] Clayton Sims: [6/25/2015 5:55:29 PM] Daniel Miller: is it a valid use case to have circular relationships between cases via indexes like child/index/parent -> mother, mother/index/first-child -> child?

<<< I think on child inices you can't do that. 

IE:

Child -> Parent -> Child 

I think right now will cause problems in detecting when to remove things.
[9:20:07 AM] Clayton Sims: I wouldn't do it now
[9:20:16 AM] Clayton Sims: and would want to discuss long-term whether it's desirable

millerdev:
re: circular relationships. It's not something I would expect to be used often (if ever). I just wanted to make sure the form builder could handle it without crashing. Maybe we should remove the circular reference from options.js (although I think it's kind of a cool test case).

emord:
Ah ok. I'm totally fine with it being in standalone only. Assuming you removed those references in the HQ branch. It does make for some pretty awful references.

millerdev:
> Assuming you removed those references in the HQ branch.

HQ will simply return a representation of the app schema. HQ will happily return them if the app happens to contain a circular reference (which I believe it is possible to construct using the app builder). I think it's not the job of any of this code to remove such references if the user created them.

millerdev:
Just realized I still need to add a `dataSourcesFilter` function for itemsets.

millerdev:
bleh, labeled the wrong PR

ctsims:
Yeah, for clarity my expectation is definitely that there is _no
circularity_ between indexing relationships and we should assume that stays
true until we make a conscious decision otherwise and resolve what it means.

PR #470, opened 2015-06-29
Remove fancybox folder

emord:
@orangejenny 

Removing the fancybox folder. Noticed because now the deploy script wants to copy the fancybox js to HQ.

cc @biyeun


millerdev:
tests passed locally

PR #471, opened 2015-06-29
Fix paste empty itext

millerdev:
http://manage.dimagi.com/default.asp?174407

This is another side-effect of allowing empty itext.

@emord
cc @proteusvacuum

### DIFF
--- a/src/javaRosa.js
+++ b/src/javaRosa.js
@@ -1845,6 +1845,9 @@ define([
                         // a little hacky, but it's a fallback default
                         nodeID = data.id.slice(data.id.lastIndexOf("/") + 1);
                     }
+                    function str(val) {
+                        return val === null || val === undefined ? "" : String(val);
+                    }
                     _.each(languages, function (lang) {
                         var prelen = name.length + lang.length + 2,
                             regexp = new RegExp("^" +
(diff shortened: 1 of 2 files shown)


src/javaRosa.js
```diff
@@ -1857,8 +1857,8 @@ define([
                                 if (!seen.hasOwnProperty(form)) {
                                     seen[form] = true;
                                     // set default value(s) for this form
-                                    var dkey = name + ":" + dlang + "-" + form;
-                                    item.set(data[dkey] || value || nodeID, form);
+                                    var dval = data[name + ":" + dlang + "-" + form];
+                                    item.set(dval !== undefined ? dval : value, form);
```
emord: `"null"` is parsed as `undefined`?

millerdev: No, but I didn't like this because it could set the itext value to `null`. Updated.

millerdev:
This is tricky because empty itext items should not be serialized at all, and empty itext values must be preserved (i.e., not replaced with a default value like the question ID). Currently values for an empty itext items will be serialized as `null` and deserialized as `""` (empty string), which feels a little wrong but everything else I tried made lots of tests fail.

PR #473, opened 2015-06-30
Upgrade jstree

emord:
This was tagged since this: https://github.com/vakata/jstree/issues/869

There's no changelog on jstree, so it's hard to see exactly what has changed since that commit. I think it follows major.minor even though there are two numbers https://github.com/vakata/jstree/commit/a0767ceb3fffe1c0b2774d28c811116892a071b8.

I did some quick tests and it still functions as expected

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -33,7 +33,7 @@
         "At.js": "1.3.0",
         "Caret.js": "0.2.2",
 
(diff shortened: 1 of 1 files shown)


millerdev:
What do you think of doing a QA run on this?

emord:
@millerdev created one http://manage.dimagi.com/default.asp?174898#968908

emord:
@millerdev passed qa. (the bug logged there was for app builder, so not related)

PR #474, opened 2015-06-30
Non Functional Case Management UI Elements

biyeun:
@millerdev feel free to cherrypick commits that will help you flush out the case management UI

screen shot 2015-06-30 at 3 21 10 pm


amsagoff:
fwiw recommend adding a way to "minimize" the app resources window (was playing around with this yesterday and the drag was getting old).

src/less-style/lib/variables.less
```diff
@@ -11,3 +13,38 @@
 @grayDarker:            #2C2C2C;
 @blueLight:             #2787d0;
 
+// Main Colors [NEW]
```
millerdev: @biyeun after grabbing a new version of [hqstyle-core.css](https://github.com/dimagi/Vellum/blob/case-mgt-mockup/tests/lib/hqstyle-core.css) it looks like these colors are not in HQ yet. Do I need to wait until HQ gets some style updates to roll this PR out?

millerdev:
Not sure what's going on here. The APP RESOURCES header has `z-index: 1` and the menu has `z-index: 1000`. Maybe it's because the menu is in a `<div class="fd-head">` with `z-index: 1`?
screenshot 2015-07-02 13 01 57

emord:
@millerdev I think you can fix it by changing `.fd-head-external-sources` to have `z-index: 0`

millerdev:
@emord thanks! that fixed it. https://github.com/dimagi/Vellum/commit/c2f7777404da8edb46b5940a67d6446908210cfc

millerdev:
Closing since all of this has bee incorporated into https://github.com/dimagi/Vellum/pull/479

src/less-style/lib/variables.less
biyeun: they're in the bootstrap 3 styles...feel free to add them to less/includes/variables.less if all that is automated for now

PR #476, opened 2015-07-01
Fix paste with multimedia

millerdev:
http://manage.dimagi.com/default.asp?174407 strikes again

cc @proteusvacuum

### DIFF
--- a/src/javaRosa.js
+++ b/src/javaRosa.js
@@ -1838,7 +1838,7 @@ define([
                     var dlang = item.itextModel.getDefaultLanguage(),
                         languages = item.itextModel.languages,
                         nodeID = "",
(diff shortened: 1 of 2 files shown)


src/javaRosa.js
```diff
@@ -1838,7 +1838,7 @@ define([
                     var dlang = item.itextModel.getDefaultLanguage(),
                         languages = item.itextModel.languages,
                         nodeID = "",
-                        mmForms = {audio: 1, image: 1, video: 1},
+                        mmForms = _.object(SUPPORTED_MEDIA_TYPES, SUPPORTED_MEDIA_TYPES),
```
millerdev: This cleanup is unrelated to the fix.

PR #478, opened 2015-07-03
Alt+Click tree node to expand/collapse all children

millerdev:
A FM mentioned that this would ease the process of editing large forms with lots of groups.

cc @kaapstorm


kaapstorm:
Nice. (Waiting for Travis.)

PR #480, opened 2015-07-06
Fix css image locations in build output

millerdev:
cc @kaapstorm


orangejenny:
Curious in what way this was breaking - jstree images not displaying?

millerdev:
> jstree images not displaying?

No. ellipsis.png and ajax-loader.gif referenced in style.js had the wrong path, I think because of [this line](https://github.com/dimagi/Vellum/blob/c4255ccff4955e5b8f03135dc6373237f95c7cf0/Makefile#L33)

PR #483, opened 2015-07-06
Fix travis build

millerdev:
Wait for successful build on most recent commit before merging

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,5 @@ node_js:
  - "0.10"
 before_script:
  - phantomjs --version
- - npm install -g mocha-phantomjs
 script: 
  - make test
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -3,6 +3,5 @@ node_js:
  - "0.10"
 before_script:
  - phantomjs --version
- - npm install -g mocha-phantomjs
```
millerdev: Turns out this line was not needed (mocha-phantomjs is in `devDependencies` in package.json). However, it caused travis to error out because it did not have a version pin and subsequently failed when v4.0.0-beta was released a few days ago.

millerdev:
:white_check_mark:

PR #484, opened 2015-07-07
style tweaks

orangejenny:
Two minor tweaks
- scrolling panel too tall in initial load, making it hard to get at the "Advanced link"
- section-level help appearing above section titles instead of in line with them

old
screen shot 2015-07-07 at 11 52 19 am

new
screen shot 2015-07-07 at 12 00 41 pm

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1404,6 +1404,7 @@ define([
         refreshMessages();
 
         $props.show();
+        this.data.windowManager.adjustToWindow();
         this.$f.find('.fd-help a').fdHelp();
 
         this.toggleConstraintItext(mug);
--- a/src/templates/question_fieldset.html
+++ b/src/templates/question_fieldset.html
@@ -1,8 +1,8 @@
 <fieldset class="accordion-group fd-question-fieldset">
     <legend class="accordion-heading">
         <% if (help) { %>
-        <div class="fd-help">
-            <a href="#" class="help pull-right"
+        <div class="fd-help pull-right">
+            <a href="#" class="help"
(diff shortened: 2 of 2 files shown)


emord:
Looks good to me. Will wait for tests

src/core.js
```diff
@@ -1410,6 +1410,7 @@ define([
         refreshMessages();
 
         $props.show();
+        this.data.windowManager.adjustToWindow();
```
millerdev: Funny, I just fixed this bug a different way in f9767db965b5740a7702224b612cb651057209ad

orangejenny: that's cleaner, I'll pull that into this PR

src/window.js
```diff
@@ -17,6 +17,7 @@ define([
             preventDoubleScrolling(this.$f.find('.fd-scrollable'));
             $(window).resize(adjustToWindow);
             $(document).scroll(adjustToWindow);
+            $(document).ready(adjustToWindow);
```
millerdev: The reason tests failed is because this line needs to be moved down below `this.data.windowManager.offset = ...` (I also moved the two lines above it since they are similar). I moved them just above the line containing `this.adjustToWindow();` and I think if you do the same we might not have a merge conflict.

orangejenny:
@millerdev I don't think calling it on ready is enough. If you start with a blank form, then add a question, the scrolling panel is the wrong size, because it was measured before .fd-props-toolbar was part of the DOM.

Reverted the most recent commit (also, that'll teach me to submit code I haven't tested...tests pass locally now).

PR #485, opened 2015-07-07
Multiline widget

emord:
This does a couple of things. 
* generalizes the javarosa widget to a multilline widget
* multiline inherits from normal
* each language has it's own error container
* actually shows the error on label itexts now

Still broken: If you load an itext with <ouput value="."> or a display/calculate condition like `. = 'blah'`. There is no error shown. (It's always been like that)

code buddy: @gcapalbo

### DIFF
--- a/src/javaRosa.js
+++ b/src/javaRosa.js
@@ -560,10 +560,11 @@ define([
         block.refreshMessages = function () {
             // TODO improve this to display each message beside the
             // form and language to which it applies
-            if (options.messagesPath) {
-                var messages = widgets.util.getMessages(mug, options.messagesPath);
-                $messages.empty().append(messages);
-            }
+            _.each(block.languages, function(lang) {
+                var name = 'itext-' + lang + '-' + block.itextType,
+                    messages = widgets.util.getMessages(mug, name);
+                $('[name='+name+']').parent().parent().find('.messages').empty().append(messages);
+            });
         };
 
         mug.on("messages-changed",
@@ -1271,7 +1272,7 @@ define([
             form = vellum.data.core.form;
         util.insertTextAtCursor(target, output, true);
         if (mug) {
-            vellum.warnOnCircularReference('label', mug, path, 'output value');
+            vellum.warnOnCircularReference('label', mug, path, 'output value', target.attr('name'));
             warnOnNonOutputableValue(form, mug, path);
         }
     }
--- a/src/javaRosa.js
+++ b/src/javaRosa.js
@@ -872,9 +872,6 @@ define([
             });
         }
 
-        // future proof for when widgets.base sets path
-        widget.path = widget.path || options.path;
-
         widget.displayName = options.displayName;
         widget.itextType = options.itextType;
         widget.form = form || "default";
--- a/src/javaRosa.js
+++ b/src/javaRosa.js
@@ -826,20 +826,8 @@ define([
         }
         options.id = id;
 
-        var widget = widgets.base(mug, options);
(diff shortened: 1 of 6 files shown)


src/javaRosa.js
```diff
@@ -560,10 +560,11 @@ define([
         block.refreshMessages = function () {
             // TODO improve this to display each message beside the
             // form and language to which it applies
```
millerdev: Since "each language has it's own error container" is this comment now obsolete?

src/javaRosa.js
```diff
@@ -560,10 +560,11 @@ define([
...
-            if (options.messagesPath) {
-                var messages = widgets.util.getMessages(mug, options.messagesPath);
-                $messages.empty().append(messages);
-            }
+            _.each(block.languages, function(lang) {
+                var name = 'itext-' + lang + '-' + block.itextType,
+                    messages = widgets.util.getMessages(mug, name);
+                $('[name='+name+']').parent().parent().find('.messages').empty().append(messages);
```
millerdev: Would be great to do a local search rather than doing a global search on the DOM. Would `$blockUI.find('[name='+name+']')...` work?

src/widgets.js
```diff
@@ -197,6 +197,38 @@ define([
         return widget;
     };
 
+    var multilineText = function (mug, options) {
+        var widget = normal(mug, options);
+
+        widget.input = $("<textarea></textarea>")
```
millerdev: Too bad `normal` is hard-coded to create it's own `<input />` DOM element, and which this is overwriting. Wonder if it would make sense to add `options.input`, that, if present, would be assigned to `widget.input` by `normal`?

tests/core.js
```diff
@@ -375,16 +375,25 @@ require([
                 mug = util.addQuestion("Text", "text");
             });
 
-            _.each(["relevantAttr", "calculateAttr", "label"], function (attr) {
+            function testValidationError (attr, property) {
+                property = property || attr;
+                assert.deepEqual(mug.messages.get(attr), []);
```
millerdev: should this do `mug.messages.get(property)` instead? and the same for the last line in this function

millerdev:
This is a very welcome refactor. Nice work!

src/javaRosa.js
emord: Ha. that doesn't actually work. Right now this only works because `block.itextType` isn't as specific as I thought:

https://github.com/dimagi/Vellum/blob/master/src/javaRosa.js#L2024
https://github.com/dimagi/Vellum/blob/master/src/javaRosa.js#L2106
https://github.com/dimagi/Vellum/blob/master/src/javaRosa.js#L2125

Probably need to rework this to add the error on the widget level.

emord:
Each widget has it's own message container: 

image

If those two are the same then only the default label will show an error, but if they're different they will both show up. 

The whole block has it's own message  container as well, but I don't think it'll be used right now.

src/javaRosa.js
emord: https://github.com/dimagi/Vellum/commit/792d855e1d40811d1c93e303a034e6e8c63a6ab8

src/widgets.js
emord: Eh. I think it would make more sense to have `normal` not create an input at all, and if you want an `input` you would inherit from `text`, `textarea` from `multilineText`, etc.

Going to pass on that sort of refactor for this PR though

emord:
Updating PR to also handle self references in logic manager. (still doesn't handle output refs though)

PR #486, opened 2015-07-08
Do not show required checkbox for minimal label

millerdev:
cc @kaapstorm


emord:
looks good after tests

millerdev:
Shame on me for not running the tests first. Who knew there is a test for that?

PR #488, opened 2015-07-09
Validation conditions can have self references

emord:
Introduced here: https://github.com/dimagi/Vellum/pull/485

@orangejenny 

code buddy @gcapalbo

### DIFF
--- a/src/logic.js
+++ b/src/logic.js
@@ -1,14 +1,22 @@
 define([
-    'vellum/util',
     'xpath',
     'xpathmodels',
     'underscore'
 ], function (
-    util,
     xpath,
     xpathmodels,
     _
 ) {
+    var XPATH_REFERENCES = [
+        "relevantAttr",
+        "calculateAttr",
+        "constraintAttr",
+        "dataParent",
+        "repeat_count",
+        "filter",
+        "defaultValue"
+    ], NO_SELF_REFERENCES = _.without(XPATH_REFERENCES, 'constraintAttr');
+
     function LogicExpression (exprText) {
         this._text = exprText || "";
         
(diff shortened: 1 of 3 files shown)


orangejenny:
tests pass locally

millerdev:
Seems like we need to add a test for a constraintAttr with a self reference to verify that does not produce a warning. That can be done in a separate PR if you're confident this fix is working and want to prioritize the hotfix over writing the test.

src/logic.js
```diff
@@ -1,14 +1,22 @@
...
 ) {
+    var XPATH_REFERENCES = [
+        "relevantAttr",
+        "calculateAttr",
+        "constraintAttr",
+        "dataParent",
+        "repeat_count",
+        "filter",
+        "defaultValue"
+    ], NO_SELF_REFERENCES = _.without(XPATH_REFERENCES, 'constraintAttr');
```
millerdev: Just noticing that previously self references only produced a warning in [labels and `calculateAttr` and `relevantAttr`](https://github.com/dimagi/Vellum/commit/4c670bb53b05238fa3374538e24ba6d3b53505f1#diff-7eb52b366866677666470e019283c8eaL1592). So this is new to be warning for the others here. I suppose they all make sense though.

millerdev:
g2g when @orangejenny is satisfied.

orangejenny:
agree with @millerdev about additional test, otherwise looks good

src/logic.js
emord: Yeah. I looked through this list and it seemed like none of them should allow you to do that, but missed `constraintAttr`. (Would have probably caused an infinite loop with something like `dataParent`, trying to find it's own parent in the data tree)

orangejenny:
Test failures are from having deleted this branch before testing completed, lest anyone else look at this PR and get startled.

PR #491, opened 2015-07-10
Fix bug introduced by xpath parser update

millerdev:
https://github.com/dimagi/js-xpath/pull/3

@emord


millerdev:
whew, master is green again

PR #493, opened 2015-07-13
Feature flag rich text

emord:
@millerdev 

Obviously nowhere near done, but every new thing should be feature flagged out. 

There's only one commit in ckeditor-dev (https://github.com/dimagi/Vellum/commit/8b195ea) Just adds the ckeditor library, but I kept it out due to noise

Here's a list of fixes on my immediate to-do list for this:

* when using arrow keys highlight the bubble
* click on bubble then delete, popover not destroyed
* upgrade to 4.5.2 (once released) (should fix occasional drag/drop bug in chrome) 
* formatting/actions (waiting on @amsagoff , @ctsims for the spec/examples of actions we will need) 
* add to advanced xpath editing
* protect against script/style tags in itext displays
* highlight everything on focus 
* enhance the copying from default language to ignore &nbsp; etc 
* autogrow text areas to content size  

I think all these fixes should be able to go in different PRs to keep PR size down easily.

Also there aren't currently any tests for this right now. There's a little bit of weirdness with the events at the moment, but I have an idea of how to work around it.

code buddy @gcapalbo


src/mugs.js
```diff
@@ -387,7 +387,13 @@ define([
                 if (lang !== defaultLang && disp === defaultDisp) {
                     disp += " [" + defaultLang + "]";
                 }
-                return $('<div>').text(disp).html();
+                if (this.options.richtext) {
+                    disp = disp.replace(/(<\/?p>)/ig,"");
+                    disp = disp.replace(/[\r\n|\n|\r]/ig," ");
+                    return $('<div>').html(disp).html();
```
emord: Currently reintroduces: https://github.com/dimagi/Vellum/pull/349 but only for those behind the feature flag. Need to come up with a whitelist of tags that aren't escaped for the future. (probably only `span`, `i`)

src/widgets.js
```diff
@@ -343,9 +468,15 @@ define([
...
+        
+        if (mug.options.richtext) {
+            insertTpl = '<span class="label label-datanode label-datanode-internal" contenteditable=false draggable=true data-value="${name}"><i class="${icon}">&nbsp;</i>${name}<i class="close">&times;</i></span>';
+        }
 
-        util.questionAutocomplete(widget.input, mug,
-                                  {property: options.path});
+        util.questionAutocomplete(widget.input, mug, {
+            property: options.path,
+            insertTpl: insertTpl,
```
emord: Little weird, but this does pass `undefined` when `richtext == false`. That allows `_.defaults` to overwrite it in src/util

src/widgets.js
```diff
@@ -591,11 +722,50 @@ define([
...
 
+    function replaceOuputRef(form, value, withClose, noOutput) {
+        var template = "<output value=\"" + value + "\" />";
+        if (noOutput) {
+            template = value;
+        }
+        var v = value.split('/'),
+            dispValue = v[v.length-1],
+            mug = form.getMugByPath(value),
+            icon = mug ? mug.options.icon: 'fcc fcc-fd-external-case',
```
emord: Has there been a `getCaseByPath` method created? Would be nice to be able to pull info from that

src/javaRosa.js
```diff
@@ -818,56 +819,69 @@ define([
...
-                        end,
-                        match;
-                    if (e.which === 8) {
-                        match = val.substr(pos - 2, 2);
-                        if (match === outputEnd) {
-                            start = val.lastIndexOf(outputBegin, pos);
-                            end = pos;
+
+            if (EXPERIMENTAL_UI) {
+                insertTpl = '<span class="label label-datanode label-datanode-internal" contenteditable=false draggable=true data-value=\'&lt;output value="${name}" /&gt;\'><i class="${icon}">&nbsp;</i>${name}<i class="close">&times;</i></span>';
```
millerdev: long line, would be nice to have it wrapped

src/javaRosa.js
```diff
@@ -44,7 +44,8 @@ define([
             'default', 'short', 'long', 'audio', 'video', 'image'
         ],
         _nextItextItemKey = 1,
-        NO_MARKDOWN_MUGS = ['Item', 'Group', 'FieldList', 'Repeat'];
+        NO_MARKDOWN_MUGS = ['Item', 'Group', 'FieldList', 'Repeat'],
+        EXPERIMENTAL_UI;
```
millerdev: Would it be hard to move all these `EXPERIMENTAL_UI` checks into `widgets.multilineText`? the implication is that every place that currently uses `multilineText` would be switched to `richText` when experimental UI is on.

src/mugs.js
```diff
@@ -387,7 +387,13 @@ define([
                 if (lang !== defaultLang && disp === defaultDisp) {
                     disp += " [" + defaultLang + "]";
                 }
-                return $('<div>').text(disp).html();
+                if (this.options.richtext) {
```
millerdev: what's the difference between this and the experimental UI flag?

src/widgets.js
```diff
@@ -229,6 +236,118 @@ define([
...
+        function addPopovers(input) {
+            input.find('[contenteditable=false]').each(function () {
+                var $this = $(this),
+                    datavalue = $this.attr('data-value'),
+                    match = datavalue.match('output value="(.*)"'),
+                    value = match ? match[1] : $this.attr('data-value');
+                $this.popout({
+                    title: '',
+                    content: value,
+                    template: '<div contenteditable="false" class="popover"><div class="arrow"></div><div class="popover-inner"><div class="popover-content"><p></p></div></div></div>',
```
millerdev: long line

src/widgets.js
```diff
@@ -229,6 +236,118 @@ define([
...
+                $(this).find('.close').click(function() {
+                    $(_this).popout('hide');
+                    _this.remove();
+                    widget.handleChange();
+                    return false;
+                });
+            });
+        }
+
+        function toRichHtml(val, form, withClose) {
```
millerdev: how do you distinguish "rich" and "plain" HTML?

src/widgets.js
```diff
@@ -229,6 +236,118 @@ define([
...
+            return toRichText(val, form, withClose).replace(/\r\n|\r|\n/ig, '<br />');
+        }
+
+
```
millerdev: extra blank line

src/widgets.js
```diff
@@ -229,6 +236,118 @@ define([
...
+        widget.input.ckeditor().promise.then(function() {
+            editor = widget.input.ckeditor().editor;
+
+            mug.on('teardown-mug-properties', function() {
+                removePopovers(widget.input);
+                if (editor) {
+                    editor.destroy();
+                }
+            });
```
millerdev: Should this event handler be removed from the mug when the event fires? like
```js
mug.on('teardown-mug-properties', function() {
    ...
}, null, "teardown-mug-properties");
```

src/widgets.js
```diff
@@ -229,6 +236,118 @@ define([
...
+            .attr("name", widget.id)
+            .addClass('input-block-level jstree-drop');
+        if (options.singleLine) {
+            widget.input.addClass('fd-input');
+        } else {
+            widget.input.addClass('fd-textarea');
+        }
+
+        widget.input.ckeditor().promise.then(function() {
+            editor = widget.input.ckeditor().editor;
```
millerdev: This async thing feels like it's going to be a source of intermittent bugs. The widget system was not designed to handle async `getValue`/`setValue`, and I'm concerned that it's going to show up in strange places.

src/widgets.js
```diff
@@ -343,9 +468,15 @@ define([
                 }
             }, !!widget.isDisabled());
         };
+        
+        if (mug.options.richtext) {
+            insertTpl = '<span class="label label-datanode label-datanode-internal" contenteditable=false draggable=true data-value="${name}"><i class="${icon}">&nbsp;</i>${name}<i class="close">&times;</i></span>';
```
millerdev: long line

src/widgets.js
```diff
@@ -591,11 +722,50 @@ define([
...
+    function toRichText(val, form, withClose) {
+        if (!val) {return "";}
+        val = val.replace('&lt;', '<').replace('&gt;', '>');
+        var el = $('<div>').html(val);
+        el.find('output').replaceWith(function() {
+            return replaceOuputRef(form, this.attributes.value.value, withClose);
+        });
+        var paths = new logic.LogicExpression(val).getPaths();
+        _.each(paths, function(path) {
+            el.html(el.html().replace(path.toXPath(), $('<div>').append(replaceOuputRef(form, path.toXPath(), withClose, true)).html()));
```
millerdev: long line

tests/copy-paste.js
```diff
@@ -900,6 +901,7 @@ require([
         describe("with multimedia", function () {
             before(function (done) {
                 util.init({
+                features: {'experimental_ui': false},
```
millerdev: funny indentation

tests/core.js
```diff
@@ -424,20 +432,29 @@ require([
...
-                    text4: call("getMugByPath", "/data/group/text4"),
-                    hidden2: call("getMugByPath", "/data/group/hidden2"),
-                    text5: call("getMugByPath", "/data/text5"),
-                    hidden1: call("getMugByPath", "/data/hidden1")
-                };
+
+            before(function (done) {
+                util.init({
+                    features: {'experimental_ui': false},
+                    core: {
```
millerdev: This change doesn't seem necessary since the `describe` is nested in another `describe` that already has vellum initialized with experimental_ui turned off.

tests/writer.js
```diff
@@ -24,7 +24,8 @@ require([
                     onReady: function () {
                         done();
                     }
-                }
+                },
+                features: {'experimental_ui': false},
```
millerdev: Hmm, there are a lot of these. What do you think of turning experimental UI off by default for tests? Maybe that would be tricky?

src/javaRosa.js
emord: https://github.com/dimagi/Vellum/pull/493/files#diff-b860087dfb0ae7f09a955a95f517a309R298

For things like xpath I have them using single line right now so that they show up as they have been before. 

I think it may make sense to split up richtext to richsingeline/richtextarea. single line would be more for display like xpaths are now/the draggable question widget. where as textarea is where you do most of your editing like the label or advanced expression editor

src/mugs.js
emord: There's not one at the moment, but I added richtext to the mugoptions as it seemed slightly more self documenting and I was at the time considering making it possible to test case data and labels separately.

tests/writer.js
emord: I didn't do that as I thought it would serve as a good and more explicit reminder to me of what tests support the ui and which don't, and it only adds noise to this PR.

It's also easy to accidentally commit tests/options.js since we'll have experimental_ui on pretty much all the time

src/javaRosa.js
millerdev: ok, sounds good. but only if we can name it richinput instead of richsingleline :)

src/widgets.js
```diff
@@ -591,11 +728,51 @@ define([
...
+                }).append($('<i>').addClass(icon).html('&nbsp;')).append(dispValue);
+        if (withClose) {
+            richText.append($("<button>").addClass('close').html("&times;"));
+        }
+        return richText;
+    }
+
+    function toRichText(val, form, withClose) {
+        if (!val) {return "";}
+        val = val.replace('&lt;', '<').replace('&gt;', '>');
```
millerdev: It's hard for me to tell all the things that are happening here and below, but it looks similar to the [XML normalization utilities](https://github.com/dimagi/Vellum/blob/031ab9af76f987fe002c80d0d5bbc29f6ca0d369/src/xml.js#L11-L64). Would one or both of those be useful here?

millerdev:
Looking good. I have one more comment that you might want to address before I merge.

src/widgets.js
emord: Yeah. Something more obvious should be used here. All this is currently doing is replacing this data-value https://github.com/dimagi/Vellum/pull/493/files#diff-e5b3eca8407b45b132e0147b8392192cR845 and I wanted it to be specific so that I ran into less weirdness here. I plan on using something better in the future once the formats are more fleshed out.

PR #494, opened 2015-07-15
Add instance reference counting for all xpath expressions

millerdev:
This will be useful as we develop case management features.
Fixes http://manage.dimagi.com/default.asp?175208

cc @kaapstorm


src/form.js
```diff
@@ -274,23 +276,63 @@ define([
...
+            }
             if (!meta) {
-                this.instanceMetadata.push(InstanceMetadata({
+                meta = InstanceMetadata({
                     src: attrs.src,
                     id: attrs.id
-                }, null, mug || null, property));
+                }, null, mug || null, property);
+                if (!attrs.src) {
+                    meta.internal = true;
```
millerdev: "internal" instance metadata (without a `src` attribute) is not serialized to XML

src/itemset.js
```diff
@@ -191,6 +218,7 @@ define([
                     afterInsert: afterDynamicSelectInsert,
                     spec: {
                         itemsetData: itemsetDataSpec,
+                        filter: itemsetDataSpec,
```
millerdev: Previously itemset filters were not serialized for copy/paste, which means they were lost.

kaapstorm:
:+1: Would you like this eyeballed by someone more familiar with it, or are you happy for me to merge?

millerdev:
@emord should probably review before merging

src/form.js
```diff
@@ -274,23 +276,63 @@ define([
...
+            if (attrs.src) {
+                meta = _.find(this.instanceMetadata, function (m) {
                     return m.attributes.src === attrs.src;
                 });
+                if (!meta) {
+                    // attrs.src not found, try to find by id
+                    var ids = _.indexBy(this.instanceMetadata, function (m) {
+                            return m.attributes.id;
+                        });
+                    meta = attrs.id && ids.hasOwnProperty(attrs.id) ? ids[attrs.id] : null;
```
emord: Why not use `_.find`?

src/form.js
```diff
@@ -274,23 +276,63 @@ define([
...
-
-            var meta = _.find(this.instanceMetadata, function (m) {
+            var meta;
+            if (attrs.src) {
+                meta = _.find(this.instanceMetadata, function (m) {
                     return m.attributes.src === attrs.src;
                 });
+                if (!meta) {
+                    // attrs.src not found, try to find by id
+                    var ids = _.indexBy(this.instanceMetadata, function (m) {
```
emord: `_.indexBy` was introduced in 1.5.2. HQ serves up 1.4.4.

This actually works right now because `underscore` is in our local-deps. I think we should probably change the version to 1.4.4 in bower and move it from local-deps to global-deps, so that we aren't loading two versions of _

emord: Could also upgrade HQ. Doesn't look like there are too many functional changes from 1.4.4 to 1.5.2, so it'll probably be easy to track them down, but my first suggestion would probably be easier/safer

millerdev: Ah crap. I thought HQ was serving 1.5.2—I've had the 1.5.2 documentation bookmarked for a while now so I'm not tempted to use newer features.

My vote is to upgrade HQ. I'd really like to get up on the latest version, but that's going to be much harder because of backward-incompatible changes they've made.

src/form.js
millerdev: Because `ids` is used again below by `getUniqueId(attrs.id, ids)`, which can do multiple lookups. I suppose that's a sufficiently far-out edge case that `_.find` wouldn't be too terrible.

src/form.js
emord: > My vote is to upgrade HQ.

Agreed. I've also been meaning to look into lodash. It seems to have more favor from what I've seen, but I don't know much about it.

Should this PR be changed to use 1.4.4 and then can look at upgrading HQ as a separate task?

millerdev: +1 for lodash, but not sure we want to have HQ loading both.

emord: Heh. HQ is already loading 1.4.4 and 1.6.0 in different places.

I also meant upgrading from underscore to lodash on HQ. (I had the thought that they were mostly compatible, but I'm not positive on that)

src/form.js
emord: Ah ok. Missed that initially

src/writer.js
```diff
@@ -95,16 +95,18 @@ define([
...
-        if (manualChildren && instanceMetadata.children) {
-            // seriously, this is what you have to do
-            // HT: http://stackoverflow.com/questions/652763/jquery-object-to-string
-            writer.writeXML($('<div>').append(instanceMetadata.children).clone().html());
+
+    var _writeInstance = function (writer, instanceMetadata) {
+        if (!instanceMetadata.internal) {
+            writer.writeStartElement('instance');
+            _writeInstanceAttributes(writer, instanceMetadata);
+            if (instanceMetadata.children) {
```
emord: I think this could be `instanceMetadata.children.length`

tests/copy-paste.js
```diff
@@ -697,6 +697,44 @@ require([
             assert(util.getMug("select/itemset"), "itemset should not be cut");
         });
 
+        it("should copy dynamic select with itemset data and instance", function () {
+            util.loadXML("");
+            paste([
+                ["id", "type", "labelItext:en-default", "labelItext:hin-default", "instances", "itemsetData"],
```
emord: This goes for others in this file as well (I realize they aren't all the same), but why not declare these once `var data= [['id',...], [...],...]` and use it like `paste[data]` `eq(mod.copy(), data)` ?

tests/form.js
```diff
@@ -262,5 +264,116 @@ define([
             form.removeMugsFromForm(mugs);
             util.assertJSTreeState("");
         });
+
+        function assertInstanceSrc(id, form, expect, message) {
+            var xml = _.isString(form) ? form : form.createXML(),
+                $xml = $(xml),
+                result = $xml.find("instance[id='" + id + "']").attr("src");
```
emord: Doubt this is would reveal bugs, but you could check it's in `head > model`

millerdev: `head > model > ...` doesn't work (at least in phantomjs) because it doesn't like the namespaced `h:head`, but `model > ...` does.

tests/parser.js
```diff
@@ -60,9 +60,8 @@ require([
                         var mug = call("getMugByPath", "/data/state");
                         assert.equal(mug.__className, "Select");
                         var xml = call("createXML"),
-                            doc = $($.parseXML(xml)),
-                            instance = '<instance id="states" src="jr://fixture/item-list:state"></instance>';
-                        assert.operator(xml.indexOf(instance), ">", -1);
+                            doc = $($.parseXML(xml));
+                        assert.equal(doc.find("instance[id=states]").length, 1, xml);
```
emord: Just having the XML with no explanation as the error message seems bad. 

Looks like the convention is `instance not found: xml` elsewhere.

Personally I'm not a huge fan of having the entire xml in the error message if it's not a diff from expected/actual, but I suppose that's personal preference.

millerdev: Fair point. In this case the XML is better than "expected -1 to be > -1" with no other information.

PR #495, opened 2015-07-15
Remove unsupported jr:preload stuff

millerdev:
Quote from @ctsims "jr:preload is not supported and we should remove it from both the documentation and from the form builder."

cc @kaapstorm @amsagoff


kaapstorm:
Green lines are good, but I always like it when there are more red lines than green lines. #lessIsMore

PR #496, opened 2015-07-15
App Resources style

millerdev:
screenshot 2015-07-15 12 16 39

See commit messages for a description of what changed

<sub>(why do I feel the need to point that out, and a little silly for doing so at the same time?)</sub>

@emord 
cc @kaapstorm

### DIFF
--- a/src/databrowser.js
+++ b/src/databrowser.js
@@ -23,21 +23,27 @@ define([
     $.vellum.plugin('databrowser', {}, {
         init: function () {
             var vellum = this,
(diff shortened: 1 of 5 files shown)


emord:
Noticed that if you click the dropdown (with copy, expand all collapse all...) and then drag one of the dividers, then it doesn't close the dropdown until you are finished with the drag. I would have expected it to close on `mousedown`.

Anyways that's non blocking and I don't think associated with this PR.

PR #497, opened 2015-07-15
Intent refactor

emord:
This refactors app callouts to work more like other mugs

cc: @gcapalbo


src/intentManager.js
```diff
@@ -51,167 +55,98 @@ define([
...
+        intents = tree.treeMap(getIntentMugs);
+        intents.map(function (intentMug) {
+            writeXML(xmlWriter, intentMug.p);
+        });
+    }
 
+    function androidIntentAppId(mug, options) {
+        var widget = widgets.text(mug, options),
+            input = widget.input;
+        input.attr('placeholder', 'Insert Android Application ID');
```
emord: May be worth adding `options.placeholder` to `text`

src/intentManager.js
```diff
@@ -13,24 +13,27 @@ define([
...
-            unknownAttributes: {},
-            initialNodeID: nodeID
-        });
-    }
-    function ODKXIntentTag(form, data) {
-        util.BoundPropertyMap.call(this, form, data);
+    var DEFAULT_XMLNS = "http://opendatakit.org/xforms",
+        unmappedIntentTags = {},
+        INTENT_SPECIFIC_SPECS = [
+                "androidIntentAppId",
```
millerdev: extra indentation?

src/intentManager.js
```diff
@@ -51,167 +55,98 @@ define([
...
+        // make sure any leftover intent tags are still kept
+        _.each(unmappedIntentTags, function (tag) {
+            writeXML(xmlWriter, tag);
+        });
 
-        widget.save = function () {
-            if (widget.mug.intentTag) {
-                widget.mug.intentTag.setAttr('extra', widget.getValidValues());
+        var intents,
+        getIntentMugs = function(node) {
```
millerdev: funny indentation

src/intentManager.js
```diff
@@ -51,167 +55,98 @@ define([
...
+            }
+            if (mug.options.dataType === 'intent') {
+                return mug;
+            } else {
+                return null;
             }
         };
+        intents = tree.treeMap(getIntentMugs);
+        intents.map(function (intentMug) {
+            writeXML(xmlWriter, intentMug.p);
```
millerdev: Would it work to move this line up to [line 128](https://github.com/dimagi/Vellum/pull/497/files#diff-ebaaf3f85de0ada3835858976fdf5ba0R128)? Then you could ignore the `treeMap` return value.

src/intentManager.js
millerdev: +1

src/intentManager.js
```diff
@@ -51,167 +55,98 @@ define([
...
-
-        widget.save = function () {
-            if (widget.mug.intentTag) {
-                widget.mug.intentTag.setAttr('response', widget.getValidValues());
-            }
-        };
-
-        return widget;
+    function serializeAttrs(value, key, mug, data) {
+        data[key] = _.omit(mug.p[key], "");
```
millerdev: subtle change here: previously the `""` key was dropped if it's value was also `""`. Now it is unconditionally dropped. Is that intentional?

src/intentManager.js
```diff
@@ -13,24 +13,27 @@ define([
...
-            extra: {},
-            response: {},
-            unknownAttributes: {},
-            initialNodeID: nodeID
-        });
-    }
-    function ODKXIntentTag(form, data) {
-        util.BoundPropertyMap.call(this, form, data);
+    var DEFAULT_XMLNS = "http://opendatakit.org/xforms",
+        unmappedIntentTags = {},
```
millerdev: Previously an intent manager (and it's encapsulated state) was created for each new vellum instance. Now it looks like intent tags are stored in a module-level global. What do you think of moving this global state to the plugin's `data` area?

src/intentManager.js
```diff
@@ -51,167 +55,98 @@ define([
...
-        
-        var input = $("<input />")
-            .attr("name", widget.id)
-            .attr("type", "text")
-            .attr('placeholder', 'Insert Android Application ID');
+    function syncMugWithIntent (mug) {
+        // called when initializing a mug from a parsed form
+        if (mug.__className === "AndroidIntent") {
+            var nodeID = mug.p.nodeID,
+                tag = _.findWhere(unmappedIntentTags, {nodeID: nodeID});
```
millerdev: Would a hash-lookup work here?
```js
tag = unmappedIntentTags.hasOwnProperty(nodeID) ?
    unmappedIntentTags[nodeID] : makeODKXIntentTag(nodeID, null);
```

src/intentManager.js
```diff
@@ -51,167 +55,98 @@ define([
...
-        widget.save = function () {
-            if (widget.mug.intentTag) {
-                widget.mug.intentTag.setAttr('extra', widget.getValidValues());
+        var intents,
+        getIntentMugs = function(node) {
+            var mug = node.getValue();
+            if (!mug || node.isRootNode) {
+                return null;
+            }
+            if (mug.options.dataType === 'intent') {
```
millerdev: Above you use `mug.__className === "AndroidIntent"`. Why a different method here?

src/intentManager.js
```diff
@@ -219,70 +154,38 @@ define([
...
                 visibility: 'visible',
-                widget: androidIntentResponse
+                widget: widgets.baseKeyValue,
+                serialize: serializeAttrs,
+            },
+            unknownAttrs: {
+                lstring: 'Unknown',
+                visibility: 'hidden',
+                presence: 'optional',
+                widget: widgets.baseKeyValue,
```
millerdev: Do this need `lstring` and `widget` if `visibility: 'hidden'`?

src/intentManager.js
```diff
@@ -219,70 +154,38 @@ define([
...
-        intentTag: null,
         init: function (mug, form) {
+            mug.p.intentXmlns = mug.p.intentXmlns || DEFAULT_XMLNS;
         },
         spec: {
             androidIntentAppId: {
                 lstring: 'Intent ID',
                 visibility: 'visible',
                 widget: androidIntentAppId,
-                serialize: function (value, key, mug, data) {
```
millerdev: Were the changes in serialize/deserialize intended to preserve backward compatibility with older versions of Vellum? I'm not sure how much it matters, but just wanted to make sure you were aware in case any bug reports came in about that. This could happen if someone is saving copied questions in Excel, for example.

src/intentManager.js
emord: Removed. I am keeping the lstring below as I think that lstring is nice for documentation purposes.

src/intentManager.js
emord: That's always been there. https://github.com/dimagi/Vellum/blob/0fca2b6e42479124cd9a188d77de8b98d0d03d3f/src/intentManager.js#L133

It seems nice to have it that way for when we have more than one intent mug though (like for printing)

src/intentManager.js
```diff
@@ -51,167 +54,90 @@ define([
...
-            var intentTag = null;
-            _.each(that.unmappedIntentTags, function (tag) {
-                if (tag.getAttr('initialNodeID') === nodeID) {
-                    intentTag = tag;
-                }
-            });
-            return intentTag;
-        };
+    function parseIntentTags(tags) {
+        var intentTags = {};
```
emord: This does cause a slight change in functionality, but I think it was a bug before.

Before if there was an unknown intent it would go into the dictionary during parsing. If you ever used 'Edit Source XML' you could go there, delete the intent, but the intent would still be in the global state, and wouldn't get deleted.

I couldn't think of another instance where loadXML would get called twice, so I felt it was safe to make the change

src/intentManager.js
emord: I thought that the excel workflow was specifically unsupported?

I suppose I could change this to accept both forms on paste, but copy will always use the new form, and then change it to the simple on in a couple of weeks. Thoughts?

src/intentManager.js
```diff
@@ -51,167 +54,90 @@ define([
...
+            var mug = node.getValue();
+            if (!mug || node.isRootNode) {
+                return null;
             }
-        };
+            if (mug.options.dataType === 'intent') {
+                writeXML(xmlWriter, mug.p);
+            } else {
+                return null;
+            }
```
millerdev: Unnecessary else clause, and the first conditional above could be merged with the next. Could also move this into the treeMap call:
```js
tree.treeMap(function (node) {
    if (mug && mug.options.dataType === 'intent') {
        writeXML(xmlWriter, mug.p);
    }
});
```

src/intentManager.js
millerdev: Yes, it's officially unsupported, but that doesn't mean people will not report bugs if things don't work the way they expect.

I made a quick attempt at preserving the old serialization format: fa41fa2923da949ab9f8ee059d34804870a3c7fe. You can cherry-pick it if you like it. I won't be offended if you think preserving backward compatibility is not worth the added serialization complexity.

src/intentManager.js
millerdev: Looks good.

src/intentManager.js
emord: https://github.com/dimagi/Vellum/commit/da8050633f39afa697483e5710cae6a9d4b6bb93

PR #502, opened 2015-07-22
Printing

emord:
The intent manager isn't as hacky as it once was, but the uploader still feels that way. Not investing time to refactor that though, as if I did I'd rather just change it to some HTML 5 uploader.

cc: @czue

### DIFF
--- a/src/intentManager.js
+++ b/src/intentManager.js
@@ -262,6 +262,14 @@ define([
             types.normal.PrintIntent = PrintIntent;
             return types;
         },
+        getAdvancedQuestions: function () {
+            var ret = this.__callOld();
+            ret.push("AndroidIntent");
+            if (this.opts().features.printing) {
+                ret.push("PrintIntent");
+            }
+            return ret;
+        },
         getMainProperties: function () {
             return this.__callOld().concat(INTENT_SPECIFIC_SPECS);
         }
--- a/src/intentManager.js
+++ b/src/intentManager.js
@@ -127,7 +127,7 @@ define([
                 mug.p.docTemplate = mug.p.androidIntentExtra['cc:print_template_reference'];
                 delete mug.p.androidIntentExtra['cc:print_template_reference'];
             } else {
-                mug.p.docTemplate = "jr://file/commcare/html/" + mug.p.nodeID+ ".html";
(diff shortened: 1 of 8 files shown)


src/uploader.js
```diff
@@ -50,22 +52,31 @@ define([
                 'description': 'Video',
                 'extensions': '*.3gp;*.mp4'
             }
-        ]
+        ],
+        text: [
+            {
+                'description': 'HTML',
```
emord: This does seem a little wrong, but the current plan is to use only HTML files, but in the hope of keeping the HQ side a little simpler I just made a generic text upload.

src/intentManager.js
```diff
@@ -219,8 +259,17 @@ define([
...
             var types = this.__callOld();
             types.normal.AndroidIntent = AndroidIntent;
+            types.normal.PrintIntent = PrintIntent;
             return types;
         },
+        getAdvancedQuestions: function () {
+            var ret = this.__callOld();
+            ret.push("AndroidIntent");
+            if (this.opts().features.printing) {
+                ret.push("PrintIntent");
```
millerdev: What's this? Another question type in the Advanced menu?! Seems like this might be a reasonable thing to add to this list of presets being added for the area mapper spec.

src/widgets.js
```diff
@@ -517,6 +517,22 @@ define([
...
+
+        widget.getValue = function() {
+            return widget.input.val();
+        };
+
+        widget.setValue = function(val) {
+            return widget.input.val(val);
+        };
+
+        widget.mug.form.vellum.initWidget(widget);
```
millerdev: wow, that's an ugly deep reach

millerdev:
Does this need QA or UAT? If not, I think it's ready to merge.

src/widgets.js
emord: Yeah. this is one of the things I wanted to refactor with media. the name doesn't even tell you what it's doing. `initWidget`? ridiculous (method in core that's only ovewritten by uploader)

emord:
No QA/UAT needed. It's feature flagged, and I think it's only on a dev branch on mobile for Aliza to check out right now

PR #503, opened 2015-07-23
populate extra fields after html upload

emord:
Not tested as it relies on the uploader and that's a hole in vellum's tests right now.

Relies on https://github.com/dimagi/commcare-hq/pull/7620 and https://github.com/dimagi/MediaUploader/pull/9

cc: @czue

### DIFF
--- a/src/intentManager.js
+++ b/src/intentManager.js
@@ -154,22 +154,27 @@ define([
         }
     }
 
+    function parseFields (html) {
+        var field_regex = /{{\s*([^}\s]+)\s*}}/gm,
+            match = field_regex.exec(html),
+            fields = {};
+
+        while (match) {
+            _.each(match.splice(1), function(field) {
+                fields[field] = field;
+            });
+            match = field_regex.exec(html);
+        }
+
+        return fields;
+    }
+
     function printTemplate(mug, options) {
         var widget = widgets.media(mug, options),
-            uploadComplete = widget.handleUploadComplete,
-            field_regex = /{{\s*(.+)\s*}}/gm;
+            uploadComplete = widget.handleUploadComplete;
 
         widget.handleUploadComplete = function (event, data, objectMap) {
-            mug.p.androidIntentExtra = {};
-            var html = data.text,
-                match = field_regex.exec(html);
-
-            while (match) {
-                _.each(match.splice(1), function(field) {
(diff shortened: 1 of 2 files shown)


czue:
looks good to me. will let @millerdev merge

src/intentManager.js
```diff
@@ -153,6 +154,33 @@ define([
         }
     }
 
+    function printTemplate(mug, options) {
+        var widget = widgets.media(mug, options),
+            uploadComplete = widget.handleUploadComplete,
+            field_regex = /{{\s*(.+)\s*}}/gm;
```
millerdev: Can you some examples of what you're trying to match here? I think it will not do what you expect:

```js
>>> re = /{{\s*(.+)\s*}}/gm
/{{\s*(.+)\s*}}/gm
>>> str = "{{ abc }} {{ def }}"
"{{ abc }} {{ def }}"
>>> re.exec(str)
["{{ abc }} {{ def }}", "abc }} {{ def "]
>>> re.exec(str)
null
>>> re.exec(str)
["{{ abc }} {{ def }}", "abc }} {{ def "]
```

A guess that may be closer to what you want: `/{{\s*([^}]+?)\s*}}/gm`

However, this still has problems. It will not allow a `}` character inside a `{{` ... `}}` delimited range, which I assume should be legal given the double delimiter. Making a parser that would allow that is not trivial (maybe not even possible?) with a single regular expression.

This seems like a good candidate to extract into a helper function and add some tests to verify it does what you expect.

Also, it might be safer to define this regexp inside `handleUploadComplete` because the `g` flag makes it stateful. If the while loop condition ever changed it might not reset the regex state every time.

emord: > However, this still has problems. It will not allow a } character inside a {{ ... }} delimited range, which I assume should be legal given the double delimiter. Making a parser that would allow that is not trivial (maybe not even possible?) with a single regular expression.

Yeah that should probably be legal but I don't think that's really necessary to implement right now. I think this is mostly a one off for one domain. Will let @amstone326 or @ctsims comment on that though

> This seems like a good candidate to extract into a helper function and add some tests to verify it does what you expect.

Originally didn't do this because intentmanager doesn't actually return anything. However that's a silly way to think. Updated with tests

PR #504, opened 2015-07-27
feature flag + add rich text to advanced xpath editor

emord:
Going into ckeditor-dev. on staging. it's feature flagged, but I'm going to put it up for regression QA to make sure I didn't muck up anything with normal vellum. 

Next PR will be pulling out some of the utilities I built and unit testing them

cc: @czue

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -732,7 +732,7 @@ define([
         if (target) {
             // the .change fires the validation controls
             if (this.opts().features.rich_text) {
-                target.ckeditor().editor.insertHtml(widgets.util.toRichText(path, _this.data.core.form, true));
+                target.ckeditor().editor.insertHtml(widgets.util.toRichText(path, _this.data.core.form, true) + " ");
             } else {
                 target.val(target.val() + path).change();
             }
--- a/src/core.js
+++ b/src/core.js
@@ -453,7 +453,7 @@ define([
     fn.getMugDisplayName = function (mug) {
         var val = mug.getDisplayName(this.data.core.currentItextDisplayLanguage || 
             this.data.javaRosa.Itext.getDefaultLanguage());
-        if (this.opts().features.experimental_ui) {
+        if (this.opts().features.rich_text) {
             val = widgets.util.toRichText(val, this.data.core.form);
         }
         return val;
--- a/src/core.js
+++ b/src/core.js
@@ -732,7 +732,7 @@ define([
         if (target) {
             // the .change fires the validation controls
             if (this.opts().features.rich_text) {
-                target.ckeditor().editor.insertHtml(widgets.util.toRichText(path, mug.form, true));
+                target.ckeditor().editor.insertHtml(widgets.util.toRichText(path, _this.data.core.form, true));
             } else {
                 target.val(target.val() + path).change();
             }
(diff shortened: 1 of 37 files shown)


src/databrowser.js
```diff
@@ -165,7 +165,7 @@ define([
                 }
                 info = info._parent;
             }
-            vellum.handleDropFinish(target, path);
+            vellum.handleDropFinish(target, path, widget.mug);
```
millerdev: This change doesn't seem right. The thing being dropped here is a data source entity path, which I think is not related to the widget's mug. Can you explain what I'm missing?

src/widgets.js
```diff
@@ -785,6 +781,19 @@ define([
         return el.html();
     }
 
+    function fromRichText(val) {
+        var el = $('<div>');
+        val = val.replace(/(<p>)/ig,"").replace(/<\/p>/ig, "\r\n").replace(/(<br ?\/?>)/ig,"\n").replace('&nbsp;', ' ').replace('\n', ' ');
```
millerdev: Why replace with `\r\n` and then later replace `\n` with space? Seems like you could end up with `\r` in some places.

src/databrowser.js
emord: https://github.com/dimagi/Vellum/blob/master/src/core.js#L725-L745

Agreed it's not the best. It has always passed in mug and i just used that to use form.

src/widgets.js
emord: Fixed

src/databrowser.js
millerdev: Maybe I should have passed in `null` as the third argument to be more explicit. It was intentional that I did not supply a third argument in this call.

emord: I should've linked to my dev branch instead of master: https://github.com/dimagi/Vellum/blob/19f7ce7895fdf928ec7a4df4fc14cd371de5ecd6/src/core.js#L736

That trickles down to:
https://github.com/dimagi/Vellum/blob/19f7ce7895fdf928ec7a4df4fc14cd371de5ecd6/src/widgets.js#L741

If ES6 was supported I'd use default params to make the code a little cleaner, but this was a an easy way to make case properties drag/droppable with bubbles.

My next PR is to refactor/add tests to some of the current functionality and I'll change the involved functions to not rely on form

src/core.js
```diff
@@ -732,7 +732,11 @@ define([
 
         if (target) {
             // the .change fires the validation controls
-            target.val(target.val() + path).change();
+            if (this.opts().features.rich_text) {
+                target.ckeditor().editor.insertHtml(widgets.util.toRichText(path, mug.form, true));
```
millerdev: Could get the form here more reliably with `_this.data.core.form`. The problem here is that sometimes `mug` can be `null` or `undefined`, which would break this.

src/databrowser.js
millerdev: I'm sticking to my line: `widget.mug` should not be passed in here. `vellum.handleDropFinish` uses the mug (i.e., the thing being dropped) to select an operator for the expression editor.

Also, see my comment on core.js.

emord: Fixed. You can probably ignore this PR for now. I'm going to crank out tests for this stuff so I feel better putting it on prod. Will ping once done

emord:
Closing this PR as it's become quite large. I'll break it up into more easily reviewable PRs

PR #505, opened 2015-07-28
Add templates to app callout question type

millerdev:
screenshot 2015-07-28 16 48 52

This will show an empty templates selector until we get HQ to supply them in the vellum options. I'm working on that next.

cc @sravfeyn @wspride

### DIFF
--- a/src/intentManager.js
+++ b/src/intentManager.js
@@ -4,13 +4,15 @@ define([
     'vellum/util',
     'underscore',
     'jquery',
+    'tpl!vellum/templates/intent_templates',
     'vellum/core'
 ], function (
     mugs,
     widgets,
     util,
     _,
-    $
+    $,
+    intent_templates
 ) {
     "use strict";
     var DEFAULT_XMLNS = "http://opendatakit.org/xforms",
@@ -39,25 +41,17 @@ define([
         var widget = widgets.text(mug, options),
             input = widget.input,
             opts = options.vellum.opts().intents,
-            tempSelect = $('<select style="width: 100%" />')
-                .attr("name", widget.id + "-template")
-                .addClass('span3 input-block-level')
-                .append($("<option />").attr("value", "").text("")),
-            templates = {},
+            tempMenu = $(intent_templates({templates: opts && opts.templates || []})),
+            templates = _.object(_.map(opts && opts.templates, function (temp) {
+                return [temp.id, temp];
+            })),
             control = $('<div class="control-row row" />')
-                .append($('<div class="span6" />').append(input))
-                .append($('<div class="span2 control-label">Template</div>'))
-                .append($('<div class="span4" />').append(tempSelect));
+                .append($('<div class="span8" />').append(input))
+                .append($('<div class="span4" />').append(tempMenu));
 
-        _.each(opts && opts.templates, function (temp) {
-            tempSelect.append($("<option />")
-                .attr("value", temp.id)
-                .text(temp.name));
-            templates[temp.id] = temp;
-        });
-
-        tempSelect.change(function () {
-            var id = tempSelect.val();
+        tempMenu.find('.dropdown-menu a').click(function (e) {
+            e.preventDefault();
+            var id = $(this).data("id");
             input.val(id).change();
             if (id && templates.hasOwnProperty(id)) {
                 var container = input.parents(".fd-props-content").first();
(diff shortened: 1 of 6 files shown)


src/intentManager.js
```diff
@@ -35,6 +35,53 @@ define([
...
+        tempSelect.change(function () {
+            var id = tempSelect.val();
+            input.val(id).change();
+            if (id && templates.hasOwnProperty(id)) {
+                var container = input.parents(".fd-props-content").first();
+                _.each({
+                    extra: "[name=property-androidIntentExtra]",
+                    response: "[name=property-androidIntentResponse]",
+                }, function (name, key) {
+                    if (templates[id].hasOwnProperty(key)) {
```
emord: Should this overwrite without this check?

I think it would be weird if you switched from area mapper to breath counter and you still kept `r1`, `r2` etc

src/intentManager.js
```diff
@@ -35,6 +35,53 @@ define([
...
+            opts = options.vellum.opts().intents,
+            tempSelect = $('<select style="width: 100%" />')
+                .attr("name", widget.id + "-template")
+                .addClass('span3 input-block-level')
+                .append($("<option />").attr("value", "").text("")),
+            templates = {},
+            control = $('<div class="control-row row" />')
+                .append($('<div class="span6" />').append(input))
+                .append($('<div class="span2 control-label">Template</div>'))
+                .append($('<div class="span4" />').append(tempSelect));
```
emord: Lots of appends here. Maybe a template? Not sure if that would make it better

tests/intentManager.js
```diff
@@ -57,6 +59,67 @@ define([
...
+                    }
+                    if (temp.response) {
+                        var response = widgets.util.getWidget(
+                                $("[name=property-androidIntentResponse]"), vellum);
+                        assert.deepEqual(response.getValue(), temp.response);
+                    }
+                });
+            });
+        });
+
```
emord: Add a test case for an unrecognized callout.

tests/options.js
```diff
@@ -99,12 +99,37 @@ define(["underscore"], function (_) {
...
+                {
+                    name: "Barcode Scanner",
+                    id: "com.google.zxing.client.android.SCAN",
+                    extra: {},
+                    response: {},
+                },
+                {
+                    name: "Breath Counter",
+                    id: "org.commcare.respiratory.BREATHCOUNT",
+                },
```
emord: Should these be passing an icon in as well for display in the dropdown?

src/intentManager.js
millerdev: Good idea. Originally I was thinking it might be convenient to use a preset to load a config, and then modify it from there. Of course that's possible anyway, and this makes the experience smoother.

src/intentManager.js
millerdev: Yeah, I thought of that, but the inner appends (`input` and `tempSelect`) seemed harder to get into a template while retaining a references to the inserted DOM elements. Maybe that's easy and I just don't know it?

tests/intentManager.js
millerdev: Not sure what you mean. How would an unrecognized callout get in the templates dropdown?

millerdev:
Updated
screenshot 2015-07-28 23 01 09

src/intentManager.js
```diff
@@ -35,6 +37,44 @@ define([
...
+            opts = options.vellum.opts().intents,
+            tempMenu = $(intent_templates({templates: opts && opts.templates || []})),
+            templates = _.object(_.map(opts && opts.templates, function (temp) {
+                return [temp.id, temp];
+            })),
+            control = $('<div class="control-row row" />')
+                .append($('<div class="span8" />').append(input))
+                .append($('<div class="span4" />').append(tempMenu));
+
+        tempMenu.find('.dropdown-menu a').click(function (e) {
```
sravfeyn: Can't easily reason what this block is doing, can add a comment or enclose it in a smaller function? (#not-blocker)

millerdev:
I think the build failure was because I accidentally restored the branch, which kicked off another test run, and then subsequently deleted it before travis was able to checkout the branch. It was green before I deleted the branch the first time.

tests/intentManager.js
emord: Oh I meant more like load a form with an appid that wasn't recognized, but I think that is covered by other tests in this file

PR #506, opened 2015-07-29
Case management tweaks

millerdev:
- Rename "App Resources" to "Case Properties"
- On form load, "Case Properties" window expanded by default

@emord 
cc @sravfeyn


sravfeyn:
g2m by me

PR #507, opened 2015-07-30
Fix printing

emord:
cc: @czue

### DIFF
--- a/src/intentManager.js
+++ b/src/intentManager.js
@@ -217,23 +217,6 @@ define([
         return fields;
     }
 
-    function printTemplate(mug, options) {
-        var widget = widgets.media(mug, options),
-            uploadComplete = widget.handleUploadComplete;
-
-        widget.handleUploadComplete = function (event, data, objectMap) {
-            mug.p.androidIntentExtra = parseFields(data.text);
-
-            uploadComplete(event, data, objectMap);
-
-            // hack: need to refresh the current mug because we change a
-            // different property in this widget
-            refreshCurrentMug();
-        };
-
-        return widget;
-    }
-
     var AndroidIntent = util.extend(mugs.defaultOptions, {
         typeName: 'Android App Callout',
         dataType: 'intent',
@@ -310,7 +293,7 @@ define([
             docTemplate: {
                 lstring: 'Document Template',
                 visibility: 'visible',
-                widget: printTemplate,
+                widget: widgets.media,
             },
             androidIntentAppId: { visibility: 'hidden' },
             androidIntentResponse: { visibility: 'hidden' },
--- a/src/intentManager.js
+++ b/src/intentManager.js
@@ -78,19 +78,27 @@ define([
     var parseInnerTags = function (tagObj, innerTag) {
         var store = {};
         _.each(tagObj.find(innerTag), function (inner) {
-            var $innerTag = $(inner);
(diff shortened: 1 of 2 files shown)


tests/static/intentManager/printing-intent.xml
```diff
@@ -24,7 +24,7 @@
 		</model>
 		<odkx:intent xmlns:odkx="http://opendatakit.org/xforms" id="print_data" class="org.commcare.dalvik.action.PRINT">
 			<extra key="text" ref="/data/text" />
-			<extra key="cc:print_template_reference" ref="jr://file/commcare/doc/data/print_data.doc" />
+			<extra key="cc:print_template_reference" ref=".">jr://file/commcare/doc/data/print_data.doc</extra>
```
emord: `ref` can only be an xpath and can't actually contain the literal of the file name

Block on @ctsims saying this is ok though.

src/intentManager.js
```diff
@@ -209,23 +217,6 @@ define([
         return fields;
     }
 
-    function printTemplate(mug, options) {
-        var widget = widgets.media(mug, options),
-            uploadComplete = widget.handleUploadComplete;
-
-        widget.handleUploadComplete = function (event, data, objectMap) {
```
emord: The html files being uploaded are generated by word and it's crashing the flash uploader without a traceback. Apparently https://github.com/dimagi/MediaUploader/pull/9 was not enough. Removing this functionality for now

czue:
:+1: code wise. will wait on feedback from @ctsims to merge

millerdev:
reviewed. looks good to me

emord:
Debugged this code on mobile with @amstone326 and @phillipm for a while yesterday. This should be good to go now

src/intentManager.js
```diff
@@ -108,8 +116,8 @@ define([
         writeInnerTagXML(xmlWriter, 'extra', properties.androidIntentExtra);
         if (properties.docTemplate) {
             writeInnerTagXML(xmlWriter, 'extra', {
-                'cc:print_template_reference': properties.docTemplate
-            });
+                'cc:print_template_reference': '.'
```
phillipm: The '.' might have to change to './text()' or something else, depending on how @ctsims weighs in.

Right now this will create the tag:
```
<extra key="cc:print_template_reference" ref=".">jr://file/commcare/text/question1.html</extra>
```
On the mobile side we were having trouble resolving/contextualizing the '.' xpath expression. Currently we have a hack to grab the node's text, if the key was 'cc:print...', instead of contextualizing the ref. This is unsustainable and we should figure out the correct xpath expression to put here such that the node's value is used.

I'm going to take a look at why the contextualization of '.' is causing issues.

phillipm: It does currently works with the printer-integration branch, so this PR should be good to go as long as we remember to come back and resolve this issue.

tests/static/intentManager/printing-intent.xml
ctsims: This won't work, we don't process this as an XPath instance, we just parse it out when we parse the form.

amstone326: Clayton, was planning on talking to you about this later. @phillipm worked on this with us on Thursday and we came up with a workaround that may or may not be acceptable, but it's a much bigger lift on HQ to do it the way you were saying, so we're trying to find a way around it. The 3 of us should chat when you can

PR #508, opened 2015-08-03
Allow duplicate intent response keys

millerdev:
@dannyroberts @emord cc @sravfeyn 

Note that this also allows duplicate "extra" keys, and I'm not sure how the mobile app would handle that. But this new feature does not seem worse than what we had before, which was that one of the duplicate key/value pairs was silently discarded.


dannyroberts:
Looks great! Will let @emord review

PR #510, opened 2015-08-08
Dont use today

emord:
http://manage.dimagi.com/default.asp?178707

@czue Should people using commtrack questions rebuild their forms/be told about this change?

cc: @gcapalbo @kaapstorm


czue:
i think it's ok since as long as they're using this it will default to server time and that will be right 99% of the time. the big issue with the case management ones is mixing the dates + datetimes, but that won't happen here

czue:
thanks for super quick turnaround!

PR #512, opened 2015-08-18
Question/case "bubbles"

emord:
Tried to make this easily reviewable, but it is large. It can also be reviewed by commit.

For the majority of changes you can look here (the first few commits are just adding a new library/feature flag): https://github.com/dimagi/Vellum/compare/0eae6f0...bubbles

@orangejenny 

code buddies: @esoergel @proteusvacuum 

selection_024
selection_025


orangejenny:
took a first pass, mostly just asking for more comments. this is going to be great.

orangejenny:
good to merge from my perspective

emord:
added a couple of commits

Should be merged with this: https://github.com/dimagi/commcare-hq/pull/8207

PR #514, opened 2015-08-23
Show date drag and drop on full screen

emord:
http://manage.dimagi.com/default.asp?179699

code buddies: @proteusvacuum @esoergel

### DIFF
--- a/src/less-style/structure.less
+++ b/src/less-style/structure.less
@@ -3,6 +3,7 @@
 .fd-ui-container {
   top: 0;
   width: 100%;
+  background-color: white;
 }
 
 .fd-toolbar {
--- a/src/window.js
+++ b/src/window.js
@@ -49,27 +49,27 @@ define([
                 availableHorizSpace,
                 position = (this.getCurrentTopOffset() === 0) ? 'fixed' : 'static',
                 $fdc = this.$f.find('.fd-ui-container');
+
             if (this.data.windowManager.fullscreen) {
                 $fdc.parent().css({height: null, width: null});
                 $fdc.css({height: null, width: null});
                 $fdc.parent().css({
-                    position: 'fixed',
                     top: 0,
                     bottom: 0,
                     left: 0,
                     right: 0,
-                    zIndex: 10000,
                     backgroundColor: 'white'
                 });
+                $fdc.css('width', $(window).width());
             } else {
                 $fdc.parent().css({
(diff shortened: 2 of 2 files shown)


orangejenny:
This looks like it's becoming a z-index arms race. Any idea why full screen has the high z-index in the first place?  @dannyroberts do you remember? I poked at it for a few minutes but didn't find the reason. Did notice that the saving modals also don't work in full screen mode.

dannyroberts:
full screen has a high zindex because it is implemented by detaching the vellum main view (using css absolute/relative positioning) and it needs to appear above all other things on the page—except for dropdowns, modals etc that need to appear above the full screen view, which it sounds like I screwed that up.

orangejenny:
@dannyroberts 

> it needs to appear above all other things on the page—except for dropdowns, modals etc that need to appear above the full screen view

Yeah, that exception is a killer, isn't it? For this case, the other styles (fixed position, full screen coordinates, opaque background) look like they're enough to cover up HQ. Was just wondering if you'd run into something specific that made the z-index also necessary, or if we could just drop it.

dannyroberts:
Um, I do think there was something that at some point was awkwardly
floating on top of the full screen view, but I don't remember what that
was. Maybe you can git blame the zindex line to see if there's a commit
message explaining the reason. If it's just part of the original commit,
it's possible I'm wrong and that I just did that preemptively and forgot to
think harder about it.

emord:
Did some more digging on this one.

Looks like `.modal-backdrop`'s `z-index: 1040`(that's the dark background that's used in saving that i think @orangejenny is talking about), and `.modal`'s `z-index: 1050`. 

Our different modals are children of `.modal`, but they are covered by the `.modal-backdrop` when < 1040. Not super obvious to me why that's happening right now.

emord:
@orangejenny I think this accomplishes what we talked about earlier today

orangejenny:
looks good. you tested this within HQ, right? (i'm surprised the fixed position isn't necessary)

emord:
Well the fixed position is still there in `fd-ui-container`. The parent of that `#formdesigner` is what i removed the fixed from. So fixed is required, but it's now a sibling of the modals, so it doesn't screw with that

orangejenny:
thanks for the explanation, i wasn't looking closely enough at the unchanged code

PR #517, opened 2015-08-25
Save to case auto

emord:
http://manage.dimagi.com/default.asp?179743#1000338

- Enable the drag+drop of questions from tree into text fields (ex. when updating a case property, dragging the question in)
- Enable auto-complete of questions in the text fields

buddy: @czue


czue:
(no blockers, but will wait for an answer to the `+new` thing and tests for merge)

src/atwho.js
```diff
@@ -0,0 +1,126 @@
+define([
```
orangejenny: glad to see this pulled into its own file

PR #519, opened 2015-08-27
New validation

emord:
http://manage.dimagi.com/default.asp?180215#1003031

cc: @czue


czue:
(needs to be updated for backwards compatibility so put a don't merge on it. @emord let me know when you're ready for re-review)

emord:
Decided to just write both forms to each form

czue:
@wpride can you confirm the addition of the `<alert>` won't break the older versions as long as the attribute format is also present?

wpride:
@czue confirmed that if an <alert> block is present in a resource when an older version installs it'll just be ignored

czue:
thanks!

PR #523, opened 2015-09-01
feature flag intents

emord:
dependent on https://github.com/dimagi/commcare-hq/pull/8089

cc: @czue

### DIFF
--- a/src/intentManager.js
+++ b/src/intentManager.js
@@ -305,6 +305,18 @@ define([
         }
     });
 
+    function noIntents(features) {
+        return !features.custom_intents && !features.templated_intents;
+    }
+
+    function templatedIntents(features) {
+        return features.templated_intents && !features.custom_intents;
+    }
+
+    function intents(features) {
+        return !noIntents(features);
+    }
+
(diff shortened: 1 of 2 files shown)


src/intentManager.js
```diff
@@ -305,6 +305,18 @@ define([
         }
     });
 
+    function noIntents(features) {
+        return !features.custom_intents && !features.templated_intents;
+    }
+
+    function templatedIntents(features) {
+        return features.templated_intents && !features.custom_intents;
```
czue: is it right that `templated_intents` and `custom_intents` can't both be true?

czue: er - see they are both true in tests - is the idea that if `custom` is true then you don't show these even if they're on?

emord: if `nointents` then there will be no intent question types (community & standard)
if `templatedIntents` then there will be a template for the intent with extra and response key value fields moved to advanced
if custom intents then it will be the same as it always has been

i changed `templatedIntents` to `onlyTemplatedIntents` as that's a little more obvious

PR #525, opened 2015-09-08
Random

emord:
No warnings for using instances (not needed since https://github.com/dimagi/js-xpath/pull/3)

Parent cases will now be at the same level as child.

@orangejenny @kaapstorm

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -632,13 +632,6 @@ define([
                     label: "Form Name",
                     slug: "formName"
                 },
-                {
-                    label: "Form ID",
-                    slug: "formID",
-                    cleanValue: function (val) {
-                        return val.replace(/ /g, '_');
-                    }
-                }
             ];
 
         $modalBody.append($('<p />').text(
--- a/src/core.js
+++ b/src/core.js
@@ -634,12 +634,6 @@ define([
                 },
             ];
 
-        $modalBody.append($('<p />').text(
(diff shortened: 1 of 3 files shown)


emord:
also adding https://github.com/dimagi/Vellum/commit/754b02c7700cc66964e6ea43e5a281ffaaf95f1d

The last time that was referenced was before requirejs was used https://github.com/dimagi/Vellum/commit/754b02c7700cc66964e6ea43e5a281ffaaf95f1d

It most likely got lost in a refactor.

orangejenny:
> Edit: lost here 4c391f9

it's hard not to love a PR titled "major refactoring" with 50k lines of changes

orangejenny:
looks fine pending tests

src/core.js
```diff
@@ -632,21 +632,8 @@ define([
                     label: "Form Name",
                     slug: "formName"
                 },
```
kaapstorm: (Careful. This trailing comma means that your array might be [one item longer][1] than you expect.)

  [1]: http://stackoverflow.com/a/7246662

orangejenny: A single trailing comma is okay (in our non-IE world), it's multiples that get weird. I like trailing commas for reducing the number of lines you have to touch, though vellum uses both styles.

kaapstorm: :+1:

PR #526, opened 2015-09-10
Image resizer

emord:
http://manage.dimagi.com/default.asp?181411#1010439

Still needs to be version flagged, but this is most of it

@orangejenny @kaapstorm

### DIFF
--- a/src/mugs.js
+++ b/src/mugs.js
@@ -1212,6 +1212,30 @@ define([
         typeName: 'Image Capture',
         icon: 'icon-camera',
         mediaType: "image/*", /* */
+        spec: {
+            imageSize: {
+                visibility: 'visible',
+                widget: widgets.dropdown,
+                defaultOptions: [
+                    { text: "Small", value: "250" },
+                    { text: "Medium", value: "500" },
+                    { text: "Large", value: "1000" },
+                    { text: "Original", value: "" },
+                ]
+            }
+        },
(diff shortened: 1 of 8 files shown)


emord:
to be merged with https://github.com/dimagi/commcare-hq/pull/8224

src/mugs.js
```diff
@@ -1508,6 +1546,7 @@ define([
             // Nestable Field List not supported in CommCare before v2.16
             group_in_field_list = opts.features.group_in_field_list,
             richText = opts.features.rich_text;
+        ENABLE_IMAGE_RESIZER = opts.features.image_resize;
```
orangejenny: Is it feasible to set a `resizeEnabled` flag on the Image mug instead of using a global flag?

src/widgets.js
```diff
@@ -653,6 +657,14 @@ define([
             });
         };
 
+        if (options.defaultOptions) {
+            widget.addOptions(options.defaultOptions);
+
+            if (options.defaultValue) {
```
orangejenny: Not quite following why this was added...should the spec in mugs.js include a defaultValue?

orangejenny:
looks good, a couple of minor questions

src/widgets.js
emord: That was the original way I implemented it, but forgot to clean it up

src/mugs.js
emord: Done

PR #528, opened 2015-09-12
Form comments hacked together

emord:
@czue threw this together in a couple minutes. Is this something like you were talking about?

image

Would need some cleanup/tests. Could also probably throw it in itext just as easily if we wanted it to be able to have different languages (assuming somethign like `value=comment` doesn't mean anything to mobile)

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1661,7 +1661,7 @@ define([
             form = this.data.core.form,
             mugs = multiselect ? mug : [mug],
             $baseToolbar = $(question_toolbar({
-                comment: mug.p.comment,
+                comment: multiselect ? '' : mug.p.comment,
                 isDeleteable: mugs && mugs.length && _.every(mugs, function (mug) {
                     return _this.isMugRemoveable(mug, form.getAbsolutePath(mug));
                 }),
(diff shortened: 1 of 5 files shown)


czue:
it's close to what I was thinking. how are you thinking people edit them?

czue:
re: itext, one other downside of using something real that goes to phone is performance (size of forms, possible runtime of forms). not sure if that's a real concern or not.

czue:
(awesome how quick this was though!)

emord:
> it's close to what I was thinking. how are you thinking people edit them?

It's just another text box. (stuck it in advanced right now)

czue:
ah nice. yeah then this is exactly what i was thinking

emord:
Verified that HQ will strip out the comment and tests are passing

@dimagi/product any feedback on this feature?

sheelio:
Is this to allow users to add comments to questions?  Does that comment display support markdown?

emord:
Yes it's to add comments to questions. @czue was saying it would be a super useful feature. It doesn't support markdown, but it could (would be a bit more work to get a preview window though)

amsagoff:
Awesome! Pro releasing this as is (a textbox in advanced) and iterating
based on feedback from FMs, who are pretty much the target users.

czue:
:tada: :balloon: :confetti_ball:

czue:
@kaapstorm @snopoke fyi from summit talks

kaapstorm:
This is great!

snopoke:
suite!

PR #529, opened 2015-09-14
add tests

emord:
@czue last PR got merged as i was adding tests :)


czue:
ah nice. saw the build passed and figured it was g2g, but makes sense you'd want to have specific tests for the feature.

PR #530, opened 2015-09-15
Escape > in itext ids

emord:
http://manage.dimagi.com/default.asp?182112

@TylerSheffels 

@orangejenny @kaapstorm

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -41,8 +41,8 @@ define([
         return $(_.template($(selector).text(), params));
     };
     
-    that.validAttributeRegex = /^[^<&'"]*$/;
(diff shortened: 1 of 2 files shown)


orangejenny:
Looks fine once tests pass. Any idea if there would be a downside to doing something much stricter (e.g., only allowing word characters)? Having just done a Unicode bug, that's also something to consider.

orangejenny:
(realize xml supports unicode, but generally find being stricter upfront saves headaches later)

emord:
@orangejenny Agree that's better long term. Not sure why it was done this way initially, but I'll look into it later. Would rather not introduce it here

PR #534, opened 2015-09-23
Fix dropdown

emord:
http://manage.dimagi.com/default.asp?182308#1014978

introduced during a refactor https://github.com/dimagi/Vellum/pull/468

(no test coverage for at.js related stuff... need to think about how to fix that)

@orangejenny cc: @TylerSheffels

### DIFF
--- a/src/expressionEditor.js
+++ b/src/expressionEditor.js
@@ -107,6 +107,9 @@ define([
         };
         var addAutocomplete = function (input, choices) {
             if (choices) {
+                if (_.isFunction(choices)) {
+                    choices = choices();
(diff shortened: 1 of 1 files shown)


TylerSheffels:
:+1:

orangejenny:
fine once tests pass

> need to think about how to fix that

was this throwing an error or just not popping up the choices? can't tell from the ticket.

emord:
> was this throwing an error or just not popping up the choices? can't tell from the ticket.

No error. just not showing the drop down. You could still manually type in values

PR #535, opened 2015-09-23
Add lookup tables feature flag

emord:
http://manage.dimagi.com/default.asp?178265#1001988

Warns the user when we find an itemset on a community project

image

@dimagi/product how does this look?

@orangejenny 
cc: @TylerSheffels

### DIFF
--- a/src/itemset.js
+++ b/src/itemset.js
@@ -27,6 +27,7 @@ define([
 ) {
     var mugTypes = mugs.baseMugTypes.normal,
         Itemset, isAdvancedItemsetEnabled, opts,
+        changeSubscriptionLink,
         END_FILTER = /\[[^\[\]]*\]$/;
 
     Itemset = util.extend(mugs.defaultOptions, {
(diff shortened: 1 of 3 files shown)


amsagoff:
New copy: "You no longer have access to Lookup Tables in your application.
Lookup Tables are available on the Standard plan and higher. Before you can
make a new version of your application, you must [change your subscription]
or delete this question."

Where "change your subscription" links to the Current Subscription page of
their project space

orangejenny:
looks fine pending tests & copy

emord:
Link updated https://github.com/dimagi/Vellum/commit/79b0dc61bc43b61119ad96eb0df93e88ef1edc0b. wait for tests

src/itemset.js
```diff
@@ -96,6 +97,12 @@ define([
                     return value;
                 },
                 validationFunc: function (mug) {
+                    if (!mug.options.lookupTablesEnabled) {
+                        return "You no longer have access to Lookup Tables in your application. " +
+                            "Lookup Tables are available on the Standard plan and higher.\n" +
+                            "Before you can make a new version of your application, " +
+                            "you must [change your subscription](" + changeSubscriptionLink + ") or delete this question.";
```
orangejenny: I suspect the brackets were meant to signify "make this text a link", not to be literal. ( @amsagoff ?)

emord: Validation function values go through the markdown evaluator

orangejenny: This looks like it's going to dump the link text onto the page...can it be HTML? If not, it's probably better to have some hint text (something like "you must change your subscription (in Project Settings)") instead of the link.

orangejenny: And to be super paranoid, if there is a link, might want to check that the link exists before including it in the message. Bugs causing us to create "#" links that go nowhere will frustrate people.

orangejenny: > Validation function values go through the markdown evaluator

Magical.

amsagoff: yes, thanks jenny

emord: https://github.com/dimagi/Vellum/commit/b0be04a0ddadd158698bd49ffc68c6f39c362f4d

checked for the link [text](link) was intentional as it goes through markdown

PR #536, opened 2015-09-23
Support markdown in groups

emord:
https://github.com/dimagi/commcare-odk/pull/389

2.23 released

cc: @TylerSheffels


TylerSheffels:
Does this need a way to know the app version? e.g. if someone has selected 2.22 will they still be able to add markdown here and it will break on the phone?

biyeun:
@emord I think @TylerSheffels  is waiting on a Q

emord:
@TylerSheffels It wouldn't have "broken" on the phone. Just shown up as plain text

PR #538, opened 2015-09-25
Improve intent paywall

emord:
http://manage.dimagi.com/default.asp?181125#1017696

Trying to load a form with an intent to a community plan:
selection_050

Custom intents with a templated intent:
selection_051

Custom intent:
selection_052

@dimagi/product Just the vellum changes. doesn't actually block the build atm. Based the error text on the lookup table error message

@orangejenny @TylerSheffels


amsagoff:
Awesome!

Do we need to display the intent, even read-only, for templated
integrations?

Let's also rename "Intent ID" to "External App"

emord:
@amsagoff That's only for people with access to custom intents. If it's just templated intents then it looks like the first picture (without the error). I suppose we don't need to, but I left it in there as it was easier and it seemed like this was a seldom used feature anyways

TylerSheffels:
:+1: from me

src/intentManager.js
```diff
@@ -4,15 +4,13 @@ define([
...
-    'tpl!vellum/templates/intent_templates',
     'vellum/core'
 ], function (
     mugs,
     widgets,
     util,
     _,
-    $,
-    intent_templates
+    $
```
orangejenny: A brief intro - something explaining how intents can be templated or custom, and...I'm not sure what else should be in the intro, because I haven't finished reading the PR yet - would make this file quite a bit better.

src/intentManager.js
```diff
@@ -319,7 +324,10 @@ define([
 
     $.vellum.plugin("intents", {}, {
         init: function() {
-            refreshCurrentMug = this.refreshCurrentMug.bind(this);
```
orangejenny: Why isn't this needed anymore?

src/widgets.js
```diff
@@ -643,12 +645,31 @@ define([
...
+            }
+
+            val = parseValue(val);
             return _.find(widget.getOptions(), function (option) {
-                return _.isEqual(option.value ? JSON.parse(option.value) : '', val);
+                return _.isEqual(parseValue(option.value), val);
             });
         };
 
+        widget.findOrAddCustom = function (value) {
```
orangejenny: super picky, but `findOrAddCustom` sounds like it returns something, either a boolean or the custom option. `addCustomIfNeeded`?

tests/intentManager.js
```diff
@@ -17,17 +18,48 @@ define([
...
 ) {
     var assert = chai.assert,
-        call = util.call;
+        call = util.call,
+        templates =  [
+            {
+                icon: "icon-map-marker",
+                name: "Area Mapper",
+                id: "com.richard.lu.areamapper",
+                extra: {ext: "value"},
```
orangejenny: what are the `extra` and `response` properties for?

tests/intentManager.js
```diff
@@ -17,17 +18,48 @@ define([
...
+                    r2: "y",
+                    r3: "z",
+                    r4: "",
+                },
+            },
+            {
+                icon: "icon-barcode",
+                name: "Barcode Scanner",
+                id: "com.google.zxing.client.android.SCAN",
+                extra: {},
```
orangejenny: Why does this need an empty `extra` and `response`?

src/widgets.js
```diff
@@ -656,6 +677,59 @@ define([
...
+            return widget.text.val();
+        };
+
+        widget.getControl = function () {
+            return control;
+        };
+
+        widget.dropdown.change(function () {
+            var selectedOption = widget.dropdown.find(':selected');
+            widget.text.attr('readonly', true);
```
orangejenny: Shouldn't this be set to false if they selected `Custom`?

tests/intentManager.js
```diff
@@ -190,5 +196,82 @@ define([
...
+                        onReady: function () {
+                        vellum = this;
+                        mug = util.addQuestion("AndroidIntent", "intent");
+                        util.clickQuestion("intent");
+                        done();
+                    }}
+                });
+            });
+
+            it("should always have one custom option", function() {
```
orangejenny: I'd add a test that the input disabling happens correctly (readonly on selecting custom, writable on selecting anything else).

src/intentManager.js
emord: https://github.com/dimagi/Vellum/commit/21dcf6f937e5a34f8368fd0926ae9aa3270858c5

src/intentManager.js
emord: It actually hasn't been needed since I pulled out field auto completion for the printing question (custom feature and had some issues with mediauploader)

src/widgets.js
emord: https://github.com/dimagi/Vellum/commit/6b37fdea2129053145c0ee87bbaeff9ad3ff1fcf

tests/intentManager.js
emord: Those were moved from later in the file.

Those can actually auto fill in the extra/response fields based on template selection (can be done in vellum, but not present in HQ, tested here https://github.com/dimagi/Vellum/blob/6aed8e75985678776567657a06a9c64f6fa2e15d/tests/intentManager.js#L158-L159)

src/widgets.js
emord: https://github.com/dimagi/Vellum/commit/e1a200982347e42b6d603d805599710057638ff5

tests/intentManager.js
emord: https://github.com/dimagi/Vellum/commit/e1a200982347e42b6d603d805599710057638ff5

PR #539, opened 2015-09-26
standardize buttons

orangejenny:
there's no logic to how we use `.btn-inverse`

@nickpell / @emord

### DIFF
--- a/src/templates/multimedia_queue.html
+++ b/src/templates/multimedia_queue.html
@@ -12,9 +12,6 @@
         <p class="text-error hqm-error hide">
             <span class="label label-important">There were errors.</span>
         </p>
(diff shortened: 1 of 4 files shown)


src/templates/multimedia_queue.html
```diff
@@ -12,9 +12,6 @@
         <p class="text-error hqm-error hide">
             <span class="label label-important">There were errors.</span>
         </p>
-        <a href="#" class="btn btn-inverse btn-mini pull-right hqm-remove">
```
emord: Why did you delete this button?

Other than this :+1: to merge

orangejenny: Seemed clutter-y in a rather small dialog - you can already select a new image to overwrite the first one, or close the dialog altogether.

PR #540, opened 2015-09-26
Allow appearance to be copied on selects

emord:
http://manage.dimagi.com/default.asp?182841#1018574

cc: @TylerSheffels

### DIFF
--- a/tests/static/itemset/test1.xml
+++ b/tests/static/itemset/test1.xml
@@ -23,7 +23,7 @@
 		</model>
 	</h:head>
 	<h:body>
-		<select1 ref="/data/question1">
+		<select1 ref="/data/question1" appearance="minimal">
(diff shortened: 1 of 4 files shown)


tests/static/itemset/test1.xml
```diff
@@ -23,7 +23,7 @@
 		</model>
 	</h:head>
 	<h:body>
-		<select1 ref="/data/question1">
+		<select1 ref="/data/question1" appearance="minimal">
```
orangejenny: ...were you also planning on writing a test?

TylerSheffels:
:+1:, will wait for a test

tests/static/itemset/test1.xml
emord: This did test it, but I suppose it was a little too sneaky. I added a more explicit test.

PR #542, opened 2015-09-30
choice -> item as default

emord:
http://manage.dimagi.com/default.asp?89892#1012375

Makes item1, item2, ... into choice1, choice2, .... by default

@amsagoff 

cc: @TylerSheffels

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -299,9 +299,7 @@ define([
             {
                 group: ["Select", 'Multiple Choice'],
                 related: [
-                    "Item"
-                    // an Itemset is added automatically when you add a new dynamic
-                    // select
+                    "Choice"
                 ],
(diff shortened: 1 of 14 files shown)


amsagoff:
you mean the reverse :)

so satisfying to see this finally happen

TylerSheffels:
Does this have any affect on existing forms?

TylerSheffels:
If the answer to that question is no, which I assume it is, then G2M

emord:
It shouldn't have any negative effect as these are just defaults. The only change would be if someone had a choices that were item1 item2 then a new choice would be choice1, choice2, but people usually change those anyways

There's actually still references to item1, item2 in tests that I didn't change, so having non-choice1 names should still be tested.

PR #543, opened 2015-09-30
Drag n drop

emord:
http://dimagi.uservoice.com/forums/176376-form-builder/suggestions/4412064-ability-to-use-output-ref-output-value-in-the-hint

This has worked for all itexts for a while. Just adding drag and drop for it.

cc: @TylerSheffels


czue:
11 votes crushed!

PR #544, opened 2015-09-30
Media section showing for hidden questions

orangejenny:
@emord

### DIFF
--- a/src/javaRosa.js
+++ b/src/javaRosa.js
@@ -1993,7 +1993,9 @@ define([
             };
             databind.constraintMediaIText = function (mugOptions) {
                 return mugOptions.isSpecialGroup ? undefined : {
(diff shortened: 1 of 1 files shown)


src/javaRosa.js
```diff
@@ -1993,7 +1993,9 @@ define([
             };
             databind.constraintMediaIText = function (mugOptions) {
                 return mugOptions.isSpecialGroup ? undefined : {
-                    visibility: 'constraintMsgItext',
+                    visibility: function(mug) {
+                        return mug.__className !== "DataBindOnly" && mug.isVisible("constraintMsgItext");
```
emord: I think `mug.isVisible("constraintAttr")` will do the same thing and is more maintainable

orangejenny: fixed

PR #546, opened 2015-10-01
Repeat count required

emord:
http://manage.dimagi.com/default.asp?183425#1022284

@phillipm 

cc: @TylerSheffels

### DIFF
--- a/src/mugs.js
+++ b/src/mugs.js
@@ -1538,7 +1538,7 @@ define([
             repeat_count: {
                 lstring: 'Repeat Count',
                 visibility: 'visible_if_present',
-                presence: 'required',
+                presence: 'optional',
                 widget: widgets.droppableText
             },
             rawRepeatAttributes: {
--- a/src/mugs.js
+++ b/src/mugs.js
@@ -1539,7 +1539,18 @@ define([
                 lstring: 'Repeat Count',
                 visibility: 'visible_if_present',
                 presence: 'optional',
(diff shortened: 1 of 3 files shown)


phillipm:
Context: This is the real fix for http://manage.dimagi.com/default.asp?182818 contrary to the previous fix attempt, https://github.com/dimagi/Vellum/pull/537, which didn't discriminate based on whether the repeat was or was not in a field list.

This looks right, though I'll let someone else review & merge since I'm not up on my javascript.

TylerSheffels:
:+1: after tests

src/mugs.js
```diff
@@ -1538,8 +1538,19 @@ define([
...
-                presence: 'required',
-                widget: widgets.droppableText
+                presence: 'optional',
+                widget: widgets.droppableText,
+                validationFunc: function (mug) {
+                    var parentMug = mug.parentMug,
+                        value = $.trim(mug.p.repeat_count);
+
+                    if (!value &&
+                        parentMug && parentMug.__className === 'FieldList') {
```
orangejenny: isn't it potentially any ancestor, not only the direct parent?

phillipm: sounds right, good catch!

emord: https://github.com/dimagi/Vellum/commit/87fca042da5475ea6987d77a64d3915cadc32ebe

PR #549, opened 2015-10-06
Atwho improvement

emord:
tooltip_058

@dimagi/product I think I recall someone mentioning that seeing the label in the dropdown would be nice. The gray text under the question id is the default lanugage's label. This also allows searching by that label

This PR also introduces some fuzzy searching so that you don't need to type exactly the right text. Notice qustion4 matches question4 in the screenshot. The values are ranked by closeness of text.

Any of this can be pulled into separate PRs.

@millerdev I started to test atwho stuff earlier, but when running the tests, the atwho containers did not clean up quickly enough. (can sync on this separately)

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -32,6 +32,7 @@
         "requirejs-plugins": "~1.0.2",
         "At.js": "emord/At.js#override-insert",
         "Caret.js": "0.2.2",
+        "fuse.js": "1.2.2",
 
         "jstree": "3.1.1",
         "langcodes": "dimagi/langcodes",
--- a/src/atwho.js
+++ b/src/atwho.js
@@ -135,9 +135,12 @@ define([
             };
         }
 
+        var mugData = cachedMugData(mug.form),
+            fuse = new fusejs(mugData, { keys: ['label', 'name'] });
+
         $input.atwho({
             at: "/data/",
-            data: cachedMugData(mug.form),
+            data: mugData,
             displayTpl: '<li><i class="${icon}" /> ${name}<br /><span class="atwho-jrtext">${displayLabel}</span></li>',
             insertTpl: options.insertTpl,
             limit: 10,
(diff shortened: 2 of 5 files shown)


src/atwho.js
```diff
@@ -29,11 +33,23 @@ define([
     var cachedMugData = timed(function(form) {
         return _.chain(form.getMugList())
                 .map(function(mug) {
+                    // probably better to use text-overflow: ellipsis
```
emord: Didn't quite get the text-overflow css to work properly. (never actually cut off the data)

src/templates/atwho_display.html
```diff
@@ -0,0 +1,7 @@
+<li>
+  <i class="<%= icon %>" /> <%= name %>
+    <% if (displayLabel) { %>
```
emord: Hidden values won't have displayLabel

bower.json
```diff
@@ -32,6 +32,7 @@
         "requirejs-plugins": "~1.0.2",
         "At.js": "emord/At.js#override-insert",
         "Caret.js": "0.2.2",
+        "fuse.js": "1.2.2",
```
millerdev: Would it make sense to pin this at `"~1.2.2"`? I believe that tells bower that the original version we used was 1.2.2, but upgrade to newer versions of 1.2 if they become available. This is great if the library uses semantic versions; not so great if not.

src/atwho.js
```diff
@@ -29,11 +33,23 @@ define([
...
+                    var defaultLanguage = mug.form.vellum.data.javaRosa.Itext.getDefaultLanguage(),
+                        itext = mug.p.labelItext,
+                        defaultLabel = itext ? itext.getForm('default').getValue(defaultLanguage) : '',
```
millerdev: I think this could be shortened to
```js
var itext = mug.p.labelItext,
    defaultLabel = itext ? itext.get() : '',
```

src/templates/atwho_display.html
amsagoff: love this!

src/atwho.js
emord: Nice. changed it.

bower.json
emord: I'm going to keep it at 1.2.2 since it's a new library from someone I'm not familiar with. If we notice that there are new features when a new version is released, it'll probably be good to decide if we can trust it.

czue:
sexy

PR #553, opened 2015-10-23
Display autocomplete options in form's current language

orangejenny:
http://manage.dimagi.com/default.asp?185094

@millerdev 

cc @NoahCarnahan

### DIFF
--- a/src/javaRosa.js
+++ b/src/javaRosa.js
@@ -1478,8 +1478,10 @@ define([
                 }
             });
 
-            if (form) {
-                form.vellum.refreshCurrentMug();
+            if (form && form.vellum.getCurrentlySelectedMug()) {
+                form.vellum.getCurrentlySelectedMug().fire({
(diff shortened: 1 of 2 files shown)


src/javaRosa.js
```diff
@@ -1477,6 +1477,10 @@ define([
                     }
                 }
             });
+
+            if (form) {
+                form.vellum.refreshCurrentMug();
```
orangejenny: Too heavy-handed?

orangejenny: this is causing all the test failures

millerdev: Oof... my first impression is yes. There should be a better way, but I'm not sure what problem this is solving so it's hard to suggest a better way without digging in deeper.

orangejenny:
@millerdev different approach: fire & listen for an event on language change

millerdev:
Excellent! I like this approach much better.

PR #554, opened 2015-10-26
Fix newlines being eaten by rich text widget

millerdev:
http://manage.dimagi.com/default.asp?184602

Best reviewed by commit.

@orangejenny 
cc @emord (you should review this when you get back from vacation)

### DIFF
--- a/src/expressionEditor.js
+++ b/src/expressionEditor.js
@@ -81,9 +81,11 @@ define([
         var setExpression = function(input, val) {
                 if (options.mug.supportsRichText()) {
                     editor = input.ckeditor().editor;
-                    editor.setData(
-                        richText.toRichText(val, options.mug.form, true)
-                    );
+                    editor.setData(richText.toRichText(
+                        val,
+                        options.mug.form,
+                        {withClose: true, isExpression: true}
+                    ));
                 } else {
                     input.val(val);
                 }
--- a/src/expressionEditor.js
+++ b/src/expressionEditor.js
@@ -79,18 +79,20 @@ define([
         };
 
         var setExpression = function(input, val) {
-            if (options.mug.supportsRichText()) {
-                input.ckeditor().editor.setData(
-                    richText.toRichText(val, options.mug.form, true)
-                );
-            } else {
-                input.val(val);
-            }
-        };
+                if (options.mug.supportsRichText()) {
+                    editor = input.ckeditor().editor;
+                    editor.setData(
+                        richText.toRichText(val, options.mug.form, true)
+                    );
+                } else {
+                    input.val(val);
+                }
+            },
+            editor;
 
         var getExpression = function(input) {
             if (options.mug.supportsRichText()) {
-                return richText.fromRichText(input.ckeditor().editor.getData());
+                return richText.fromRichText(editor.getData());
             } else {
                 return input.val();
             }
--- a/src/expressionEditor.js
+++ b/src/expressionEditor.js
@@ -68,7 +68,9 @@ define([
     addOp(FunctionHandler, "selected", "has selected value");
 
     function showXPathEditor($div, options) {
-        var editorContent = $div;
+        var editorContent = $div,
+            richTextOptions = {withClose: true, isExpression: true},
+            form = options.mug.form;
         options = _.defaults(options, {
             leftPlaceholder: "Hint: drag a question here.",
             rightPlaceholder: "Hint: drag a question here.",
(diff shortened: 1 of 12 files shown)


millerdev:
cc @kaapstorm

kaapstorm:
:+1: Awesome

src/expressionEditor.js
```diff
@@ -79,18 +79,22 @@ define([
...
+                } else {
+                    input.val(val);
+                }
+            },
+            editor;
 
         var getExpression = function(input) {
             if (options.mug.supportsRichText()) {
-                return richText.fromRichText(input.ckeditor().editor.getData());
+                return richText.fromRichText(editor.getData());
```
orangejenny: This line throws a js error when you go to the advanced expression editor.

orangejenny: (editor is undefined)

millerdev: Thanks for finding this. I'll look into it.

src/richText.js
```diff
@@ -213,55 +216,130 @@ define([
...
-     * Takes a value from xml and replaces xpaths with bubbles to display in the
-     * editor
+     * Replace <output> tags with bubble markup
      *
-     * This does not add any <p> or <br> tags, only newlines as used in xml
+     * @param withClose - Create bubbles with close buttons if true.
+     * @param escape - If true, escape HTML except for bubble markup.
+     */
+    function bubbleOutputs(text, form, withClose, escape) {
+        var el = $('<div>').html(text),
```
emord: text should be escaped like https://github.com/dimagi/Vellum/pull/554/files#diff-85ed993f6f09d78b2af63dbea82386a3R333

selection_001

millerdev:
@emord @orangejenny updated.

Majorly refactored so CKEditor is now encapsulated in an editor object that provides a more convenient interface for vellum. This would theoretically allow us to more easily swap out CKEditor with another editor in the future if that becomes necessary.

I think this should now be easy enough to review with the PR diff rather than by commit, although the commit messages may have some useful context.

src/widgets.js
```diff
@@ -248,112 +240,37 @@ define([
...
+        mug.on('teardown-mug-properties', function() {
+            editor.destroy();
+        }, null, "teardown-mug-properties");
 
-            editor.on('dataReady', function (e) {
-                addCloseButton(widget, widget.input);
-                addPopovers(widget.input);
-            });
+        editor.on('change', function() {
+            widget.handleChange();
```
emord: I think these could be:
```
mug.on('teardown-mug-properties', editor.destroy, null, 'teardown-mug-properties');
editor.on('change', widget.handleChange);
```

tests/utils.js
```diff
@@ -344,8 +352,42 @@ define([
...
+     * Usage:
+     *
+     *  it("will always fail", function (done) {
+     *      someAsyncFunction(asyncatch(function (arg) {
+     *          doSomethingThatThrowsAnError();
+     *          done();
+     *      }));
+     *  });
+     */
+    function asyncatch(fn) {
```
emord: `asyncCatch` ? combining the c's looks a little strange to me

emord:
Just non blocking comments left. Looks good to merge pending your decision on `CKEDITOR_BASEPATH`

src/richText.js
```diff
@@ -22,51 +22,203 @@
  *
  * Any other HTML has undefined behavior
  */
+
+(function () {
+    // set CKEditor base path before loading ckeditor
+    var require = window.requirejs,
+        path = require.toUrl("vellum/../lib/ckeditor/").replace(/\?.*/, "");
+    window.CKEDITOR_BASEPATH = path;
```
millerdev: Tested locally in standalone mode (built and not) and in local HQ with `VELLUM_DEBUG` set to each of `dev`, `dev-min`, and not set.

tests/utils.js
millerdev: I vacillated on that one. Going to leave as is.

src/mugs.js
```diff
@@ -356,7 +356,8 @@ define([
                 .replace("$2", property)
                 .replace("$3", String(vis)));
         },
-        getDisplayName: function (lang) {
+        getDisplayName: function (lang, escape) {
+            if (escape === undefined) { escape = true; }
```
emord: I don't think you overwrite `undefined` anymore. Do you think we should stop using `_.isUndefined` in future code?

millerdev: yes. I read up on it and `x === undefined` should be safe. I've stopped using `_.isUndefined` because I think it's ugly.

PR #555, opened 2015-10-30
Improve markdown detection

millerdev:
http://manage.dimagi.com/default.asp?185830

@orangejenny cc @kaapstorm

### DIFF
--- a/src/javaRosa.js
+++ b/src/javaRosa.js
@@ -1052,6 +1052,18 @@ define([
         return widget;
     };
 
+    function looksLikeMarkdown(val) {
+        /* Regex checks (in order):
+         * ordered lists
+         * unordered lists
+         * strikethrough
+         * headings
+         * italics/bold/bold italics
+         * links
+         */
+        return /^\d+\. |^\* |~~.+~~|# |\*{1,3}\S.*\*{1,3}|\[.+\]\(\S+\)/m.test(val);
+    }
+
     var itextMarkdownWidget = function (mug, language, form, options) {
         options = options || {};
         var parent = options.parent;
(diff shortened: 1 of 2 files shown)


src/javaRosa.js
```diff
@@ -1052,6 +1052,18 @@ define([
...
+    function looksLikeMarkdown(val) {
+        /* Regex checks (in order):
+         * ordered lists
+         * unordered lists
+         * strikethrough
+         * headings
+         * italics/bold/bold italics
+         * links
+         */
+        return /^\d+\. |^\* |~~.+~~|# |\*{1,3}\S.*\*{1,3}|\[.+\]\(\S+\)/m.test(val);
```
millerdev:
```diff
+ return /^\d+\. |^\* |~~.+~~|# |\*{1,3}\S.*\*{1,3}|\[.+\]\(\S+\)/m.test(val);
- return /^\d+\. |^\*|~~.+~~|# |\*{1,3}\S+\*{1,3}|\[.+\]\(\S+\)/m.test(val);
```

kaapstorm:
:+1:

orangejenny:
This regex also lives in HQ: https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/app_manager/translations.py#L526

I've considered making it a variable that gets passed to vellum and decided against it, because it would still need to live in two places (once in HQ, once in tests/options.js)...but on the other hand, being updating it in options.js might make it more obvious that it also needs to be updated in HQ.

millerdev:
@orangejenny: transferring a regexp between Python and JS could lead to problems if certain features were used. For example back references, which I thought of using for matching the end delimiter for *italics*/**bold**/***bold italics***, have different syntax in [JavaScript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp#grouping-back-references) and [Python](https://docs.python.org/3/library/re.html#regular-expression-syntax) (they're called "group references" in Python).

orangejenny:
Yeah, fair point. The current regex is the same in python & js, but it may not always be.

PR #559, opened 2015-11-04
Preserve contiguous spaces in rich text widget

millerdev:
Prior to this PR multiple contiguous spaces were collapsed into a single space by CKEditor. Additionally, adding multiple contiguous spaces with CKEditor resulted in non-breaking spaces to be encoded in the XForm, which broke expression validation at build/deploy time.

Fixes http://manage.dimagi.com/default.asp?185331

@emord cc @sravfeyn

### DIFF
--- a/src/richText.js
+++ b/src/richText.js
@@ -434,10 +434,11 @@ define([
     /**
      * Convert plain text to HTML to be edited in CKEditor
      *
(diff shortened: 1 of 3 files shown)


sravfeyn:
Looks good to me, will let emord merge it

PR #560, opened 2015-11-05
Work around CKEditor selection bug

millerdev:
http://manage.dimagi.com/default.asp?185791

This actually fixes two bugs. Unfortunately I could not come up with a way to reproduce the CKEditor selection bug in an automated test, so I had to test it manually.

@emord cc @sravfeyn

### DIFF
--- a/tests/richText.js
+++ b/tests/richText.js
@@ -314,22 +314,6 @@ define([
             assert.equal(editor.getValue(), text);
         });
 
-        it("should insert expression into expression editor", function (done) {
-            editor.setValue('one two', function () {
-                assert.equal(editor.getValue(), 'one two');
-                editor.select(3);
-                // temporarily change to expression editor
-                options.isExpression = true;
-                try {
-                    editor.insertExpression('/data/text');
-                    assert.equal(editor.getValue(), "one/data/text two");
-                } finally {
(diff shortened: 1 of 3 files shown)


sravfeyn:
Looks good, will let emord review.

tests/richText.js
```diff
@@ -345,73 +329,50 @@ define([
...
-                        if (child !== null) {
-                            return child;
-                        }
-                        inner = null;
+        _.each([
+            ["one two", 3, "one/data/text two"],
+            ["one two", 4, "one /data/text two"],
+            ["one\n\ntwo", 3, "one/data/text\n\ntwo"],
+            ["one\n\ntwo", 4, "one\n/data/text\ntwo"],
+            /* TODO make these tests pass
```
emord: Just did some quick testing and I could insert past the second newline there. These scenarios do work in the UI correct?

millerdev: Yes, they work in the UI. It's just that the programmatic selection logic gets confused beyond the first (or second?) line, probably because I'm counting nodes wrong in my selection logic. But that shouldn't affect any use of `ckSelect` right now since the index parameter is always zero in the only place it's called (outside of tests).

PR #561, opened 2015-11-06
Upgrade to mocha-phantomjs 4.0.1

millerdev:
The `--bail` option did not work in v3.5.3. See https://github.com/nathanboktae/mocha-phantomjs/issues/215

Don't forget to run `npm update mocha-phantomjs` after merging this. And you'll need to run that command again when switching to branches made before this is merged.

@emord @orangejenny 
cc @sravfeyn

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
         "jshint": ">=1.1.0",
         "mocha": ">=1.9.0",
         "mocha-istanbul": ">=0.2.0",
-        "mocha-phantomjs": "3.5.3",
+        "mocha-phantomjs": "~4.0.1",
         "http-server": "*",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -21,7 +21,7 @@
         "jshint": ">=1.1.0",
         "mocha": ">=1.9.0",
         "mocha-istanbul": ">=0.2.0",
```
emord: I think we can get rid of `istanbul` and `mocha-istanbul`, or do we have code coverage that I'm not aware of?

millerdev: Agreed. I was just thinking about that last night when I was working on this PR.

PR #562, opened 2015-11-11
Provide help message instead of warning

emord:
http://manage.dimagi.com/default.asp?184259

Don't really like this solution, but it'll take a little longer to make messages have different levels warnings/error/info. It's next up on my PID list.

cc: @gcapalbo

### DIFF
--- a/src/mugs.js
+++ b/src/mugs.js
@@ -903,17 +903,8 @@ define([
                 xpathType: 'generic',
                 serialize: serializeXPath,
                 deserialize: deserializeXPath,
-                validationFunc: function (mug) {
-                    var paths = new logic.LogicExpression(mug.p.defaultValue).getPaths();
-                    paths = _.filter(paths, function (path) {
-                        return path.initial_context !== xpathmodels.XPathInitialContextEnum.EXPR;
-                    });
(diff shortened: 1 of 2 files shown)


src/mugs.js
```diff
@@ -903,17 +903,8 @@ define([
...
-                        return path.initial_context !== xpathmodels.XPathInitialContextEnum.EXPR;
-                    });
-                    if (paths.length) {
-                        return "You are referencing a node in this form. " +
-                               "This can cause errors in the form";
-                    }
-                    return 'pass';
-                }
+                help: "Referencing a node in this form may cause errors and/or " +
+                    "unexpected behavior",
```
gcapalbo: i think in js we try to avoid the trailing comma on dicts right?

emord: Vellum has them all over. trailing commas should only cause problems in non ES5 compliant browsers. and those browsers will have many issues trying to load hq

millerdev:
I agree with @ctsims: "it's hard to stress how likely people are to make user errors if they reference /data/ in <setvalue> events"

This change may result in more bugs being reported because people are referencing `/data/...` in the default value without realizing how it works.

PR #563, opened 2015-11-12
Don't delete instances on dereference

emord:
http://manage.dimagi.com/default.asp?186582#1044106

Original issue: https://github.com/dimagi/Vellum/blob/624473daf7b71ab85058ad0603d346826be00361/src/logic.js#L248-L249

Before it would delete the reference if there was only reference to the instance and then try to re-add it, but this isn't possible with manually written instances to support growth charts and it deleted all the data.

This expands `instanceMetadata.internal` definition to mean "not referenced in the form"

@millerdev 

cc: @gcapalbo

### DIFF
--- a/src/form.js
+++ b/src/form.js
@@ -154,7 +154,6 @@ define([
                     refs[mug.ufid] = {};
                 }
                 refs[mug.ufid][property || "."] = null;
-                that.internal = false;
             }
         };
 
@@ -309,11 +308,6 @@ define([
                         meta = null;
                     }
                     this.knownInstances[attrs.id] = attrs.src;
-                } else if (meta.attributes.id !== attrs.id) {
-                    // support renaming when there is no reference
-                    if (meta.internal) {
-                        meta.attributes.id = attrs.id;
-                    }
                 }
             } else if (attrs.id) {
                 // attrs has no src, find by id
@@ -471,7 +465,7 @@ define([
                 });
             if (meta) {
                 if (meta.dropReference(mug, property)) {
-                    meta.internal = true;
+                    this.instanceMetadata = _.without(this.instanceMetadata, meta);
                     return true;
                 }
             }
@@ -481,10 +475,8 @@ define([
          * Drop all instance references from the given mug/property
          */
         dropAllInstanceReferences: function (mug, property) {
-            _.each(this.instanceMetadata, function (meta) {
-                if (meta.dropReference(mug, property)) {
-                    meta.internal = true;
-                }
+            this.instanceMetadata = _.filter(this.instanceMetadata, function (meta) {
+                return !meta.dropReference(mug, property);
             });
         },
         // todo: update references on rename
--- a/src/form.js
+++ b/src/form.js
@@ -211,7 +211,8 @@ define([
             _this.fireChange(e.mug);
         });
         this.instanceMetadata = [InstanceMetadata({})];
-        this.knownInstances = {}; // {<instance id>: <instance src or children>}
+        // {<instance id>: { src or children: <instance src or children>}
+        this.knownInstances = {};
         this.enableInstanceRefCounting = opts.enableInstanceRefCounting;
         this.errors = [];
         this.question_counter = 1;
@@ -307,7 +308,7 @@ define([
                         attrs.id = getUniqueId(attrs.id, ids);
                         meta = null;
                     }
-                    this.knownInstances[attrs.id] = attrs.src;
+                    this.knownInstances[attrs.id] = { src: attrs.src };
                 }
             } else if (attrs.id) {
                 // attrs has no src, find by id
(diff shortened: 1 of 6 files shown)


src/form.js
```diff
@@ -308,6 +309,11 @@ define([
                         meta = null;
                     }
                     this.knownInstances[attrs.id] = attrs.src;
+                } else if (meta.attributes.id !== attrs.id) {
+                    // support renaming when there is no reference
```
emord: Works because the reference is dropped then renamed then readded to the mug. Used in model iteration

src/form.js
```diff
@@ -154,6 +154,7 @@ define([
                     refs[mug.ufid] = {};
                 }
                 refs[mug.ufid][property || "."] = null;
+                that.internal = false;
```
millerdev: The meaning of this flag is sufficiently opaque that I think it would make sense to document how it affects instance metadata. It might make sense to put that in a docstring for the `InstanceMetadata` class.

src/form.js
```diff
@@ -465,7 +471,7 @@ define([
                 });
             if (meta) {
                 if (meta.dropReference(mug, property)) {
-                    this.instanceMetadata = _.without(this.instanceMetadata, meta);
+                    meta.internal = true;
```
millerdev: I don't understand this change. Also, the docstring is now outdated and should be updated if this is kept. It seems that this should not have caused an instance without a `src` to be dropped since it finds the instance by `src`. Can you explain?

millerdev:
> This expands `instanceMetadata.internal` definition to mean "not referenced in the form"

Previously I believe it meant "reference with data source (unknown `src`) and therefore will not be written in XML". It seems that conflating those two meanings might lead to some tricky edge cases.

An instance with `internal === true` should be one that is referenced by some expression, but not otherwise known by Vellum. This might happen, for example, of someone types an expression like `instance('foo')/...` where `foo` does not reference a known data source. The `internal` flag denotes that Vellum should not create an instance node `foo`.

src/form.js
millerdev: Not sure if this is referring to instances added with "Instance ID" and "Instance URI" fields in model iteration? FWIW, those fields should be removed since we now get a list of known data sources from HQ and no longer need to have the user enter those details manually. Any sufficiently advanced user who wants to add an instance that is not available in the list of data sources provided by HQ can manually edit XML to do so.

src/form.js
```diff
@@ -308,6 +309,11 @@ define([
...
                     }
                     this.knownInstances[attrs.id] = attrs.src;
+                } else if (meta.attributes.id !== attrs.id) {
+                    // support renaming when there is no reference
+                    if (meta.internal) {
+                        meta.attributes.id = attrs.id;
+                    }
                 }
             } else if (attrs.id) {
                 // attrs has no src, find by id
```
millerdev: Seems like another possible fix for the bug would be to amend this branch to recognize and mark as not internal any instance that contains child nodes (but no src attribute). We may also need to somehow track those instances in the `knownInstances` mapping.

millerdev: Or maybe not. This method is probably not the place where data instances (ones without a src attribute) are incorrectly being marked as internal.

millerdev: > We may also need to somehow track those instances in the knownInstances mapping.

The fact that we are not tracking data instances (without `src`) in `form.knownInstances` is definitely the problem. I think instead of having

`knownInstances = {<instance id>: <instance src>}`

we should have

`knownInstances = {<instance id>: <instance metadata>}`

emord:
now tracks in knownInstances

src/form.js
```diff
@@ -320,7 +320,11 @@ define([
                         meta.attributes.src = this.knownInstances[attrs.id];
                     }
                 } else if (this.knownInstances.hasOwnProperty(attrs.id)) {
-                    attrs.src = this.knownInstances[attrs.id];
+                    if (_.isObject(this.knownInstances[attrs.id]))
```
millerdev: This could accidentally catch a string created with `new` (not that we ever do that, but...). It would be a little safer to do it this way:
```js
if (_.isString(this.knownInstances[attrs.id])) {
    attrs.src = this.knownInstances[attrs.id];
} else {
    attrs.children = this.knownInstances[attrs.id];
}
```
Although even that has a smell. What about making the values in `knownInstances` be an object like `{src: <src uri>}` or `{children: <children obj>}`? Then you could iterate over properties of the value and assign all found attributes to `attrs`. The bonus of this approach is that it does not require a type check and we could easily add more attributes in the future if needed.

Also, the if body should be delimited with `{}`.

src/form.js
```diff
@@ -211,7 +211,7 @@ define([
             _this.fireChange(e.mug);
         });
         this.instanceMetadata = [InstanceMetadata({})];
-        this.knownInstances = {}; // {<instance id>: <instance src>}
+        this.knownInstances = {}; // {<instance id>: <instance src or children>}
```
millerdev: Ugh, I just realized this will probably break copy/paste for xpath expressions that reference data instances. See [xpath serialization in mugs.js](https://github.com/dimagi/Vellum/blob/manual-instance/src/mugs.js#L705-L718) as well as [`parseInstanceRefs` in form.js](https://github.com/dimagi/Vellum/blob/manual-instance/src/form.js#L381). Might want to add a test for that.

emord:
Changed the knownInstances definition and it properly copy and pastes with children

millerdev:
Wow, what happened with the tests? Otherwise, looks good.

emord:
Looks like it was due to a jshint upgrade not working properly. Pinning it now until I can look at it harder as it wasn't obvious

PR #566, opened 2015-11-19
Grunt

emord:
Adding in a grunt file to watch changes and do automated testing and linting. would be nice to get builds in here instead of a make file as well, but that's a little larger

@millerdev 

cc: @benrudolph

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -16,6 +16,14 @@ module.exports = function(grunt)  {
             }
         }
     },
+    watch: {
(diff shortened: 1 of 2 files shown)


benrudolph:
niiice

orangejenny:
the future is here.

orangejenny:
would be nice to mention this in vellum's main README

millerdev:
excellent! agree with @orangejenny on the README comment

PR #567, opened 2015-11-19
No close button adjustment for drag container

emord:
This happens before the close button is added. Right now there's dead space between drag container and close button

@millerdev 

cc: @benrudolph

### DIFF
--- a/src/richText.js
+++ b/src/richText.js
@@ -64,17 +64,6 @@ define([
             element.replaceWithChildren();
         },
         init: function() {
-            // TODO: PR to ckeditor to make changing drag ui supported
-            // Leave 15px on the left side so that users can actually
-            // interact with the close button
-            var width = $(this.element.$).innerWidth() - 15;
-            var height = $(this.element.$).outerHeight() + 4;
-            this.dragHandlerContainer.setStyles({
-                width: width + 'px',
-                height: height + 'px',
-                left: '0px'
-            });
-
             // Add close button
             var that = this,
                 buttons = this.element.getElementsByTag("button"),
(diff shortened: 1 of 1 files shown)


millerdev:
Cannot click close button after dragging bubble.

- Create a Text question
- Drag a case property into Label (English)
- Click on blob in Label (English)
- Drag to a different place in the label
- Try to delete bubble with its close button

Long term I'd like to consider removing the close button from bubbles since I think it doesn't add that much value; bubbles can easily be removed with normal text editing (backspace/delete/cut). I think they were originally added because gmail has them on email address bubbles in the To field. But notice that gmail does not allow you to navigate (with the text cursor) anywhere except after the last address entered in the To field. The gmail team had to add that button to make it possible to delete an address that is not the last one in field. We don't have that problem, and therefore we don't need the close button. Also, I'd like to add some other icons there. For example, if the bubble has date formatting applied to it, has a predicate filter, etc.

FYI @amsagoff

amsagoff:
FWIW I've already started using the close button, it's handy when your hand
is already on the mouse. Though agree it may not be necessary and the space
could potentially be put to better use.

Unless you feel strongly about it, let's make sure to add some analytics to
the close button before release!

emord:
Fixed by placing the drag container after the close button initialization

millerdev:
@amsagoff my only hesitation in doing the initial roll-out of bubbles with close buttons is that it may impede our ability to remove it later. I think if people first use the bubbles without the close button it will be obvious and intuitive for them to delete bubbles using normal text editing actions. However, if bubbles have close buttons from the beginning some users may think that is the only way to remove them and get confused if we ever decide to remove it. Less is more!

Some questions I'm asking myself:
- What is the potential down-side to removing the close button? Answer: user can't figure out how to remove a bubble. Seems unlikely to me.
- Under what circumstances would we decide to add close buttons if we didn't have them?  Answer: user finds it difficult to remove a bubble using other methods. Seems unlikely to me.
- Why have a close button? Answer: easy to remove a bubble with the mouse.
- Why not have close buttons? Answers:
  - less visual noise
  - implementation is simpler and less hacky
  - the close button creates a special zone on the bubble that cannot be used to click or drag the bubble (e.g., user clicks bubble to select it, accidentally deletes it because the mouse is a little too far to the right)
- How can we measure if close buttons are good or bad?

amsagoff:
...that is a very compelling argument you just laid out. @biyeun any strong feelings?

biyeun:
i'm with @millerdev on this actually. That close button is a real pain in the ass for making sure that ui works as intended across all browsers. As a text element, it seems intuitive to just delete it. This reduces complexity in both the UI and in the code, so +1 all around.

amsagoff:
The team has spoken. Adios, close button.

PR #568, opened 2015-11-20
Don't show data value when there isn't one defined

emord:
https://groups.google.com/forum/#!topic/commcare-users/uPdIy-9SQUk

cc: @benrudolph

### DIFF
--- a/src/parser.js
+++ b/src/parser.js
@@ -189,7 +189,7 @@ define([
 
         var mug = form.mugTypes.make(role, form);
         mug.p.nodeID = nodeID;
-        mug.p.dataValue = nodeVal;
+        mug.p.dataValue = nodeVal || undefined;
 
         if (extraXMLNS && (extraXMLNS !== form.formUuid)) {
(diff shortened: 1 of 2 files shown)


benrudolph:
haha noticed this one last night when reading that email

benrudolph:
i couldn't figure out the difference though. one allows you to reference /data/ and the other doesn't? the one that loads via /data/ seems obsolete and should just be done via case mgmt

emord:
Default Data Value writes a string in the data tree to that node. You can only use it for strings and xpaths. It's legacy

Default Value uses `setvalue` so you can use xpath expresions for the same thing, but also have the power to reference cases/fixtures/whatever inside vellum. The help message actually tells you that using /data/ could be an issue and isn't supported. (used to be a warning/error, but vellum doesn't currently have a distinction between the two and it was getting in the way of some other forms.)

benrudolph:
cool thanks for the explanation

benrudolph:
good after tests

PR #569, opened 2015-11-20
Show responses  for print question

emord:
No longer hide the extra response from print templates

cc: @amstone326 @benrudolph


amstone326:
woo thank you

millerdev:
I can't visualize in my head what this does. Does it add another field in the intent properties?

emord:
It just shows the "extra response" field from intents. I had previously hidden it because it wasn't going to be used in print questions

PR #572, opened 2015-11-24
Casemgmt qa

emord:
Fixes:
1. http://manage.dimagi.com/default.asp?187725
2. http://manage.dimagi.com/default.asp?187726
3. http://manage.dimagi.com/default.asp?187727

Also removes implicit support of rich text for key value widgets. Long term those should probably support rich text, but the key value widget (only used in intents) needs a refactor and save to case is not released to everyone.

added config helper plugin to ckeditor to support placeholders

@millerdev 

cc: @benrudolph


lib/ckeditor/dimagi-README.md
```diff
@@ -9,6 +9,7 @@ Use the following plugins:
 * floatingspace
 * undo (for change events)
 * widget
+* [config helper](http://ckeditor.com/addon/confighelper) (for placeholder support)
```
millerdev: What does config helper do that we need?

emord: It supports placeholders. like "Hint: drag a question here". So that when we define an input like `droppableText` whenever we initialize it there is still a placeholder

millerdev: :+1:

PR #574, opened 2015-11-25
Move data source widgets

emord:
Just moving code around to get past a circular import in my dev branch.

cc: @benrudolph


src/datasourcewidgets.js
```diff
@@ -0,0 +1,296 @@
...
+                options.onOptionsLoaded(data);
+            }
+        });
+
+        if (options.hasAdvancedEditor) {
+            widget.getUIElement = function () {
+                var query = widgets.util.getUIElementWithEditButton(
+                        widgets.util.getUIElement(widget.input, labelText),
+                        function () {
+                            mug.form.vellum.displaySecondaryEditor({
```
emord: Had to change this from `vellum` to `mug.form.vellum` I don't think that there should ever an instance where `mug.form.vellum` doesn't exist (and tests seem to agree with me)

millerdev: :+1:

PR #577, opened 2015-12-01
Allow self reference in calculate

emord:
http://manage.dimagi.com/default.asp?187877

code buddy: @TylerSheffels

### DIFF
--- a/src/logic.js
+++ b/src/logic.js
@@ -17,7 +17,7 @@ define([
         "repeat_count",
         "filter",
         "defaultValue"
-    ], NO_SELF_REFERENCES = _.without(XPATH_REFERENCES, 'constraintAttr');
+    ], NO_SELF_REFERENCES = _.without(XPATH_REFERENCES, 'constraintAttr', 'calculateAttr');
 
     function LogicExpression (exprText) {
         this._text = exprText || "";
--- a/tests/logic.js
+++ b/tests/logic.js
@@ -55,7 +55,7 @@ require([
                     "filter",
                     "defaultValue"
                 ],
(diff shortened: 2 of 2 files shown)


millerdev:
Would it be valid to have a calculate of `. + 1`? What would that mean? When does it get recalculated? My understanding is that bind calculates use some kind of continuous calculation mode. This either means that they are calculated when any of their dependencies change or that they are calculated just in time when they are needed (not sure which). If that is true, then it seems like the example above would either create an infinite calculation loop or erratically increment the calculated value each time it is referenced (undefined behavior?). If that is the case, then I imagine the warning was there to protect innocent users from dragging a question into its own calculate.

@ctsims may know the answer, or maybe he can correct my understanding of how calculates work.

It seems like we have had a few requests recently where it would be handy to show some warning message when a warning condition is created, but allow the user to dismiss it and not be bothered by it again. Not sure how hard this would be to implement (definitely harder than this PR), but this seems like a perfect place for it.

TylerSheffels:
@emord I don't think I have appropriate context for this, so I will defer to @millerdev

emord:
I think we're waiting on an answer from either @ctsims or @wpride to answer @millerdev question

ctsims:
Calculates can't be self-referential. "X = X + 1" doesn't have a stable
value.

-Clayton

wpride:
@ctsims @emord If the '.' is in the predicate (within '[' and ']') then it doesn't refer to current() and so is valid

emord:
Closing this for now as it's a bit more work to only warn when not in filters.

PR #578, opened 2015-12-02
adding csrf token to save button

calellowitz:
@orangejenny @millerdev @emord cc: @gcapalbo 
This modifies the save button to add CSRF header to its post request if one exists. As soon as CSRF protection on hq goes live, it wont be possible to save forms in the formbuilder without this. The token will be passed in through the options object.
If there is not a csrf token passed to the save button, no behavior will change.
Merging is not urgent because it will be a bit before the CSRF protection is up but I do not see a downside to merging early.

### DIFF
--- a/lib/SaveButton.js
+++ b/lib/SaveButton.js
@@ -4,11 +4,13 @@ var SaveButton = {
         options: {
             save: "Function to call when the user clicks Save",
             unsavedMessage: "Message to display when there are unsaved changes and the user leaves the page"
+            csrftoken: "Token required in headers of AJAX request to prevent CSRF"
         }
     */
     init: function (options) {
         var button = {
             disabled: false,
+            csrftoken: options.csrftoken,
             $save: $('<span/>').text(SaveButton.message.SAVE).click(function (e) {
                 button.fire('save', e);
             }).addClass('btn btn-success'),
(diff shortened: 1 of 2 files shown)


lib/SaveButton.js
```diff
@@ -51,7 +53,10 @@ var SaveButton = {
                     success = options.success || function () {},
                     error = options.error || function () {},
                     that = this;
-                options.beforeSend = function () {
+                options.beforeSend = function (xhr) {
+                    if (that.csrftoken && !this.crossdomain) {
```
millerdev: what's `this.crossdomain`?

calellowitz:
@millerdev 
crossDomain is a property of jquery ajax objects. Its true when the request goes to a different domain than the current page. That check prevents us from sending a request with the csrf token to anywhere other than hq. Looks like there was a typo though. the D should have been capitalized. fixed now

millerdev:
I guess there's no easy way to test this.

emord:
@calellowitz Does this need to have the header added as well (would only affect case management)?

https://github.com/dimagi/Vellum/blob/master/src/datasources.js#L124

(I think these are the only 2 ajax calls in vellum)

calellowitz:
@emord It shouldnt. GETs are considered safe requests, so the CSRF middleware lets them through by default.

PR #579, opened 2015-12-03
hide scrollbar for databrowser

emord:
http://manage.dimagi.com/default.asp?188390

code buddy @TylerSheffels

### DIFF
--- a/src/less-style/structure.less
+++ b/src/less-style/structure.less
@@ -62,11 +62,11 @@
   color: @white;
   font-size: 17px;
   font-weight: bold;
-  height: 25px;
+  height: @treeBrowserHeadHeight;
   padding: 0;
   text-align: center;
   z-index: 3;
-  line-height:25px;
+  line-height: @treeBrowserHeadHeight;
   width: 100%;
 
   > .pull-right {
--- a/src/less-style/structure.less
+++ b/src/less-style/structure.less
@@ -62,7 +62,7 @@
   color: @white;
   font-size: 17px;
   font-weight: bold;
-  height: 24px;
+  height: 25px;
   padding: 0;
   text-align: center;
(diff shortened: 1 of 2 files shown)


millerdev:
Seems like this will work. Would it be better to fix the thing that's causing the content to be sized larger than the `.fd-accessory-pane` height? (seems like that's the real bug)

src/less-style/structure.less
```diff
@@ -50,7 +50,7 @@
 
 .fd-accessory-pane {
   overflow-x: hidden;
-  overflow-y: auto;
+  overflow-y: hidden;
```
orangejenny: super picky, but this could now be a single `overflow` rule

millerdev:
GTM on test pass if fixing the content sizing would is too hard or if we want to get this in and come back to it later

emord:
@millerdev Looks like it's only off by one pixel, but I haven't been able to find where that 1 px comes from. I'll look into it again later.

emord:
Looks like it was because `line-height: 25px; height: 24 px`. Bumped up `height` to equal `line-height`

millerdev:
Excellent! this is looking good

> it was because `line-height: 25px; height: 24 px`.

What do you think about using a variable to make the two the same, thereby making the linkage explicit? Something like `@lineHeight: 25px; line-height: @lineHeight; height: @lineHeight;`

Or maybe remove the `line-height` (or maybe `height`) rule altogether given that `height` is being set and assuming there will never be enough text to wrap to multiple lines?

Feel free to just merge if this is getting too nit-picky or you feel it's not worth the effort.

emord:
Would have preferred just getting rid of one of the rules, but there's some funkiness with the minimize button if I do that (goes a little off center)

PR #580, opened 2015-12-07
Better error message

emord:
Suggestion from field-dev

code buddy @TylerSheffels

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1683,9 +1683,9 @@ define([
             var mugs = _this.getCurrentlySelectedMug(true),
                 errorMessage;
             if (mugs.length > 1) {
-                errorMessage = "You are about to delete multiple questions. This cannot be undone";
+                errorMessage = "Deleting multiple questions cannot be undone.";
             } else if (mugs.length && form.getChildren(mugs[0]).length) {
-                errorMessage = "You are about to delete a question with child nodes. This cannot be undone";
+                errorMessage = "Deleting a question with children cannot be undone.";
             }
             if (errorMessage) {
                 _this.alert(
--- a/src/core.js
+++ b/src/core.js
@@ -1680,11 +1680,17 @@ define([
                 isCopyable: !multiselect && mug.options.isCopyable
             }));
         $baseToolbar.find('.fd-button-remove').click(function () {
-            var mugs = _this.getCurrentlySelectedMug(true);
+            var mugs = _this.getCurrentlySelectedMug(true),
+                errorMessage;
+            if (mugs.length > 1) {
+                errorMessage = "You are about to delete multiple questions. This cannot be undone";
+            } else if (mugs.length && form.getChildren(mugs[0]).length) {
+                errorMessage = "You are about to delete a question with child nodes. This cannot be undone";
+            }
             if (mugs.length > 1 || (mugs.length && form.getChildren(mugs[0]).length)) {
                 _this.alert(
                     "Delete Questions?",
-                    "This cannot be undone.",
+                    errorMessage,
                     [{
                         title: "Cancel",
                     }, {
--- a/src/core.js
+++ b/src/core.js
@@ -1687,7 +1687,7 @@ define([
             } else if (mugs.length && form.getChildren(mugs[0]).length) {
                 errorMessage = "You are about to delete a question with child nodes. This cannot be undone";
             }
-            if (mugs.length > 1 || (mugs.length && form.getChildren(mugs[0]).length)) {
+            if (errorMessage) {
                 _this.alert(
                     "Delete Questions?",
                     errorMessage,


src/core.js
```diff
@@ -1680,11 +1680,17 @@ define([
...
         $baseToolbar.find('.fd-button-remove').click(function () {
-            var mugs = _this.getCurrentlySelectedMug(true);
+            var mugs = _this.getCurrentlySelectedMug(true),
+                errorMessage;
+            if (mugs.length > 1) {
+                errorMessage = "You are about to delete multiple questions. This cannot be undone";
+            } else if (mugs.length && form.getChildren(mugs[0]).length) {
+                errorMessage = "You are about to delete a question with child nodes. This cannot be undone";
+            }
             if (mugs.length > 1 || (mugs.length && form.getChildren(mugs[0]).length)) {
```
orangejenny: nitpick: making this `if (errorMessage)` would make it simpler to change the error condition logic in the future

emord: d'oh meant to do that and totally slipped my mind.

TylerSheffels:
Jenny beat me to it, :-1: after tests finish

src/core.js
```diff
@@ -1680,11 +1680,17 @@ define([
...
             }));
         $baseToolbar.find('.fd-button-remove').click(function () {
-            var mugs = _this.getCurrentlySelectedMug(true);
-            if (mugs.length > 1 || (mugs.length && form.getChildren(mugs[0]).length)) {
+            var mugs = _this.getCurrentlySelectedMug(true),
+                errorMessage;
+            if (mugs.length > 1) {
+                errorMessage = "You are about to delete multiple questions. This cannot be undone";
+            } else if (mugs.length && form.getChildren(mugs[0]).length) {
+                errorMessage = "You are about to delete a question with child nodes. This cannot be undone";
```
millerdev: I'm curious about the reasoning behind this change.

Unfortunately [users don't read](http://www.joelonsoftware.com/uibook/chapters/fog0000000062.html). What do you think about simplifying to a single short error message: "Deleting multiple questions cannot be undone."

millerdev: Notable quote from that article: "experience shows that the more words you put on that dialog box, the fewer people will actually read it."

millerdev:
> Jenny beat me to it, :+1: after tests finish

lol, in gmail that came out as

screenshot 2015-12-07 12 17 12

src/core.js
orangejenny: Agree that shorter & more direct is better, but if these are rephrased, I do think it's worth using two distinct messages here. Users are likely to think of a select as a single question.

orangejenny: Related, users don't say "node." "Select question" or "question and its choices" or "complex question."

emord: @orangejenny Would "question and it's children" be ok? complex question seems vague and this can also refer to groups/repeats

millerdev: Ugh, writing succinct error messages that are actually helpful is so hard. I do not like this confirmation dialog at all. It interrupts the workflow to ask a silly question that is unexpected and annoying. This is why I want to implement undo.

Another thing to consider is that people are often speed-reading, so they may only read the first few words, and maybe the last few words. Starting with "You are about to..." forces the reader to slow down and read more carefully. I'm standing by my suggestion: the succinctness of "Deleting multiple questions..." is worth the minor inaccuracy for select questions. An alternative that would avoid the inaccuracy would be to delete select questions without prompting, but that would make the implementation more complicated.

I'm really curious about the original suggestion that prompted this, but I have not been able to find it. Jon, can you point me to it?

emord: @millerdev  It was actually in a field report out from Erin, not field/dev. Here's the relevant text

> @Matt/Jon-- interesting! Now definitely makes sense why we are putting the pop-up box for some deletions and and not others. I'm wondering if there's a way we could communicate that (it was kind of confusing to me, and I would presume it might be for a new user, too). Maybe even something as simple as making the delete pop-up box text more specific-- e.g. instead of "you're about to delete a question. Are you sure" something along the lines of "you're about to delete multiple questions and/or questions that have child nodes" (make wording nicer, of course)

orangejenny: Agree "complex question" is vague, that was kind of a CYA suggestion.

Having two messages doesn't increase complexity/verbosity for the user, it only increases complexity for us.

millerdev: Ah, so she was confused at seeing it sometimes but not others, which is fair. I think adding a few extra words about deleting multiple questions should give a hint about why it shows up sometimes but not others.

I love how she concludes:
> Also, though good for new app builders, this pop up box is actually pretty annoying if you’re deleting questions en masse and know what you’re doing.

Right on. I think adding too much more text could make it even more annoying. :)

millerdev: Don't want to let you hanging @emord.

My vote is for one short message for both cases: "Deleting multiple questions cannot be undone."

But I'm willing to compromise. @orangejenny I'm not quite sure what your preference is. You said "shorter & more direct is better" and also want two separate messages. Do these seem good to you?

- `mugs.length > 1` -> "Deleting multiple questions cannot be undone."
- mug has children -> "Deleting a question with children cannot be undone."

orangejenny: Yeah, I think two short messages is the way to go. Your wording sounds fine to me.

emord: https://github.com/dimagi/Vellum/commit/eb70bfd13814a3fb69a7071f48bcbf73268496c1

millerdev:
:+1: from me. I'll let @orangejenny merge if she's happy

orangejenny:
works for me. appreciate everyone's attention to detail.

PR #581, opened 2015-12-07
Master z-index

orangejenny:
Don't merge yet - going use this locally for a few more days.

### DIFF
--- a/src/less-style/structure.less
+++ b/src/less-style/structure.less
@@ -4,7 +4,7 @@
   top: 0;
   width: 100%;
   background-color: white;
-  z-index: 1;
+  z-index: @zindexUIContainer;
 }
 
 .fd-toolbar {
@@ -60,29 +60,39 @@
 
 .fd-head {
   color: @white;
+  background-color: @cc-brand-low;
   font-size: 17px;
-  font-weight: bold;
+  font-weight: normal !important;
   height: @treeBrowserHeadHeight;
   padding: 0;
-  text-align: center;
-  z-index: 3;
+  text-align: left;
   line-height: @treeBrowserHeadHeight;
   width: 100%;
+  position: relative;
 
   > .pull-right {
     padding-right: 10px;
   }
+
   .fd-head-max-indicator i {
     color: @cc-neutral-hi;
   }
-}
 
-.fd-head:hover .fd-head-max-indicator i {
-  color: @white;
+  &:hover .fd-head-max-indicator i {
+    color: @white;
+  }
+
+  > h2 {
+    font-size: @baseFontSize * 0.8  ;
+    padding: 1px @contentPaddingVellum 0;
+    margin: 0;
+    font-weight: normal;
+    line-height: 23px;
+    text-transform: uppercase;
+  }
 }
 
 .fd-form-saving {
-  @zIndex: 1100;
   position: fixed;
   top: 0;
   bottom: 0;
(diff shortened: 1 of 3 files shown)


src/less-style/structure.less
```diff
@@ -4,7 +4,7 @@
   top: 0;
   width: 100%;
   background-color: white;
-  z-index: 1;
+  z-index: @zindexUIContainer;
```
orangejenny: Wasn't able to figure out why this uses a z-index at all, but leaving it alone.

src/less-style/structure.less
```diff
@@ -60,29 +60,39 @@
 
 .fd-head {
```
orangejenny: A bunch of these rules were getting overwritten by the second definition of `.fd-head` lower in this file

src/less-style/structure.less
```diff
@@ -98,7 +108,6 @@
     @size: @imageSize + @padding * 2;
 
     opacity: 1;
-    z-index: @zIndex + 1;
```
orangejenny: This was never necessary - my mistake from a while back.

src/less-style/tree-content.less
```diff
@@ -89,10 +89,6 @@
   }
 }
 
-#jstree-marker {
```
orangejenny: Didn't find any references to this id. Thought it might be the markup that appears when you drag something from the tree, but it isn't.

src/less-style/structure.less
```diff
@@ -60,29 +60,39 @@
...
 .fd-head {
   color: @white;
+  background-color: @cc-brand-low;
   font-size: 17px;
-  font-weight: bold;
+  font-weight: normal !important;
   height: @treeBrowserHeadHeight;
   padding: 0;
-  text-align: center;
-  z-index: 3;
```
orangejenny: Got rid of the z-index here altogether, couldn't find a reason it's necessary.

src/less-style/structure.less
emord: fullscreen mode on HQ. footer was covering vellum:

https://github.com/dimagi/Vellum/commit/87f102b7a166df1aebcfe60139c8c42444ddcb84

src/less-style/tree-content.less
```diff
@@ -115,7 +111,6 @@
 
 .fd-head-external-sources {
   background-color: @cc-neutral-low;
-  z-index: 0;
```
orangejenny: This was added as a bugfix for the language dropdown appearing below the header (https://github.com/dimagi/Vellum/commit/c2f7777404da8edb46b5940a67d6446908210cfc), isn't necessary now that `.fd-head` doesn't have a z-index.

src/less-style/tree-content.less
orangejenny: Found where this was added (https://github.com/dimagi/Vellum/pull/291), looks like it was needed to override full-screen mode's z-index. Full-screen no longer uses z-index, so this should be safe to remove. Tested that I can drag & drop questions in full-screen mode.

emord: :+1: there are still references to this inside bower_components in the jstree library. I don't think we reference it directly though

src/less-style/structure.less
orangejenny: Found https://github.com/dimagi/Vellum/pull/520, will add comment.

millerdev:
Looking great. Thanks for digging into this!

orangejenny:
Did a bit more testing locally in standalone and HQ. Good to merge.

PR #582, opened 2015-12-08
Atwho tests

emord:
Was a little hesitant changing some things with atwho, so decided to write some tests for it.

buddy: @TylerSheffels

### DIFF
--- a/src/atwho.js
+++ b/src/atwho.js
@@ -3,12 +3,14 @@ define([
     'jquery',
     'fusejs',
     'vellum/richText',
+    'vellum/util',
     'tpl!vellum/templates/atwho_display'
 ], function (
     _,
     $,
     fusejs,
     richText,
+    util,
     atwhoDisplay
 ) {
     var that = {};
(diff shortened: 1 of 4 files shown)


czue:
> Was a little hesitant changing some things with atwho, so decided to write some tests for it.

:heart:

src/atwho.js
```diff
@@ -56,7 +52,7 @@ define([
                     return choice.name && !_.isUndefined(choice.displayLabel);
                 })
                 .value();
-    }, 500);
+    }, 0);
```
millerdev: What's the practical outcome of this change? Does it mean this will be recalculated every time `cachedMugData` is called? If yes, can the `timed` decorator/wrapper be removed?

tests/atwho.js
```diff
@@ -0,0 +1,152 @@
...
+            }
+            before(function () {
+                $('.atwho-container').remove();
+                $("body").append(el);
+                input = el.children().first();
+                atwho.questionAutocomplete(input, mug);
+                input.val('/data/');
+                input.keyup();
+                atwhoview = getDisplayedAtwhoViews();
+                assert(atwhoview.length, 1);
```
millerdev: Should say `assert.equal`?

src/atwho.js
emord: Ah. that slipped in when I was debugging tests. https://github.com/dimagi/Vellum/pull/583

PR #584, opened 2015-12-08
Properly display mugs with unknown functions

emord:
http://manage.dimagi.com/default.asp?188772#1055825

@orangejenny 

cc: @TylerSheffels

### DIFF
--- a/src/richText.js
+++ b/src/richText.js
@@ -394,7 +394,7 @@ define([
 
         // only support absolute path right now
         if (!form.getMugByPath(xpath) && !/instance\('casedb'\)/.test(xpath)) {
-            return value.replace('<', '&lt;').replace('>', '&gt;');
+            return $('<span>').text(value);
         }
 
         return makeBubble(form, xpath, extraAttrs);
--- a/tests/richText.js
+++ b/tests/richText.js
@@ -260,6 +260,7 @@ define([
                     ['<output value="/data/text" /> <tag /> <output value="/data/othertext" />',
                      '{text} &lt;tag /&gt; {othertext}'],
                     ["{blah}", "{blah}"],
+                    ['<output value="unknown(/data/text)" />', '<span>&lt;output value="unknown(/data/text)"&gt;&lt;/output&gt;</span>'],
                 ],
                 ico = icon('fcc-fd-text');
 


orangejenny:
thank you :)

src/richText.js
```diff
@@ -394,7 +394,7 @@ define([
 
         // only support absolute path right now
         if (!form.getMugByPath(xpath) && !/instance\('casedb'\)/.test(xpath)) {
-            return value.replace('<', '&lt;').replace('>', '&gt;');
+            return $('<span>').text(value);
```
millerdev: Is this supposed to create a bubble? (I think it does not from local testing) If not, should it say `$('<span>').text(value).html()`? If yes, should it have a `label-datanode` class so it gets handled by the bubble widget?

If I'm not mistaken, this creates markup in ckeditor along the lines of:
```
label text <span>&lt;output value="unknown(/data/text)"&gt;&lt;/output&gt;</span> more text
```
which appears in the label ckeditor text area like this:
```
label text <output value="unknown(/data/text)" /> more text
```
What happens if the label is edited and the first 19 characters are deleted so it looks like
```
value="unknown(/data/text)" /> more text
```
Is the span element still in there? Does it get handled properly on getting the value out of the editor? Looks like it does in my local tests, but not sure if it will in all cases.

emord: > Is this supposed to create a bubble?

No, but it should return a jquery object: https://github.com/dimagi/Vellum/blob/a75987b6af73bcd53fe5df36f3c47638ba4301b6/src/richText.js#L388

> What happens if the label is edited and the first 19 characters are deleted

The span is there during the current question being edited. If the question is reloaded the span is gone as this function is only called for xpaths/output values.

This should not be an issue as all interactions that go through ckeditor should be done with getData which will rip out the span before writing to the mugs value.


the reason it was undefined before is here: https://github.com/dimagi/Vellum/blob/a75987b6af73bcd53fe5df36f3c47638ba4301b6/src/richText.js#L428 I could have also had that be able to handle text, but just made this function have a predictable output instead

millerdev: Thanks for the explanations. That makes a lot of sense that it `replacePathWithBubble()` needs to return a jquery object but it wasn't in some cases.

However, I think the resulting non-bubble thing should not be wrapped in a span tag. It's a little tricky to accomplish this. PR coming...

PR #585, opened 2015-12-09
Remove span tag around unrecognized output value

millerdev:
And humanize XML (`<output></output>` -> `<output />`).

See also https://github.com/dimagi/Vellum/pull/584

@emord

### DIFF
--- a/src/richText.js
+++ b/src/richText.js
@@ -394,7 +394,7 @@ define([
 
         // only support absolute path right now
         if (!form.getMugByPath(xpath) && !/instance\('casedb'\)/.test(xpath)) {
-            return $('<span>').text(value);
+            return $('<span>').text(value).contents();
         }
 
         return makeBubble(form, xpath, extraAttrs);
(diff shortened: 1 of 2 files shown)


src/richText.js
```diff
@@ -394,7 +394,7 @@ define([
 
         // only support absolute path right now
         if (!form.getMugByPath(xpath) && !/instance\('casedb'\)/.test(xpath)) {
-            return $('<span>').text(value);
+            return $('<span>').text(value).contents();
```
emord: Nice. I somehow didn't know about `contents`

src/richText.js
```diff
@@ -411,13 +411,15 @@ define([
             replacer, result;
         if (escape) {
             replacer = function () {
-                var id = util.get_guid();
-                places[id] = replacePathWithBubble(form, this.outerHTML);
+                var id = util.get_guid(),
+                    text = xml.humanize($("<div>").append($(this).clone()));
```
emord: Why not put this in `replacePathWithBubble` ?

millerdev: The value passed to `replacePathWithBubble()` is expected to be a string, and it's not convenient to convert `this` to a string that can easily be `xml.humanize`'d. I had tried `xml.humanize(this.outerHTML)`, but that didn't work (not sure why).

millerdev: Ah, further testing shows that I should be using `xml.normalize()` instead of `xml.humanize()`, which makes sense since ckEditor is an HTML editor, not a plain text editor. New PR coming.

PR #587, opened 2015-12-14
remove display before adding new one

emord:
http://manage.dimagi.com/default.asp?189183#1059677

cc: @TylerSheffels

### DIFF
--- a/src/templates/language_selector.html
+++ b/src/templates/language_selector.html
@@ -1,11 +1,9 @@
-<div id="language-selector">
-    <li class="divider"></li>
-    <li class="nav-header">Display</li>
-    <% _.each(languages, function(language) { %>
-    <li><a tabindex="-1" href="#" class="fd-display-item" data-code="<%-language.code%>"><i class="icon icon-ok"></i><%-language.name%><%
-        if (language.name !== language.code && language.code !== '_ids') {
-            %> (<%-language.code%>)<%
-        }
-        %></a></li><%
-    }); %>
-</div>
+<li class="divider language-selector"></li>
+<li class="nav-header language-selector">Display</li>
+<% _.each(languages, function(language) { %>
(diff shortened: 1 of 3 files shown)


src/templates/language_selector.html
```diff
@@ -1,4 +1,4 @@
-<ul>
+<div id="language-selector">
```
orangejenny: curious why this is now a div

emord: Before it said `<ul>`, but it was removed here: https://github.com/dimagi/Vellum/pull/587/files#diff-e5b3eca8407b45b132e0147b8392192cL1421 so it was just `<li>...`

That would make it very hard to remove here: https://github.com/dimagi/Vellum/pull/587/files#diff-e5b3eca8407b45b132e0147b8392192cR1421 I didn't see any visual differences by wrapping it in a div, although to be honest I wasn't sure if:
```
<ul>
  <li>
  <div>
    <li>
  </div>
</ul>
``` was a correct pattern (this is inserted into a list which is why it doesn't need `<ul>`

orangejenny: Ah, I get it now (hadn't read it closely before, just noticed the tag change). That markup - mixing divs and lis as children of a ul - is rather sketchy. Two alternate thoughts:
- Only append the language selector if it isn't already there (add a flag id/class to the divider or the nav-header)
- Give each of these LIs a class and remove everything with that class in _makeLanguageSelectorDropdown

emord: https://github.com/dimagi/Vellum/commit/d514f2fe0cfad864746fcf36e511dfa4527f0665

> Only append the language selector if it isn't already there (add a flag id/class to the divider or the nav-header)

Couldn't do that since technically you could add/remove a translation in source xml (although formbuilder might strip that out since it is passed in a list of valid languages)

orangejenny: was wondering why this markup is getting removed & re-added at all. that explains that.

orangejenny:
looks fine once tests pass

millerdev:
:+1:

emord:
:white_check_mark:

PR #589, opened 2015-12-14
minimum case property for databrowser

emord:
@millerdev 

cc: @TylerSheffels

### DIFF
--- a/src/databrowser.js
+++ b/src/databrowser.js
@@ -32,7 +32,7 @@ define([
             head = pane.find(".fd-head-external-sources");
             headHeight = head.outerHeight(true) || 0;
             pane.data("min-size", headHeight)
-                .height(tree.height() * DATABROWSER_HEIGHT)
+                .height(Math.min(tree.height() * DATABROWSER_HEIGHT, headHeight * 12))
                 .resize(function () {
                     if (pane.height() > headHeight + 100) {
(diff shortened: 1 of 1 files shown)


emord:
@amsagoff signed off on my local environment (tested with new small breadcrumbs)

src/databrowser.js
```diff
@@ -32,7 +32,7 @@ define([
             head = pane.find(".fd-head-external-sources");
             headHeight = head.outerHeight(true) || 0;
             pane.data("min-size", headHeight)
-                .height(tree.height() * DATABROWSER_HEIGHT)
+                .height(Math.min(tree.height() * DATABROWSER_HEIGHT, headHeight * 12))
```
millerdev: This makes the default data browser height whichever is smaller: 33% of the tree height or 300px (25px * 12). On very small screens (height < 900px) this will choose 33% and on larger screens (height > 900px) it will choose 300px. Did I get that right?

emord: That's correct. It's not the screen size though. It's the tree height. So, it's window size - browser toolbar - HQ header - breadcrumbs - vellum header - HQ footer

It looked ok locally with small breadcrumbs (on 1366px), but may change again depending on the coming UI changes

millerdev:
Kicked the build, gtm when tests pass.

PR #590, opened 2015-12-20
B3

orangejenny:
Opening for review - can't be merged until app manager in HQ is ready. Easiest to read commit by commit with whitespace off. Apologies for length, there isn't a great way to break this up. I also took a refactor pass through src/less-style, mostly adding nesting and comments.

Known issues:
- [x] icon to add a new intent callout question is missing
- [x] accordion icons don’t change when body is expanded or collapsed
- [x] modal to view form source spills out of boundaries for long forms
- [x] various TODOs marked in the PR (mostly functionality I don't know well and need to test)
- [x] horizontal scroll bar (https://github.com/dimagi/Vellum/pull/590#issuecomment-167625596)
- [x] remove stickyover?
- [x] on wide screens, the "Label" and "Turn Off Text Formatting" spans appear side-by-side instead of on top of each other
- [x] In the modal for overwriting someone else's changes, the XML diff spills out of the modal

Some screenshots:
screen shot 2015-12-19 at 10 02 59 pm
screen shot 2015-12-19 at 10 03 40 pm
screen shot 2015-12-19 at 10 04 19 pm
screen shot 2015-12-19 at 10 04 46 pm
screen shot 2015-12-19 at 10 06 48 pm


src/richText.js
```diff
@@ -90,13 +90,19 @@ define([
...
-                $(this.dragHandlerContainer.$).children("img").stickyover({
-                    title: displayId + '<small>' + xpath + '</small>',
+                // Remove ckeditor-supplied title attributes, which will otherwise override popover title
+                $(this.dragHandlerContainer.$).children("img").removeAttr("title");
+                // TODO: check with emord this can use popover...and if so, remove stickyover altogether
+                $(this.dragHandlerContainer.$).children("img").popover({
+                    trigger: 'hover',
+                    container: 'body',
+                    placement: 'bottom',
+                    title: '<h3>' + displayId + '</h3><div class="text-muted">' + xpath + '</div>',
```
millerdev: Should `xpath` be HTML-escaped here? I see it wasn't before, but looks like a possible opportunity for HTML injection.

src/expressionEditor.js
```diff
@@ -72,8 +72,8 @@ define([
             richTextOptions = {isExpression: true},
             form = options.mug.form;
         options = _.defaults(options, {
-            leftPlaceholder: "Hint: drag a question here.",
-            rightPlaceholder: "Hint: drag a question here.",
+            leftPlaceholder: "Drag question here",
+            rightPlaceholder: "Drag question here",
```
millerdev: Should be made consistent with https://github.com/dimagi/Vellum/pull/590/files#diff-b860087dfb0ae7f09a955a95f517a309R333, which reads "Drag a question here"

millerdev:
Looking great!

emord:
Looks great!

I did notice that in standalone vellum that minimizing the case properties pane caused some weird behavior with scroll bars. When I first open there was a horizontal and vertical scroll bar but when minimizing the pane the horizontal one went away. I'm not sure if this has any effect when embedding into HQ

src/expressionEditor.js
orangejenny: https://github.com/dimagi/Vellum/commit/94462c05777ea458c98a68abb6073be6d7f8d885

(I'm going for as few characters as possible, as the inputs are now a little smaller)

src/richText.js
orangejenny: sounds reasonable. made the `.text(foo).html()` construct its own function for readability.  https://github.com/dimagi/Vellum/commit/2845bfe2a4273514d682902110a5633bd1cd9f1c

orangejenny:
@emord chrome? what screen resolution? guessing that's related to screen size, but haven't been able to reproduce.

emord:
@orangejenny chrome. linux. screen res 1920 x 1080.

Looks like this on load 

selection_010

Then minimize (it's hard to see but there's the bottom of the screen, white space (size of scroll bar) case properties):

image

If I expand the case properties it starts to work correctly. Only happens on page laod

amsagoff:
With the new B3 padding, do we lose any screen height in terms of #
questions that fit in the tree? Thinking especially of resolution 1280 x 720

orangejenny:
@amsagoff Question tree is virtually identical in B2 and B3. In the properties window, we gain a little bit of vertical space because the accordion headers and sections are more compact now. Here are screenshots on 1280x800 (I can't do exactly 720 on this machine):

B2
screen shot 2015-12-30 at 12 09 55 pm

B3
screen shot 2015-12-30 at 12 09 46 pm

amsagoff:
Nice, thanks!

millerdev:
Still looking great!

emord:
can you remove .DS_Store and add a .gitignore for those?

also you changed datasourcewidgets.js to data_source_widgets.js. I think the current convention is to use that for HTML files, but js files use camel case like ignoreButRetain.js and javaRosa.js

orangejenny:
QA passed, this is good to go. Would like to deploy today (coordinating with Jeremy on on documentation changes that need to happen at the same time).

millerdev:
woohoo! \0/

czue:
holy :cloud: ! congrats!

PR #591, opened 2015-12-21
Encapsulate xpath library

emord:
encapsulate xpath into it's own file so we can be sure that vellum uses the same xpath parser throughout. uses new js-xpath

Doesn't really make a difference at the moment, but pulling it out from my current dev branch

cc: @esoergel

### DIFF
--- a/src/xpath.js
+++ b/src/xpath.js
@@ -1,0 +1,17 @@
+define([
+    'xpath',
+    'xpathmodels'
+], function (
+    xpath,
+    makeXPathModels
+) {
+    /* global xpathmodels */
+    /*jshint -W020 */
+    xpathmodels = makeXPathModels();
+    /*jshint +W020 */
+    
+    return {
+        xpath: xpath,
+        xpathmodels: xpathmodels
+    };
+});
--- a/src/core.js
+++ b/src/core.js
@@ -4,7 +4,7 @@ define([
     'require',
     'save-button',
     'underscore',
-    'xpathmodels',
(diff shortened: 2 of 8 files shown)


src/xpath.js
```diff
@@ -0,0 +1,17 @@
+define([
+    'xpath',
+    'xpathmodels'
+], function (
+    xpath,
+    makeXPathModels
+) {
+    /* global xpathmodels */
+    /*jshint -W020 */
+    xpathmodels = makeXPathModels();
```
millerdev: curious why this has to be global?

src/xpath.js
```diff
@@ -0,0 +1,17 @@
...
+    /*jshint +W020 */
+    
+    return {
+        xpath: xpath,
+        xpathmodels: xpathmodels
```
millerdev: what do you think of renaming this to `models` so the reference elsewhere would be `xpath.models`?

src/xpath.js
millerdev: nvm, figured it out

src/xpath.js
millerdev: https://github.com/dimagi/Vellum/pull/593

src/xpath.js
emord: Yeah. I couldn't think of a great way for it to not be global due to the way the xpath library is set up.

Would definitely be for cleaning it up so you pass the models to the parser, but not sure how feasible it is at the moment.

PR #592, opened 2015-12-22
absolutePath refactor

emord:
Refactor all of the random places that break the mug barrier when getting absolute path

cc: @esoergel


esoergel:
> **emord** wants to merge 2 commits into `master` from `rage-cleanup`

:rage4: 
Not sure I have enough context to merge this change, but it seems reasonable.

PR #593, opened 2015-12-22
Clean up xpath module

millerdev:
@emord

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -73,7 +73,7 @@ define([
     var isMac = /Mac/.test(navigator.platform);
 
     var DEBUG_MODE = false;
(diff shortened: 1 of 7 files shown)


millerdev:
Just had a case where a test timed out on travis, so I kicked the build. I think we're getting close to needing to increase the mocha test timeout above 2s (not ideal, really want to avoid this) or do some test refactoring to speed them up. My guess is that there's a reference leak somewhere.

emord:
@millerdev that's happened to me occasionally. Do you happen to remember what test it was? We should probably keep a list of the tests where this happens consistently.

millerdev:
Unfortunately I do not remember. I thought of making a note of it just after I kicked the build (too late).

PR #594, opened 2015-12-22
add parens to ordered list

emord:
http://manage.dimagi.com/default.asp?189603

cc: @esoergel

### DIFF
--- a/tests/markdown.js
+++ b/tests/markdown.js
@@ -103,7 +103,6 @@ require([
                 util.addQuestion("Text", 'markdown_question');
                 $('[name=itext-en-label]').val("1) first\n 2) second").change();
                 assert(markdownVisible());
-                $('.markdown-trigger').first().click();
             });
         });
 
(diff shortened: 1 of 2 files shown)


tests/markdown.js
```diff
@@ -97,6 +97,14 @@ require([
                 $('[name=itext-en-label]').val("no markdown").change();
                 util.assertXmlEqual(call('createXML'), NO_MARKDOWN_XML, {normalize_xmlns: true});
             });
+
+            it("should support ordered list with parens", function() {
+                util.loadXML("");
+                util.addQuestion("Text", 'markdown_question');
+                $('[name=itext-en-label]').val("1) first\n 2) second").change();
+                assert(markdownVisible());
+                $('.markdown-trigger').first().click();
```
millerdev: non-blocking: What's this line doing? If it's necessary, why not call `toggleMarkdown()` instead?

orangejenny:
Doesn't https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/app_manager/translations.py#L527 need to be updated?

PR #595, opened 2016-01-12
#form

emord:
@millerdev 

Commtrack and key value widgets (save to case and intent manager response in form builder) don't support rich text/hashtag syntax at the moment. Commtrack because it uses filters. key value widgets because we'd need to create a lot of ckeditor widgets (and the widget refactor would probably be a little messy for an advanced feature)

Ignore but retain also ignores #form syntax because those are manually written anyway.

When turning off bubbles it should reference /data/ instead of #form:

image

image

It's a big PR but a lot is XML/test changes to look at #form instead of /data The first commit should contain all xml changes and shows what xml will look like after this.

cc: @sravfeyn


src/javaRosa.js
```diff
@@ -1041,7 +1047,7 @@ define([
         };
 
         widget.save = function () {
-            widget.setItextValue(widget.getValue());
+            widget.setItextValue(outputToHashtag(widget.getValue()));
```
emord: Little bit of magic here. Always save the hashtag form in `mug.p.labelItext`

src/javaRosa.js
```diff
@@ -896,15 +900,17 @@ define([
...
 
         widget.getItextValue = function (lang) {
-            var itextItem = widget.getItextItem();
+            var itextItem = widget.getItextItem(), value;
             if (!lang) {
                 lang = widget.language;
             }
-            return itextItem && itextItem.get(widget.form, lang);
+            value = itextItem && itextItem.get(widget.form, lang);
+            return mug.supportsRichText() ? outputToHashtag(value) : outputToXPath(value);
```
emord: Displays the hashtag when rich text is supported, otherwise use old /data form

src/parser.js
```diff
@@ -22,13 +22,24 @@ define([
...
-        var attributes = $(element)[0].attributes,
-            attrMap = {};
+        var attributes = _.chain($(element)[0].attributes)
+                .map(function (value) {
+                    return [value.nodeName, value.nodeValue];
+                }) .object().value(),
+            attrKeys = _.keys(attributes);
+
+        _.each(attrKeys, function (key) {
+            if (key.startsWith('vellum:') &&
```
emord: This is used for `rawAttributes` so we want to ignore all the vellum attributes unless it's ignore or lock as those are processed elsewhere

src/widgets.js
```diff
@@ -188,7 +190,11 @@ define([
 
             var position = util.getCaretPosition(input[0]);
             var oldvalue = input.val();
-            input.val(value);
+            if (value && widget.hasLogicReferences) {
```
emord: The same magic as in javarosa. Show xpaths when bubbles are turned off.

src/xpath.js
```diff
@@ -5,8 +5,48 @@ define([
...
+            return toHashtag(xpath_.toXPath());
+        }
+    });
+
+    window.xpathmodels = xpathmodels;
+
+    return {
+        parser: xpath,
+        models: xpathmodels,
+        setHashtagToXPathDict: function (translationDict) {
```
emord: I originally included these three methods in logic, but tried to stick to a process of use logic manager only for when you need to keep track of the particular references and use xpath for normalizing/translating to hashtags

emord:
https://github.com/dimagi/Vellum/compare/b11d429...hashtag-form-final that ignores most of the xml changes

src/form.js
```diff
@@ -209,6 +217,7 @@ define([
         this.formName = 'New Form';
         this.mugMap = {};
         this.tree = new Tree('data', 'control');
+        xpath.addHashtag('#form', '/data');
```
millerdev: This is modifying global state. I understand that the global state it's modifying is an implementation detail of the xpath parser. But we should avoid building APIs that abuse that global state so it will be easy to port on the day that the xpath parser does not use global state (and then we can create an xpath parser instance with it's own local state for each form).

I suggest that we make a function like `form.addHashtag(hashtag, xpath)`, and that function will modify a `hashtagToXPath` object which is owned by the form. `hashtagToXPath` will be passed to `makeXPathModels()` and the result of that call is also owned by the form. That way all of the global state is left as only an implementation detail of the xpath parser library.

src/form.js
```diff
@@ -897,14 +907,27 @@ define([
...
         },
-        _updateMugPath: function (mug, oldPath, newPath) {
-            var map = this.mugMap;
-            delete map[oldPath];
-            if (_.isUndefined(newPath)) {
+        _updateMugPath: function (mug, oldHashtag, newHashtag) {
+            var map = this.mugMap, newPath;
+            delete map[normalizeHashtag(oldHashtag)];
+            if (oldHashtag) {
+                xpath.removeHashtag(oldHashtag);
```
millerdev: `removeHashtag` should also be a method on the form.

Actually, maybe it should remain a method of the `xpath` object, but the `xpath` object should be created and owned by the form rather than being a global. That's a better idea than what I outlined above because it encapsulates the details of creating an xpath parser within the `xpath` module rather than bloating the form with those details + methods. If you go that route then this call will become `form.xpath.removeHashtag(oldHashtag)`

millerdev: To create a new xpath parser, the form constructor should do something like
```js
this.hashtagToXPath = {};
this.xpath = xpath.createParser(this.hashtagToXPath);
```
`xpath.createParser` should create the object that is currently exported by the `vellum/xpath` module.

src/form.js
```diff
@@ -27,6 +25,16 @@ define([
         exporter = e;
     });
 
+    function normalizeHashtag(xpath_) {
```
millerdev: What do you think about making this a method of the `xpath` object?

src/ignoreButRetain.js
```diff
@@ -163,6 +161,7 @@ define([
         },
         parseBindElement: function (form, el, path) {
             if (this.data.ignore.active) {
+                path = xpath.parser.parse(path).toXPath();
```
millerdev: I forget, is it possible that this could throw an exception if `path` is not a invalid xpath expression? Would it be worth adding a test for that case?

src/javaRosa.js
```diff
@@ -1573,6 +1579,27 @@ define([
...
+                                output = $(this),
+                                value = output.attr('vellum:value') || output.attr('value'),
+                                ref = output.attr('vellum:ref') || output.attr('ref');
+                            if (value) {
+                                return tempOutput.attr('value', xpath.parser.parse(value).toHashtag())[0].outerHTML;
+                            } else if (ref) {
+                                return tempOutput.attr('ref', xpath.parser.parse(ref).toHashtag())[0].outerHTML;
+                            }
+                        });
+                        itForm.setValue(lang, value.html().replace(/&lt;/g, '<').replace(/&gt;/g, '>'));
```
millerdev: Should/could this use `xml.normalize()` instead of these replace calls?

src/javaRosa.js
```diff
@@ -1723,6 +1750,19 @@ define([
             // will be properly set as such.
             // 4. duplicate itext ids will be automatically updated to create
             // non-duplicates
+            
+            function hashtags(outputRef) {
+                var value = $(outputRef).attr('value') || $(outputRef).attr('ref'),
+                    key = !!$(outputRef).attr('value') ? 'value' : 'ref',
```
millerdev: is `!!` necessary here?

src/javaRosa.js
```diff
@@ -1723,6 +1750,19 @@ define([
...
+                    parsed = xpath.parser.parse(value),
+                    hashtag = parsed.toHashtag(),
+                    xpath_ = parsed.toXPath();
+                if (xpath_ === hashtag) {
+                    return '<output ' + key + '="' + xpath_ + '"/>';
+                } else {
+                    return '<output ' + key + '="' + xpath_ + '" vellum:' + key + '="' + hashtag + '"/>';
```
millerdev: Is it possible that `xpath_` could ever contain a quote character (`"`)? If it does this will construct a broken tag.

src/logic.js
```diff
@@ -83,8 +85,10 @@ define([
                         }
                     } else if (node instanceof xpath.models.XPathFuncExpr) {
                         this._addInstanceRef(node);
+                    } else if (node instanceof xpath.models.HashtagExpr) {
+                        hashtags.push(node);
                     }
-                    children = node.getChildren();
+                    children = node.getChildren ? node.getChildren() : [];
```
millerdev: What kind of node does not have a `getChildren` method? I assume from this context it's a `HashtagExpr`? If that's the case, it would be good to implement `HashtagExpr.getChildren()` to return `[]`.

src/modeliteration.js
```diff
@@ -336,18 +336,21 @@ define([
...
             } else {
                 oldPath = currentPath + "/item";
+                oldHash = hashPath + "/item";
                 if (/\/@count$/.test(mug.p.repeat_count)) {
                     mug.p.repeat_count = "";
                 }
             }
             mug.form.vellum.handleMugRename(
-                mug.form, mug, nodeID, nodeID, currentPath, oldPath, oldParent);
+                mug.form, mug, nodeID, nodeID, hashPath, oldHash, oldParent);
```
millerdev: Looks like `oldPath` and `currentPath` are not used anymore. Rather than modifying the names here, would it make sense to change `mug.absolutePath` to `mug.hashtagPath` and not change this function in any other way?

src/parser.js
```diff
@@ -22,13 +22,24 @@ define([
...
+                key !== 'vellum:ignore' &&
+                key !== 'vellum:lock'
+              ) {
+                delete attributes[key.slice(7)];
+                attributes[key.slice(7)] = attributes[key];
```
millerdev: minor: why delete first? it's being set on the next line.

src/parser.js
millerdev: Looks like this causes all `vellum:attrname` attribute values (except the excluded ones) to be moved to `attrname`. Are we sure we want that for every `vellum:` attribute (except for the excluded ones) going forward? Can we white-list attributes that we want to be translated in this way rather than automatically translating every single attribute (except for `vellum:ignore` and `vellum:lock`, of course) that starts with `vellum:`?

src/mugs.js
```diff
@@ -468,6 +468,20 @@ define([
         }
     });
 
+    Object.defineProperty(Mug.prototype, "hashtagPath", {
```
millerdev: The number of places where `mug.absolutePath` has been replaced with `mug.hashtagPath` is ... staggering. I'm wondering if it would have made more sense to change the implementation of `mug.absolutePath` to return what is now returned by `mug.hashtagPath`? Although on looking I see that there is still a lot of code that use `absolutePath`.

How does one know when to use each property? Is there a simple rule for that? Is it the case that `mug.hashtagPath` is now preferred and `mug.absolutePath` is deprecated? Whatever it is, it would be good to document it here.

millerdev:
Initial review pass is complete.

src/parser.js
emord: Turns out this is not needed at all https://github.com/dimagi/Vellum/commit/5e0b29373a68543695c7fbeef24089c6570caba6

It was useful when I was working on this piece by piece since this function would magically make things work, but since I implemented parseVellumAttrs this is not needed because all vellum attributes will be parsed there.

src/form.js
emord: https://github.com/dimagi/Vellum/commit/2d906504a04da53f5058c12901ca6daecba961fb

src/ignoreButRetain.js
emord: https://github.com/dimagi/Vellum/commit/08f62410d15fa239a0dc4351267e2bcff49da5f4

src/javaRosa.js
emord: https://github.com/dimagi/Vellum/commit/8d89186850a30c04c29b99639e6352d26cb315ac

src/logic.js
emord: https://github.com/dimagi/Vellum/commit/70aabb77ac33a15e8d44ef3985214330c80e15b8

src/modeliteration.js
emord: https://github.com/dimagi/Vellum/commit/e79c3c5c4f890b9adf8e03fe91042f9f50c7365a

src/mugs.js
emord: I did this because it was much easier changing the calls one by one instead of changing absolutePath everywhere all at once.

The only places using `absolutePath` are:
* ignoreButRetain: I decided it shouldn't  support hashtags since those are handwritten
* core.js: this is for drag/drop when the mug doesn't support rich text
* form.js: when adding hashtag translations
* atwho.js: to search when doing /data/
* saveToCase.js: doesn't support because key value widget
* modeliteration: uses attributes which isn't supported by js-xpath's hashtag implementation

src/form.js
```diff
@@ -898,14 +837,27 @@ define([
...
+            var map = this.mugMap, newPath;
+            delete map[this.normalizeHashtag(oldHashtag)];
+            if (oldHashtag) {
+                this.removeHashtag(oldHashtag);
+            }
+            if (_.isUndefined(newHashtag)) {
                 newPath = mug.absolutePath;
+                newHashtag = mug.hashtagPath;
+            } else {
+                newPath = newHashtag.replace(/^#form/, '/' + this.tree.rootNode.rootNodeId);
```
millerdev: would it make sense to use `this.getBasePath(true)` instead of `'/' + this.tree.rootNode.rootNodeId`?

src/widgets.js
```diff
@@ -13,7 +14,8 @@ define([
     $,
     atwho,
     util,
-    richTextUtils
+    richTextUtils,
+    xpath
```
millerdev: `xpath` is unused

src/writer.js
```diff
@@ -246,7 +250,10 @@ define([
                 opts.writeCustomXML(xmlWriter, mug);
             }
             if (opts.writeControlRefAttr) {
-                xmlWriter.writeAttributeString(opts.writeControlRefAttr, mug.absolutePath);
+                var hashtag = mug.hashtagPath;
+                if (hashtag) {
```
millerdev: when would this be false? seems like the `if` is not needed since it was always written before

src/xpath.js
```diff
@@ -4,9 +4,35 @@ define([
...
 ) {
     return {
-        createParser: function () {
+        makeXPathModels: function (hashtagToXPath) {
+            return xpath.makeXPathModels({
+                isValidNamespace: function (namespace) {
+                    return namespace === 'form';
+                },
+                hashtagToXPath: function (hashtagExpr) {
+                    if (hashtagToXPath[hashtagExpr]) {
```
millerdev: should probably use `hasOwnProperty`

src/form.js
emord: https://github.com/dimagi/Vellum/commit/a3988314114d5cf601e933b80ff68b81d3fa23ca

src/widgets.js
emord: https://github.com/dimagi/Vellum/commit/cf152b09bd7a0cfe37c51759846fcb89b33d67bc

src/writer.js
emord: https://github.com/dimagi/Vellum/commit/31532cab72afd06d95cdad25d1550c3773cfd3c8

xpath fails when you pass in `""`, `null`, `undefined`, but we have various places where those are legitamate values that are passed in. `xmlWriter.writeAttributeString` handled these values by not writing

src/xpath.js
emord: https://github.com/dimagi/Vellum/commit/c35ba07bb96b1caf83ecfa4e8f8d9f177ddd382a

src/ignoreButRetain.js
emord: Double checked this and it would never actually get here because https://github.com/dimagi/Vellum/blob/master/src/parser.js#L599 would have failed to load the form already

src/mugs.js
millerdev: I think it makes sense to document that list in a comment here and try to keep it updated if/as things change.

millerdev:
Reviewed changes

orangejenny:
@emord test failure looks real

orangejenny:
@emord nevermind, fixing it

orangejenny:
@millerdev added https://github.com/dimagi/Vellum/pull/595/commits/cead74d364e2c0eafb582e63d2e8e8559969dd65 to fix test

millerdev:
@emord I think this is ready to merge, but wanted to clear with you because of the "review. don't merge" label. Do you know why that's still here?

emord:
Removed. looks like i added it when I added QA flag. not sure why

emord:
(should be merged with https://github.com/dimagi/commcare-hq/pull/10346)

src/fuse.js
```diff
@@ -55,7 +65,8 @@ define([
 
             return {
                 id: mug.ufid,
-                name: mug.absolutePath,
+                name: mug.absolutePath || mug.hashtagPath,
```
millerdev: non-blocking: when would a mug have empty `absolutePath` and non-empty `hashtagPath`?

src/fuse.js
```diff
@@ -0,0 +1,104 @@
...
+
+    function mugToData(mug) {
+        if (mug) {
+            var defaultLabel = mug.form.vellum.getMugDisplayName(mug);
+
+            return {
+                id: mug.ufid,
+                name: mug.absolutePath || mug.hashtagPath,
+                hashtagPath: mug.hashtagPath,
+                absolutePath: mug.absolutePath,
```
millerdev: non-blocking: when would a mug have empty `absolutePath` and non-empty `hashtagPath`?

millerdev: this comment was meant to be on line 68

emord: I forgot to remove that. fixed

millerdev:
woohoo!!

czue:
#awesome

PR #596, opened 2016-01-12
Speed up tests

emord:
Turns out cool features slow down tests

Before: `./test  103.82s user 1.18s system 96% cpu 1:48.64 total`
After: `./test  60.94s user 1.32s system 95% cpu 1:05.35 total`

Removes atwho from vellum tests. Atwho is currently tested by creating it's own input and has a mock form.

Depends on https://github.com/dimagi/commcare-hq/pull/9859

@millerdev 

cc: @sravfeyn

### DIFF
--- a/src/atwho.js
+++ b/src/atwho.js
@@ -17,21 +17,23 @@ define([
 
     // stripped down version of http://codepen.io/ImagineProgramming/storydump/javascript-memoization-timeout
     var timed = function timed(f, timeout) {
-        var cache = {}, time = 0;
+        var time = 0;
 
         return function TimedMemoizedFunction() {
             var now = +new Date(),
                 timedOut = (now - time) >= timeout,
-                form = arguments[0];
+                form = arguments[0],
+                cache = form.vellum.data.atwho;
+            // atwho isn't actually defined anywhere, but data is an object from core
 
-            if(timedOut || _.isUndefined(cache[form.formUuid])) {
-                cache[form.formUuid] = f.apply(f, arguments);
+            if(timedOut || _.isUndefined(cache)) {
+                cache = f.apply(f, arguments);
                 if (timedOut) {
                     time = now;
                 }
             }
 
-            return cache[form.formUuid];
+            return cache;
         };
     };
 
--- a/src/atwho.js
+++ b/src/atwho.js
@@ -23,17 +23,17 @@ define([
             var now = +new Date(),
                 timedOut = (now - time) >= timeout,
                 form = arguments[0],
-                cache = form.vellum.data.atwho;
-            // atwho isn't actually defined anywhere, but data is an object from core
+                atwhoData = form.vellum.data.atwho,
+                cache = atwhoData.cache;
 
             if(timedOut || _.isUndefined(cache)) {
-                cache = f.apply(f, arguments);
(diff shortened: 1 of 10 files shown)


emord:
Elapsed times from travis are down from 5:07 to 4:13

src/atwho.js
```diff
@@ -17,20 +17,21 @@ define([
 
     // stripped down version of http://codepen.io/ImagineProgramming/storydump/javascript-memoization-timeout
     var timed = function timed(f, timeout) {
-        var cache, time = 0;
+        var cache = {}, time = 0;
```
millerdev: This is the kind of module-level global that will leak between separate instances of vellum on the same page. The leak could be avoided by saving these as properties of `this.data.atwho` in [`init()`](https://github.com/dimagi/Vellum/blob/01155fa227dddfaadea348d72c511bfff99df107/src/uploader.js#L268) or [`loadData()`](https://github.com/dimagi/Vellum/blob/01155fa227dddfaadea348d72c511bfff99df107/src/lock.js#L25).

millerdev: However, it might be tricky to get to `vellum.data.atwho` in the places that reference this.

millerdev: If there is a mug in the place where you need to reference it, you can do `mug.form.vellum.data.atwho`

millerdev: Hmm, maybe I'm all wrong here. It looks like there is a cache created per atwho-enabled input on the page? If that's true, then it looks like it's local to the input so not leaking to other instances.

millerdev:
Let me know what you think about the comments on the atwho plugin cache. I think this is good to merge if you agree that there is no leak.

millerdev:
Oh, and awesome that you found this!

src/atwho.js
emord: Thanks I made this into a real plugin. I was originally just going for practicality since there should be a 1 to 1 from formuuid to vellum instance and it was getting thrown away every 500 ms anyways, but it's just as easy to do a deep reach into form to get the atwho data.

Looks like the caching was broken by a refactor. It should have been caching globally before but it wasn't caching at all

src/atwho.js
```diff
@@ -17,11 +17,14 @@ define([
...
-        var cache, time = 0;
+        var time = 0;
 
         return function TimedMemoizedFunction() {
             var now = +new Date(),
-                timedOut = (now - time) >= timeout;
+                timedOut = (now - time) >= timeout,
+                form = arguments[0],
+                cache = form.vellum.data.atwho;
+            // atwho isn't actually defined anywhere, but data is an object from core
```
millerdev: `form.vellum.data.atwho` is initialized [here](https://github.com/dimagi/Vellum/blob/32c3d34c99c0d57c098af3bfb2ec58ded4556599/src/base.js#L129). You might want to do something like
```js
cache = form.vellum.data.atwho.cache;
...
// later when refilling the cache
cache = form.vellum.data.atwho.cache = f.apply(f, arguments);
```
as is I'm not sure it's ever updating `form.vellum.data.atwho` with new cache values.

emord: nice catch: https://github.com/dimagi/Vellum/commit/c03def5661f176ff7d998b4ab58b81e13e983f34

millerdev:
:speedboat:

PR #599, opened 2016-01-25
Change to itemset

emord:
http://manage.dimagi.com/default.asp?191099#1079355

@dimagi/product implements moving from select to lookup table, but not the other way around because you can't delete "Lookup Table Data" in an itemset

cc: @sravfeyn

### DIFF
--- a/src/itemset.js
+++ b/src/itemset.js
@@ -227,7 +227,7 @@ define([
                             }
                             return '';
                         }
-                        return typeName === "SelectDynamic" ? "" : "Can only change to a dynamic single answer";
+                        return typeName === "SelectDynamic" ? "" : "Can only change to a Single Answer Lookup Table";
                     },
                     validChildTypes: ["Itemset"],
                     maxChildren: 1,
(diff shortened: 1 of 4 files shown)


src/itemset.js
```diff
@@ -234,7 +242,15 @@ define([
...
-                        return typeName === "MSelectDynamic" ? "" : "Can only change to a dynamic multiple answer";
+                        if (typeName.match(/^M?Select$/)) {
+                            if (mug.form.getChildren(mug).length > 0) {
+                                return "Cannot change to Multiple/Single Choice " +
+                                      "question if it has Choices. " +
+                                      "Please remove all Choices and try again.";
+                            }
+                            return '';
+                        }
+                        return typeName === "MSelectDynamic" ? "" : "Can only change to a dynamic single answer";
```
millerdev: this looks like an unintentional change: dynamic multiple answer -> dynamic single answer

src/itemset.js
```diff
@@ -296,7 +312,17 @@ define([
             var ret = this.__callOld();
             ret.push('filter');
             return ret;
-        }
+        },
+        changeMugType: function (mug, type) {
+            var changeToItemset = false;
+            if (mug.__className.match(/^M?Select/) && type.match(/^M?SelectDynamic$/)) {
+                changeToItemset = true;
+            }
```
millerdev: nitpick: above four lines could be condensed to
```js
changeToItemset = mug.__className.match(/^M?Select/) && type.match(/^M?SelectDynamic$/);
```

src/itemset.js
emord: THnaks. I changed it to use Single/Multiple Answer Lookup Table to match what the names are

PR #600, opened 2016-01-25
Cleanup

emord:
@orangejenny We had some overlapping cleanup in our branches so I pulled them out here

cc: @sravfeyn


.gitignore
```diff
@@ -8,3 +8,4 @@ npm-debug.log
 *~
 *.swp
 .idea/
+*.DS_Store
```
millerdev: I think the `*` is not needed.

millerdev:
Cancelled intermediate test run to ease travis load.

orangejenny:
Updated .gitignore per Daniel, and removed the change to dataSourceWidgets.js, which depends on other B3 changes.

Looks good to me.

PR #601, opened 2016-01-26
Change text for Required Checkbox

sheelio:
After a discussion with Amelia, wanted to change the text for required questions.
@amsagoff @dimagi/product


emord:
failure related to js-xpath. rekick build on #599 merge (although i doubt this would break a test)

PR #603, opened 2016-02-03
Copy/paste markdown

emord:
http://manage.dimagi.com/default.asp?215446

@millerdev @benrudolph

### DIFF
--- a/src/javaRosa.js
+++ b/src/javaRosa.js
@@ -1834,6 +1834,9 @@ define([
                     if (hasText && !value.autoId) {
                         data[name] = value.id;
                     }
+                    if (value.hasMarkdown) {
+                        data[name + ':hasMarkdown'] = value.hasMarkdown;
+                    }
                 };
                 options.deserialize = function (data, name, mug, errors) {
                     var item = mug.p[name],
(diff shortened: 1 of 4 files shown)


src/javaRosa.js
```diff
@@ -1834,6 +1834,9 @@ define([
                     if (hasText && !value.autoId) {
                         data[name] = value.id;
                     }
+                    if (value.hasMarkdown) {
+                        data[name + ':hasMarkdown'] = value.hasMarkdown;
```
millerdev: I imagine `value.hasMarkdown` is normally a boolean value, which means this code will be fine. Is there any possibility that it might not be a boolean value but still evaluate to `true`? If yes, do you think it would be worth changing this line to `data[name + ':hasMarkdown'] = true;`?

millerdev:
:+1: on test pass

src/javaRosa.js
emord: It will only ever be true or false, and I don't imagine that will change

https://github.com/dimagi/Vellum/blob/master/src/javaRosa.js#L51
https://github.com/dimagi/Vellum/blob/master/src/javaRosa.js#L1089

PR #604, opened 2016-02-03
Witchcraft

orangejenny:
In Chrome on OSX, the case properties panel's scroll bar is visible on top of the XML source modal.

http://stackoverflow.com/questions/16874546/strange-z-index-behavior-with-scrollbars-under-chrome

screen shot 2016-02-03 at 4 54 39 pm

@emord

### DIFF
--- a/src/less-style/structure.less
+++ b/src/less-style/structure.less
@@ -46,6 +46,9 @@
   overflow-x: hidden;
   overflow-y: auto;
   height: @containerHeight;
+
+  // Prevent scrollbars underneath the modal from bleeding through on OSX
+  transform: translate3d(0, 0, 0);
 }
 
 .fd-accessory-pane {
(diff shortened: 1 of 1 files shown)


emord:
Accurate branch name. Does this happen with the other modals?

orangejenny:
Oops, put the style on the wrong selector - it goes on the scrolling panel, not on the z-indexed panel. 

Yes, affects all modals, and all apparently scrolling containers underneath them. Thanks Google. A little odd that it hasn't come up in HQ, but I suppose there aren't many places outside of Vellum where we have scrollbars anywhere other than the far right of the screen.

orangejenny:
Or perhaps more accurately, thanks Apple.

millerdev:
:space_invader: :zap:

czue:
:japanese_goblin: :apple:

PR #605, opened 2016-02-08
Upgrade underscore

emord:
@millerdev @orangejenny 

Right now vellum is overridding HQ's underscore (1.6.0) with our own packged one (1.4.4) changes here (look at 1.5.0 through 1.6.0): http://underscorejs.org/#1.6.0

This upgrades underscore and stops it from being packaged in local-deps

As far as I know there's no bugs that this causes yet.

cc: @benrudolph

### DIFF
--- a/tests/form.js
+++ b/tests/form.js
@@ -361,7 +361,7 @@ define([
                 [{id: "old", src: "new://3"}, "old-1", "src"],
                 [{id: "old", src: "old://"}, "old", "src"],
                 [{id: "any", src: "old://"}, "old", "src"],
-                [{id: "known", src: null}, "known", "known://"],
+                [{id: "known", src: undefined}, "known", "known://"],
                 [{id: "blank", src: "blank://"}, "blank", "src"],
                 [{id: null, src: "new://5"}, "data-1", "src"],
                 //[{id: null, src: null}, "data-1", "src"],     Error!
--- a/bower.json
+++ b/bower.json
@@ -21,7 +21,7 @@
         "tests"
     ],
     "dependencies": {
-        "underscore": "~1.4.4",
+        "underscore": "~1.6.0",
         "jquery": "~1.11.1",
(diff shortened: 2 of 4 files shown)


tests/form.js
```diff
@@ -361,7 +361,7 @@ define([
                 [{id: "old", src: "new://3"}, "old-1", "src"],
                 [{id: "old", src: "old://"}, "old", "src"],
                 [{id: "any", src: "old://"}, "old", "src"],
-                [{id: "known", src: null}, "known", "known://"],
+                [{id: "known", src: undefined}, "known", "known://"],
```
emord: @millerdev I couldn't find a place where this would actually be `undefined`

The reason I changed it was for (http://underscorejs.org/#1.5.0):

> The defaults function now only overrides undefined values, not null ones.

emord:
with some extra work we could possibly revert this: https://github.com/dimagi/Vellum/commit/66c8642c7a6756c161da4b0d535d3ce71b7cd03b

tests/form.js
orangejenny: > The defaults function now only overrides undefined values, not null ones.

ha, this was the one change that caused problems when upgrading underscore in HQ's app manager

benrudolph:
oh this is exciting!

millerdev:
Oh, nice! I had forgotten that underscore was not in the "excluded" group of dependencies. I thought vellum was using HQ's underscore, and I thought that was the reason we couldn't upgrade (if I knew this I may have already upgraded to the latest version). Glad to see this happening.

tests/form.js
millerdev: I think `{id: "known", src: undefined}` is the same as `{id: "known"}`, fwiw (no action required)

millerdev:
is this good to merge, or does it need QA?

emord:
I don't think this needs QA.

PR #610, opened 2016-02-17
Fix button alignment in lookup table data

orangejenny:
@emord 

before
screen shot 2016-02-17 at 12 41 31 pm

after
screen shot 2016-02-17 at 12 41 39 pm

Switched to icon because "Edit" padding changed & "Edit" no longer fits cleanly (on my screen, which isn't all that small).


orangejenny:
test failure looks real, will fix

millerdev:
:+1: on test pass

PR #611, opened 2016-02-17
🍌

emord:
@millerdev 

New functionality:

image

cc: @snopoke

### DIFF
--- a/src/bananas.js
+++ b/src/bananas.js
@@ -134,9 +134,23 @@ define([
         return output;
     }
 
+    function Parser(xpathParser) {
+        return {
+            parse: function (input) {
+                var parsed = xpathParser.parse(toXPath(input, xpathParser));
+                parsed.toBanana = function() {
+                    return toBanana(input, xpathParser);
+                };
+                return parsed;
+            },
+            models: xpathParser.models,
+        };
+    }
+
     return {
         toBanana: toBanana,
         toXPath: toXPath,
         transform: transform,
+        Parser: Parser,
     };
 });
--- a/src/bananas.js
+++ b/src/bananas.js
@@ -1,0 +1,76 @@
+define([], function() {
+    var OUTSIDE_BANANA = 0,
+        INSIDE_BANANA = 1,
+        DELIMITER = "🍌";
+
+    function getSymbols(string) {
+        var index = 0;
+        var length = string.length;
+        var output = [];
+        for (; index < length - 1; ++index) {
+            var charCode = string.charCodeAt(index);
+            if (charCode >= 0xD800 && charCode <= 0xDBFF) {
+                charCode = string.charCodeAt(index + 1);
+                if (charCode >= 0xDC00 && charCode <= 0xDFFF) {
+                    output.push(string.slice(index, index + 2));
+                    ++index;
+                    continue;
+                }
+            }
+            output.push(string.charAt(index));
+        }
+        output.push(string.charAt(index));
+        return output;
+    }
+
+    function parse(input, transform) {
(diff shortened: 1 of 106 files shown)


src/bananas.js
```diff
@@ -0,0 +1,180 @@
+define([
+    'vellum/xpath',
+], function(
+    xpath
+) {
+    var OUTSIDE_BANANA = 0,
+        INSIDE_BANANA = 1,
+        DELIMITER = "🍌",
```
emord: lol. i figured it would be the least likely candidate to already be used in a file

czue:
:banana: s are the new :coffee: s 

:monkey_face:

src/richText.js
```diff
@@ -364,7 +366,7 @@ define([
      */
     function makeBubble(form, xpath, extraAttrs) {
         function _parseXPath(xpath, form) {
-            if (/instance\('casedb'\)/.test(xpath)) {
+            if (/^🍌#case/.test(xpath)) {
```
snopoke: using the delimiter directly without the constant make it pretty easy to miss

tests/bananas.js
```diff
@@ -0,0 +1,87 @@
...
+                    "🍌#case/type/prop🍌 = 🍌#case/type/prop2🍌",
+                    "#case/type/prop = #case/type/prop2",
+                    "prop = prop2",
+                ],
+                ["🍌🍌", "🍌", "🍌"],
+                ["🍊you glad I didn't use 🍌", "🍊you glad I didn't use 🍌", "🍊you glad I didn't use 🍌"],
+                ["🍌#case/type/prop🍌 = 🍌", "#case/type/prop = 🍌",  "prop = 🍌"],
+                ["🍌#case/type/prop🍌 = 🍌🍌", "#case/type/prop = 🍌", "prop = 🍌"],
+                ["🍌#case/type/🍌🍌prop🍌 = 🍌🍌", "#case/type/🍌prop = 🍌", "🍌prop = 🍌"],
+                ["🍌🍌#case/type/🍌🍌prop🍌 = 🍌🍌", "🍌#case/type/🍌prop🍌 = 🍌", "🍌#case/type/🍌prop🍌 = 🍌"],
```
snopoke: :100: :banana:

snopoke:
Don't have enough context to review properly

proteusvacuum:
Looks like you did a bunch of work to make this PR apeeling.

src/bananas.js
millerdev: I got overwhelmed by :banana:s last night when I originally tried to review this PR. They are cute, but I think there are serious disadvantages to using them as a delimiter:

- it's a [4-byte character](https://mothereff.in/byte-counter#%F0%9F%8D%8C)
- there's no easy way to type it on a standard keyboard
- using an obscure character doesn't mean you don't need to escape it properly (the data we're dealing with may contain it and we should handle that situation properly)

I personally don't like this delimited-bubbles thing at all (hashtag expression parsing just got even more complex, and I'm not convinced we need it), but if I had to vote I would say back-tick (`` ` ``) would be a better delimiter.

emord: Ha. yeah. I figured this wouldn't actually go out, but it was more fun to code with than boring ` :)

emord:
Closing this one because travis can't build using branches with unicode.

https://github.com/travis-ci/travis-ci/issues/5653

src/richText.js
emord: :+1: https://github.com/dimagi/Vellum/commit/0e701f9c0411b56af954b65b735eff0e3b253754

PR #614, opened 2016-02-17
Undo poc

emord:
blah

@dimagi/product going to put this on staging. Expecting this needs a trip through UAT?

cc: @snopoke


amsagoff:
Indeed - lets uat!

src/mugs.js
```diff
@@ -475,6 +475,17 @@ define([
...
     });
 
+    Object.defineProperty(Mug.prototype, "previousSibling", {
+        get: function () {
+            var tree = this.form.tree,
+                node = this.form.tree.getNodeFromMug(this),
+                parentMug = this.parentMug,
+                parentNode = parentMug ? tree.getNodeFromMug(parentMug) : tree.getRootNode(),
+                mugPosition = parentNode.children.indexOf(node);
+            return (mugPosition === 0) ? parentMug : parentNode.children[mugPosition - 1].getValue();
```
millerdev: What do you think of delegating to the tree? There's a lot of tree implementation details in here. This getter would then `return this.form.tree.getPreviousSibling(this)`

src/undomanager.js
```diff
@@ -0,0 +1,59 @@
...
+    $,
+    _,
+    UNDO_ALERT
+) {
+    function alertShown() {
+        return $('.fd-undo-delete').length;
+    }
+
+    function createAlert() {
+        $('.fd-scrollable-tree').prepend(UNDO_ALERT);
```
millerdev: I think the location of this alert might be problematic in large forms where the user has scrolled the question tree.

millerdev:
excellent!

czue:
super cool!

src/mugs.js
millerdev: Also, consider adding a docstring to explain what "previous sibling" means in the context of

- a question whose previous sibling is it's parent
- the first question in the tree
- any other non-obvious details

src/undomanager.js
emord: https://github.com/dimagi/Vellum/commit/908e94aeaceb9ddeb22f291e2915b0b6277aa69f

millerdev:
Looking good. Should we add *awaiting QA* to this until UAT is finished? Or were you planning to deploy it behind a flag?

emord:
Passed QA http://manage.dimagi.com/default.asp?217694

emord:
:tada:

czue:
:recycle:

PR #615, opened 2016-02-18
Bananas

emord:
@millerdev 

New functionality:

image

cc: @snopoke


emord:
accidentally against wrong branch

PR #616, opened 2016-02-18
Bananas

emord:
@millerdev 

New functionality:

image

cc: @snopoke


src/bananas.js
```diff
@@ -0,0 +1,185 @@
...
+                    // overestimate users or Vellum devs
+                    text += DELIMITER + currentReference;
+                    currentReference = "";
+                }
+            }
+        }
+
+        if (state === INSIDE_BANANA) {
+            // end of string, without the end delimiter
+            text += DELIMITER + currentReference;
```
millerdev: This is an error condition, correct? I think it should throw an exception (fail the transform) rather than guessing at the intent and trying to fix up broken input. It should fail in a way that developers can see that something is broken. Otherwise we have to support broken input forever and/or we'll be masking bugs in vellum. The same applies above in the `else { // end of string...` clause.

src/bananas.js
```diff
@@ -0,0 +1,185 @@
...
+
+    function getSymbols(string) {
+        var index = 0;
+        var length = string.length;
+        var output = [];
+        for (; index < length - 1; ++index) {
+            var charCode = string.charCodeAt(index);
+            if (charCode >= 0xD800 && charCode <= 0xDBFF) {
+                charCode = string.charCodeAt(index + 1);
+                if (charCode >= 0xDC00 && charCode <= 0xDFFF) {
```
millerdev: what are the magic codepoint numbers? would it make sense to use named constants?

src/bananas.js
```diff
@@ -0,0 +1,185 @@
...
+        }
+        return parsedHashtags.toHashtag();
+    }
+
+    /*
+     * transforms bananas based on transformFn
+     */
+    function transform(input, transformFn) {
+        if (!input) { return input; }
+        input = getSymbols(input);
```
millerdev: looks like input changes type from a string to an array. might want to use a new variable name to make that clear

millerdev:
In addition to sending the case references back to HQ as JSON, what if we put a JSON object containing case references (really all hashtags that reference external resources) in the the form XML? Something like this:
```xml
<h:html>
  <h:head>
    <vellum:hashtags>{"#case/type/property":null}</vellum:hashtags>
    ...
  </h:head>
  ...
</h:html>
```
Seems like that would have two advantages:
- It would make it easy to extract all case references from the form.
- Allow us to know what things are valid hashtags that can be converted to bubbles at form load time. I think this would make it unnecessary to escape bubbles with backticks.

The values stored in JSON object will be `null` for hashtags that can be converted directly to case references without any additional information. In the future when we have hashtags that are fuzzier and the xpath cannot be constructed directly from the hashtag text, we will put the XPath expression in the value.

src/form.js
```diff
@@ -4,7 +4,7 @@ define([
     'jquery',
     'vellum/tree',
     'vellum/logic',
-    'vellum/xpath',
+    'vellum/bananas',
```
millerdev: not a fan of this change. it's less meaningful to a new person looking at the code

src/form.js
```diff
@@ -133,34 +133,51 @@ define([
...
 
         //make the object event aware
         util.eventuality(this);
     }
 
     Form.prototype = {
-        addHashtag: function(hashtag, xpath) {
-            this.hashtagDictionary[hashtag] = xpath;
+        validHashtags: function() {
+            return _.keys(this.hashtagDictionary);
```
millerdev: I think this should just return `this.hashtagDictionary` so the code that references it can do a hash lookup rather than a linear search through an array. Better yet, change it to
```js
isValidHashtag: function(tag) {
    return this.hashtagDictionary.hasOwnProperty(tag);
}
```

src/logic.js
```diff
@@ -216,15 +216,18 @@ define([
                     pathString = isHashtag ? path.toHashtag() : path.pathWithoutPredicates(),
                     pathWithoutRoot = isHashtag ? '' : pathString.substring(1 + pathString.indexOf('/', 1)),
                     refMug = form.getMugByPath(pathString),
-                    xpath = path.toHashtag();
+                    xpath = path.toHashtag(),
+                    knownHashtag = pathString.startsWith('#case') && _.contains(form.validHashtags(), xpath);
```
millerdev: It's not clear to me why we are treating `#case` hashtags specially here. It seems that we may have other external hashtag references in the future that may need to hit this same code path. Would it make sense to do `pathString.startsWith('#')`?

millerdev: Same for all places below that are looking for `#case` literals.

emord:
> Allow us to know what things are valid hashtags that can be converted to bubbles at form load time. I think this would make it unnecessary to escape bubbles with backticks.

I don't think that's possible as we want to keep bubbles in between loads even if they are invalid.

Example:
1) reference #form/question1 in a calculate
2) delete #form/question1
3) save and exit form
4) when we come back to the form #form/question1 is no longer a valid hashtag, but we want it to stay a bubble. Without some form of escaping how would we know what to turn into a bubble?

There's also the case that if we just have a find/replace with valid hashtags, it doesn't actually know the context of the hashtag, so it could be a hashtag reference, or it could be a string. We can't always use the xpath parser because it's also possible that someone saved an invalid xpath.

src/richText.js
```diff
@@ -94,13 +97,13 @@ define([
...
                     title: '<h3>' + util.escape(displayId) + '</h3>' +
-                           '<div class="text-muted">' + util.escape(xpath) + '</div>',
+                           '<div class="text-muted">' + util.escape(widget.mug.form.normalizeHashtag(xpath)) + '</div>',
                     html: true,
                     content: '<p>' + labelText.text() + '</p>',
                     template: '<div contenteditable="false" class="popover rich-text-popover">' +
                         '<div class="popover-inner">' +
                         '<div class="popover-title"></div>' +
-                        '<div class="popover-content"><p></p></div>' +
+                        (isCase ? '' : '<div class="popover-content"><p></p></div>') +
```
millerdev: Would it make sense to invert this expression and use `isForm` (a new variable similar to `isCase`, but true if the path starts with `#form/`)? That way if we add more external reference hashtags similar to `#case/` we won't need to change this code.

src/richText.js
```diff
@@ -456,39 +459,8 @@ define([
...
-
-        _.each(paths, function(path) {
-            var newPath = replacePathWithBubble(form, path);
-            el.html(el.html().replace(
-                new RegExp(RegExp.escape(path).replace(/ /g, '\\s*'), 'mg'),
-                $('<div>').append(newPath).html()
-            ));
-        });
-        return el.html();
+        text = form.normalizeBanana(text).replace('<', '&lt;').replace('>', '&gt;');
```
millerdev: I'm not sure what's supposed to come out of `normalizeBanana()`, but this is a naive way to escape XML. I think `xml.normalize()` may be what you want.

tests/.jshintrc
```diff
@@ -20,6 +20,7 @@
         "requirejs": false,
         /* MOCHA */
         "describe": false,
+        "context": false,
```
millerdev: what does this do? I couldn't find any documentation on it.

millerdev:
So is this a correct statement of the problem we're solving with escaping hashtags:

User types something that looks like a hashtag into an xpath expression, and the xpath parser thinks it is a hashtag, but it's really not. We want to preserve that and not show it to the user as a (red?) bubble?

I thought the whole reason we picked `#` as a prefix for these special expression elements was because that would never happen. I think we should solve that problem by inlining any unrecognized hashtag expression directly in the xpath that gets sent to mobile. So if the user types this:

`#form/valid_question + #case/unknown + #case/real/property`

it will be displayed to the user as (imagine colored bubbles instead of HTML tags):

*&lt;blue>valid_question&lt;/blue> + &lt;red>#case/unknown&lt;/red> + &lt;blue>property&lt;/blue>*

it will be converted to this xpath:

`/data/valid_question + #case/unknown + instance("casedb")/whatever...`

Ideally the user should be able to place the cursor after the red bubble and press backspace to turn the bubble into text (*#case/unknown*) similar to editing Gmail email addresses and correct the spelling or choose a valid case property from the at-who list. If they really want the literal *#case/unknown* to be in the xpath then they will just have to deal with it turning into a red bubble when loading the form (actually preferably right after typing a space after *#case/unknown* on initial expression entry).

--

:banana: digression

@emord went :banana:s ... and now vellum has bubbles, bananas, hashtags, and xpath expressions.

Is it correct to say a banana is an escaped hashtag, which is presented to the user as a bubble in a rich text widget? And users know them as "bubbles" or maybe "widgets" and they can type hashtag expressions to create them and they will never use the term "banana" to refer to them?

Currently the vellum source inconsistently uses the terms "bubble" and "banana" to refer to the same thing (I think so anyway), and sometimes things coming out of a functions like `toBanana()` are assigned to variables named `hashtag`, which doesn't seem right.

FWIW, I think "bubble" is a better, if not great, noun to use for these new rich expression references, certainly no less confusing to end users than "banana." I'm sorry, but I am not amused. I think the term "banana" adds more confusion than clarity to the problem we're trying to solve.

millerdev:
Oof... that came off more harshly than I wanted it to. Next time I won't procrastinate on the review until Friday afternoon.

emord:
To answer the first half:

The problem we're solving by escaping the hashtags is:

User types in an incorrect xpath expression, for example `(#form/question = #case/type/prop`, or `format-date(#form/date, '%y)`. The xpath parser doesn't know how to interpret this as a `)` is missing from the first and a `'` is missing from the second. If they drag these question/properties in, it will display correctly the first time, but if they navigate away from the page and come back there is no way to create the bubbles in the UI as the parser fails.

We could use naive find and replace but that could cause problems with expresions like `#form/question = '#case/type/prop'` where `'#case/type/prop'` should be a string (would imagine that we never see this in the short term, but we don't really have a clear picture of the end of hashtags and what/how we will use them, so it's possible there will be a more real edge case in the future). 

Find and replace also wouldn't solve the issue of copy/pasting questions from other forms, which could contain references that are valid in the original form, but not the second. I was originally not concerned with that, but copying questions and bubbles not coming along with them seems strange. (it would doable if it is a valid xpath expression, but as noted above we have to support bubbles in invalid expressions)

We could also attempt to use the xpath parser, and fall back to some special parser to attempt to find all the hashtags. That wouldn't be difficult now as we have two types of  hashtags to look for, but could get very complex later on.

I think doing simple escaping solves all of these problems in a way that is still human readable and easily parseable.

For the second half:

Agreed bananas isn't a very meaningful word (much like mug, or vellum which I constantly find myself slipping out when talking about form builder in a non dev setting). I was only intending it to be used as the textual representation that is internal to us. I always thought of a "bubble" as a UI widget that was separate from hashtag/xpath. For instance bubbles on prod bubble when it detects an `instance('casedb')` reference, or `/data/`, but in the future will bubble for hashtags.

Anyways I chose banana as it was already in our vocabulary from the weekly syncs and it made developing a little less drab. Perhaps the method names should be converted to something similar to `toEscapedHashtag`?

Also agree that there are some inconsistencies in naming as I've thrown this last work together in the past week after FM UAT feedback (and wanted to get it into a round of QA while I leave for vacation next week). Will definitely fix those, so it's more clear.

src/bananas.js
emord: I left it like this because it seems strange to error on possibly valid inputs that contain a new delimiter

src/bananas.js
emord: https://github.com/dimagi/Vellum/commit/506e49c4646956c3c32e966a9c6d9394e3b855e5 with tests: https://github.com/dimagi/Vellum/commit/896f344200e3ebfc2b62799f282ddc6d02ac72f1

src/bananas.js
emord: https://github.com/dimagi/Vellum/commit/896f344200e3ebfc2b62799f282ddc6d02ac72f1

src/form.js
emord: https://github.com/dimagi/Vellum/commit/5d945abcef09e2261a7d9df2cc5a58647e01ed0a

src/form.js
emord: https://github.com/dimagi/Vellum/commit/3306ee7ebb4e345868240fc3ed5f89ba4b189828

src/logic.js
emord: https://github.com/dimagi/Vellum/commit/93c5baa56f52a7e3900bad65b22612a02372399f

src/richText.js
emord: https://github.com/dimagi/Vellum/commit/93c5baa56f52a7e3900bad65b22612a02372399f

src/richText.js
emord: https://github.com/dimagi/Vellum/commit/10aa8bf00fcce33ce58ea64ef0b56922b32eb4fc

tests/static/escapedHashtags/invalid-xpath.xml
```diff
@@ -0,0 +1,28 @@
...
+	<h:head>
+		<h:title>Untitled Form</h:title>
+		<model>
+			<instance>
+				<data xmlns:jrm="http://dev.commcarehq.org/jr/xforms" xmlns="http://openrosa.org/formdesigner/939C596B-AD0C-4BB8-8717-F00707F71909" uiVersion="1" version="1" name="Untitled Form">
+					<text />
+					<hidden />
+				</data>
+			</instance>
+			<bind vellum:nodeset="#form/text" nodeset="/data/text" type="xsd:string" vellum:relevant="#invalid/xpath (`#form/hidden`" relevant="(`#form/hidden`" />
```
millerdev: What do you think of removing the hashtag escape delimiters from the (invalid) non-`vellum:` attribute? Additionally, would it be possible to convert all hashtags to valid xpath sub-expressions?

So ``relevant="(`#form/hidden`"`` becomes ``relevant="(/data/hidden"``

That way invalid xpath (or at least anything that is not valid according to our parser) is sent through as close to valid xpath as possible. I can imagine a scenario where mobile supports a syntactical form that our parser does not recognize, and it would be nice if those expressions could be entered in the form builder and passed on to mobile without non-standard escaping characters.

src/escapedHashtags.js
```diff
@@ -0,0 +1,189 @@
...
+            output.push(string.charAt(index));
+        }
+        output.push(string.charAt(index));
+        return output;
+    }
+
+    /*
+     * extends xpath parser to be aware of escaped hashtags
+     */
+    function Parser(hashtagDictionary) {
```
millerdev: Should not be capitalized. The convention is to capitalize constructor functions that are used to instantiate objects with the `new` operator.

millerdev:
Is this all correct?

Escaped hashtags are used for two things:

- all hashtags in vellum-internal xpath expressions are escaped. This makes it easy and reliable to convert hashtags to bubbles in the rich text widget.
- xpath expressions that cannot be parsed by js-xpath are prefixed with `#invalid/xpath ` and hashtags are escaped before being written to XML.

Notably, valid xpath expressions are written in `vellum:` XML attributes without extra escaping. The escaping is unnecessary in this case because hashtags can be reliably located with our extended hashtag xpath parser.

This seems like a useful bit of documentation, maybe a docstring in *src/escapedHashtag.js*.

src/form.js
```diff
@@ -145,29 +145,49 @@ define([
     }
 
     Form.prototype = {
-        addHashtag: function(hashtag, xpath) {
-            this.hashtagDictionary[hashtag] = xpath;
+        isValidHashtag: function(tag) {
+            return this.hashtagDictionary.hasOwnProperty(this.normalizeHashtag(tag));
+        },
+        addHashtag: function(hashtag, xpath, dontOverwrite) {
```
millerdev: The databrowser module confusingly has a function with the same name and number of arguments, except the third argument is `vellum`. What do you think of making this a two-arg function `addHashtag(hashtag, xpath)`, which adds hashtag if it does not already exist, and adding a new function `overwriteHashtag(hashtag, xpath)`?

Also, reasoning about the default value of this `dontOverwrite` argument is confusing (it defaults to undefined, which means overwrite by default). Double negatives are hard.

src/form.js
```diff
@@ -145,29 +145,49 @@ define([
...
         },
-        normalizeXPath: function (xpath_, xpathParser) {
-            // if it's not an xpath just return the original string
+        normalizeHashtag: function (xpath_) {
+            return this.normalize('toHashtag', xpath_);
+        },
+        normalizeXPath: function (xpath_) {
+            return this.normalize('toXPath', xpath_);
+        },
+        hashtagsInXPath: function (xpath_) {
```
millerdev: Suggestion: rename to `getHashtagsInXPath`. At first I this would return a boolean value (true if there are hashtags in the xpath, false if not).

src/form.js
```diff
@@ -145,29 +145,49 @@ define([
...
+        normalize: function (methodName, xpath) {
             // try catch is needed as workaround for having an itemset without
-            // the itemset plugin enabled
+            // the itemset plugin enabled and invalid xpaths
             try {
-                return xpath_ ? this.xpath.parse(xpath_).toHashtag() : xpath_;
+                return xpath ? this.xpath.parse(xpath)[methodName]() : xpath;
             } catch (err) {
-                return xpath_;
+                return xpath.startsWith('#invalid/xpath ') ? xpath.slice(15) : xpath;
```
millerdev: Should this expression be re-raised when `methodName !== "toEscapedHashtag"`?

src/form.js
```diff
@@ -844,7 +864,7 @@ define([
         },
         _updateMugPath: function (mug, oldHashtag, newHashtag) {
             var map = this.mugMap, newPath;
-            delete map[this.normalizeHashtag(oldHashtag)];
+            delete map[this.normalizeEscapedHashtag(oldHashtag)];
```
millerdev: Why do the keys to `mugMap` need to be passed through `normalizedEscapedHashtag`? It seems like those keys are never invalid xpath, nor do they need to be converted to bubbles, which seem like the only reasons you'd want to use an escaped hashtag.

src/logic.js
```diff
@@ -216,15 +216,18 @@ define([
                     pathString = isHashtag ? path.toHashtag() : path.pathWithoutPredicates(),
                     pathWithoutRoot = isHashtag ? '' : pathString.substring(1 + pathString.indexOf('/', 1)),
                     refMug = form.getMugByPath(pathString),
-                    xpath = path.toHashtag();
+                    xpath = path.toHashtag(),
+                    knownHashtag = pathString.startsWith('#case') && form.isValidHashtag(xpath);
```
millerdev: Why does this need to check for `startsWith('#case')`? Same below.

src/logic.js
```diff
@@ -237,6 +240,12 @@ define([
             _.each(expr.instanceRefs, function (ignore, id) {
                 form.referenceInstance(id, mug, property);
             });
+            _.each(expr.hashtags, function (hashtag) {
+                if (/^#case/.test(hashtag.toHashtag())) {
+                    form.referenceInstance('casedb', mug, property);
+                    form.referenceInstance('commcaresession', mug, property);
```
millerdev: I think we do not want to make the logic manager have to know which hashtags reference which instances. It doesn't seem like the right place for it because of the hard coded `#case` and `casedb` and `commcaresession`, which are all form properties. What about moving this into the form with an API like `form.referenceHashtag(hashtag, mug, property)`? The form knows about hashtag namespaces like `#form` and `#case`, and it knows about instances, so adding the mapping between them there seems better than here.

src/logic.js
```diff
@@ -368,6 +377,44 @@ define([
...
+                .filter(function(ref) {
+                    return ref.path.startsWith('#case');
+                })
+                .map(function(ref) {
+                    var info = ref.path.split('/'),
+                        prop = info[2];
+                    if (prop === 'case_name') {
+                        prop = 'name';
+                    }
+                    return [ref.sourcePath.replace(/^#form/, '/data'), prop];
```
millerdev: Every place where this PR matches a prefix like this it should include the `/` after the prefix. So this should be `/^#form\//`.

millerdev: also, why not use the xpath parser to convert hashtags to xpath here?

tests/copy-paste.js
```diff
@@ -457,7 +457,7 @@ define([
                 ["id", "type", "calculateAttr"],
                 ["/radius", "DataBindOnly", "42"],
                 ["/pi", "DataBindOnly", "3.1415"],
-                ["/circumference", "DataBindOnly", "2 * #form/pi * #form/radius"],
+                ["/circumference", "DataBindOnly", "2 * `#form/pi` * `#form/radius`"],
```
millerdev: How hard would it be to use the same convention here as in XML?
- prefix invalid xpath expressions with `#invalid/xpath ` and escape hashtags
- do not escape hashtags in valid xpath expressions

millerdev: Seems semi-likely that humans may edit this format (in Excel), and it would be nice to not make them jump through more syntax hoops.

src/escapedHashtags.js
emord: https://github.com/dimagi/Vellum/pull/616/commits/4f4c9eed46a69865fac73a4c5db73cf2af8d356e

src/form.js
emord: https://github.com/dimagi/Vellum/pull/616/commits/4f4c9eed46a69865fac73a4c5db73cf2af8d356e

src/form.js
emord: https://github.com/dimagi/Vellum/pull/616/commits/aff80d6fd4a552e2d1ca6ff93bece3e7d3952835

src/logic.js
emord: https://github.com/dimagi/Vellum/pull/616/commits/0000beedb5c49b3e0be6b5a9c9977e20f765c195

src/logic.js
emord: https://github.com/dimagi/Vellum/pull/616/commits/e8868117a4a04ec1d62ebea4b21e9352f8a9a33c

https://github.com/dimagi/Vellum/pull/616/commits/2294c9e9443f667a352688cbc15b3866b922b04d

src/form.js
emord: https://github.com/dimagi/Vellum/pull/616/commits/89de3cff4f2ebfe33a6537a4660f0ff2d710ea26

tests/static/escapedHashtags/invalid-xpath.xml
emord: https://github.com/dimagi/Vellum/pull/616/commits/79041453649f61433a1b700f6457f702bc3ce03b

tests/copy-paste.js
emord: https://github.com/dimagi/Vellum/pull/616/commits/257cc28bfe24e9c9abd9b6a3713c000c6653ffb1

tests/copy-paste.js
```diff
@@ -444,15 +444,15 @@ define([
             util.loadXML("");
             paste([
                 ["id", "type", "calculateAttr", "conflictedNodeId"],
-                ["/radius", "DataBindOnly", "42", "null"],
```
emord: Copy/paste never actually used this format as JSON.parse would parse this into the number 42, not a string. Which means that if you opened the browser on this test you could never select radius or pi questions.

src/mugs.js
```diff
@@ -719,6 +719,15 @@ define([
             data.instances = _.extend(data.instances || {},
                                       mug.form.parseInstanceRefs(value));
         }
+        try {
+            if (value) {
+                value = mug.form.xpath.parse(value.toString()).toHashtag();
+            }
+        } catch (err) {
+            if (_.isString(value) && !value.startsWith('#invalid/')) {
+                value = '#invalid/xpath ' + value;
```
millerdev: how is this deserialized? is the `#invalid/xpath ` prefix stripped before displaying to the user? where does that happen (it doesn't appear to be handled below in `deserializeXPath`, which seems not great because serialize/deserialize are supposed to be symmetrical.

src/util.js
```diff
@@ -3,7 +3,7 @@ define([
     'underscore',
     'jsdiff',
     'vellum/markdown',
-    'vellum/xpath',
+    'vellum/escapedHashtags',
```
millerdev: curious why this was renamed? this module handles standard xpaths, xpaths with hashtags, and xpaths with escaped hashtags. All of those are derivatives of xpath. It's not a big deal and I'm not requesting a rename, just wondering why it was changed?

tests/copy-paste.js
```diff
@@ -989,6 +989,22 @@ define([
...
         });
 
+        it("should paste invalid xpaths correctly", function() {
+            util.loadXML("");
+            paste([
+                ["id", "type", "calculateAttr"],
+                ["/invalid", "DataBindOnly", "(42"],
+                ["/invalid2", "DataBindOnly", "#invalid/xpath (42"],
+            ]);
+            util.clickQuestion('invalid');
```
millerdev: seems like it might be worth checking the value of `util.getMug("invalid").p.calculateAttr` and also the Calculate Condition input field to verify that they do or do not have the `#invalid/xpath ` prefix (I'm not sure which is correct in each case).

Also might be worth adding a test with an escaped hashtag in the (invalid) expression to verify that that is preserved and properly converted to a bubble in the text input on paste.

src/util.js
emord: This didn't actually change. `xpath` just wasn't used, since the form contains its own xpath parser. it is now using a utility from `escapedHashtags`

src/mugs.js
emord: Looks like I was confused between serialize/deserialize. Symmetrical now https://github.com/dimagi/Vellum/pull/616/commits/61bcf758e46dae89157b848f3c886ac403222baf

emord:
Updated with:

1. allows output ref (changes it to output value on write) https://github.com/dimagi/Vellum/pull/616/commits/119ded8797381a0aa8331ed249b13a5b9cc80d61
2. updated tests to use rich_text feature flag and only write vellum attributes when the form explicitly enables bubbling https://github.com/dimagi/Vellum/compare/119ded8797381a0aa8331ed249b13a5b9cc80d61...a1c7999cb542f9993fa42274ab4498a4d98280ff?expand=1

src/mugs.js
millerdev: are bubbles in `value` escaped with backticks? Is there a test for serialize/deserialize that contains an expression with bubbles (that exercises this code)?

src/form.js
millerdev: still thinking about this?

tests/copy-paste.js
emord: https://github.com/dimagi/Vellum/pull/616/commits/2edc65e5e70cac57ddad732524c79191bd00ac09

src/mugs.js
emord: The copy-paste tests should be testing this I believe https://github.com/dimagi/Vellum/pull/616/commits/404506461f7f5e246b5c707b19fd7b1d8efd69bc https://github.com/dimagi/Vellum/pull/616/commits/1f4c067fcd82a9acbe99c293cbc3f93288fe0e2a

src/form.js
emord: Ah. I see what you mean now. 

Looks like I did this kind of poorly and there are a few edge cases that depend on this. This is here, because it lazily removes #invalid/xpath from any mug properties since they can come from a few places (like parser plugins and copy-paste). I wanted it to be sure that #invalid/xpath was didn't cause an error for the user , but now I'm not sure that's the best option.

It's possible that this should be a code clean up item

src/mugs.js
millerdev: :+1:

src/form.js
```diff
@@ -1035,7 +1035,7 @@ define([
             return value;
         },
         isCaseReference: function (path) {
-            return /instance\('casedb'\)/.test(path) ? true : false;
+            return /^#case/.test(path);
```
millerdev: should check for trailing slash `/^#case\//`

src/form.js
millerdev: OK, not going to block on this, but I think it's important to revisit in a separate PR. I think it should be changed to something like
```js
} catch (err) {
    if (methodName !== "toEscapedHashtag") {
        throw err
    }
    return xpath.startsWith('#invalid/xpath ') ? xpath.slice(15) : xpath;
}
```
`toEscapedHashtag` is the only normalize method that knows or cares about the `#invalid/xpath ` prefix, and other methods should not be modifying the output or swallowing expressions that happen to have that prefix.

czue:
:banana:

PR #617, opened 2016-02-19
Don't deselect mug on collapse

emord:
http://manage.dimagi.com/default.asp?217706

cc: @snopoke


millerdev:
looks great!

I think unrelated bug:
- add two text questions in a new form (question1 and question2)
- click question1
- edit Display Condition in expression editor
- drag/drop question2 from tree into LHS expression input
- click question2

expected: prompt for save/cancel on expression editor
got: no promt, changes lost

PR #620, opened 2016-03-01
Move fuse to form

emord:
A refactor of how we do fuzzy finding.

Instead of caching in src/atwho.js. this moves fuse (fuzzy search library) to the form. This means that we no longer need to go through all mugs everytime an autocomplete is created, and in the future for #case references if datasources load while you are already autocompleting new references should pop up seamlessly

cc: @calellowitz

### DIFF
--- a/src/fuse.js
+++ b/src/fuse.js
@@ -1,0 +1,76 @@
+define([
+    'underscore',
+    'fusejs',
+    'vellum/util',
+], function (
+    _,
+    fusejs,
+    util
+) {
+    var FUSE_CONFIG = {
+        keys: ['label', 'name', 'absolutePath']
+    };
+
+    function Fuse(form) {
+        var _this = this;
+        this.form = form;
+        this.dataset = generateNewFuseData(form);
+        this.fusejs = new fusejs(this.dataset, FUSE_CONFIG);
+
+
+        function addToDataset(e) {
+            _this.dataset = _.filter(_this.dataset, function (mug) {
+                return mug.id !== e.mug.ufid;
+            });
+            _this.dataset.push(mugToData(e.mug));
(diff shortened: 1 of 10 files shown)


src/fuse.js
```diff
@@ -0,0 +1,76 @@
...
+], function (
+    _,
+    fusejs,
+    util
+) {
+    var FUSE_CONFIG = {
+        keys: ['label', 'name', 'absolutePath']
+    };
+
+    function Fuse(form) {
```
millerdev: non-blocking: a docstring explaining what this is would be nice

src/fuse.js
```diff
@@ -0,0 +1,76 @@
...
+            };
+        }
+        return null;
+    }
+
+    function generateNewFuseData (form) {
+        return _.chain(form.getMugList())
+                .map(mugToData)
+                .filter(function(choice) {
+                    return choice.name && !_.isUndefined(choice.displayLabel);
```
millerdev: non-blocking: didn't we decide to use `!== undefined` a while ago?

emord: Ah yeah. we did. This was copied from src/atwho.js, which I think predates that.

tests/atwho.js
```diff
@@ -3,154 +3,88 @@ define([
...
+
+        function displayAtwho(callback) {
+            var mug = util.clickQuestion('one')[0];
+            var input = $('[name=property-relevantAttr]');
+            input.val('/data/').keyup();
+            assert.strictEqual(getDisplayedAtwhoViews().length, 1);
+            try {
+                callback(mug);
+            } catch (err) {
+                throw err;
```
millerdev: this `catch` block (above two lines) can be omitted

PR #621, opened 2016-03-03
Revert "Move fuse to form"

orangejenny:
Reverts dimagi/Vellum#620

http://manage.dimagi.com/default.asp?219902

This line is causing the bug - https://github.com/dimagi/Vellum/pull/620/files#diff-7eb52b366866677666470e019283c8eaR1086 - but I'm not sure why. Reverting only that line causes other problems; seems safer to revert the whole PR.

https://github.com/dimagi/Vellum/pull/614, currently in QA, doesn't have this problem, but not being able to delete multiple questions (including selects and non-empty groups) is fairly annoying, better not to wait for QA to finish.

@millerdev 

fyi @emord


millerdev:
:+1: on test pass

orangejenny:
passing!

orangejenny:
Another fun fact: I haven't been able to write a test that exercises the bug, even a test that exactly mimics clicks:
```
// add questions
$(".fd-question-type[data-qType='Text']:first").click();
$(".fd-question-type[data-qType='Text']:first").click();

// shift+click to select all questions
$(".fd-tree .jstree-anchor:first").click();
var shiftClick = jQuery.Event("click");
shiftClick.shiftKey = true;
$(".fd-tree .jstree-anchor:last").trigger(shiftClick);

// click delete button, then modal confirmation button
$(".fd-button-remove").click();
$(".btn-primary:visible").click();
```
...so this is going to remain untested for now.

PR #624, opened 2016-03-16
Fix alert's auto id setting

emord:
http://manage.dimagi.com/default.asp?218243#1108422

Before it thought the correct auto id would be questionid-alert instead of questionid-constraintMsg. Looks like it's been broken for quite a while

@millerdev 

cc: @biyeun


tests/copy-paste.js
```diff
@@ -76,7 +76,6 @@ define([
                         "labelItext:hin-default",
                         "constraintMsgItext:en-default",
                         "constraintMsgItext:hin-default",
-                        "constraintMsgItext",
```
emord: Itext's shouldn't be copied because they're autoid. I don't think constraintMsgAttr needs to be copied either, but it's a bigger lift to do that.

PR #625, opened 2016-03-16
adding ga tracking for case management in the form builder

calellowitz:
@emord cc: @NoahCarnahan 
tracking when case properties are dragged and dropped into the form builder

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -764,10 +764,10 @@ define([
                         targetType = "Validation";
                         break;
                     case 'property-calculateAttr':
-                        targetType = "Calculation"
+                        targetType = "Calculation";
                         break;
                     default:
-                        targetType = "Expression Editor"
+                        targetType = "Expression Editor";
                         break;
                 }
                 if (path.contains('casedb')) {
--- a/src/core.js
+++ b/src/core.js
@@ -756,12 +756,12 @@ define([
 
             if (window.analytics) {
                 if (path.contains('casedb')) {
-                    window.analytics.usage("Case Management", "Form Builder", "Drag and Drop");
+                    window.analytics.usage("Case Reference", "Drag and Drop", "Calculation");
                 } else {
                     window.analytics.usage(
-                        "Question Reference",
+                        "Form Reference",
                         "Drag and Drop",
-                        _this.data.core.currentlyEditedProperty
+                        "Calculation"
                     );
                 }
             }
--- a/src/core.js
+++ b/src/core.js
@@ -770,14 +770,14 @@ define([
                         targetType = "Expression Editor";
                         break;
                 }
-                if (path.contains('casedb')) {
-                    window.analytics.usage("Case Reference", "Drag and Drop", targetType);
-                } else {
+                if (_this.data.core.form.isFormReference(path)) {
                     window.analytics.usage(
                         "Form Reference",
                         "Drag and Drop",
                         targetType
                     );
+                } else {
+                    window.analytics.usage("Case Reference", "Drag and Drop", targetType);
                 }
             }
         }
--- a/src/core.js
+++ b/src/core.js
@@ -756,7 +756,7 @@ define([
 
             if (window.analytics) {
                 if (path.contains('casedb')) {
-                    window.analytics.usage("Case Management", "Form Builder", "Click and Drag");
+                    window.analytics.usage("Case Management", "Form Builder", "Drag and Drop");
                 } else {
                     window.analytics.usage(
                         "Question Reference",
--- a/src/core.js
+++ b/src/core.js
@@ -755,11 +755,15 @@ define([
             }
 
             if (window.analytics) {
-                window.analytics.usage(
-                    "Question Reference",
-                    "Drag and Drop",
-                    _this.data.core.currentlyEditedProperty
-                );
+                if (path.contains('casedb')) {
+                    window.analytics.usage("Case Management", "Form Builder", "Click and Drag");
+                } else {
+                    window.analytics.usage(
+                        "Question Reference",
+                        "Drag and Drop",
+                        _this.data.core.currentlyEditedProperty
+                    );
+                }
             }
         }
 
--- a/src/core.js
+++ b/src/core.js
@@ -770,14 +770,14 @@ define([
                         targetType = "Expression Editor";
                         break;
                 }
-                if (_this.data.core.form.isFormReference(path)) {
+                if (_this.data.core.form.isCaseReference(path)) {
+                    window.analytics.usage("Case Reference", "Drag and Drop", targetType);
+                } else {
                     window.analytics.usage(
                         "Form Reference",
                         "Drag and Drop",
                         targetType
                     );
-                } else {
-                    window.analytics.usage("Case Reference", "Drag and Drop", targetType);
                 }
             }
         }
--- a/src/core.js
+++ b/src/core.js
@@ -755,13 +755,28 @@ define([
             }
 
             if (window.analytics) {
+                var targetType;
+                switch (target[0].id) {
+                    case 'property-relevantAttr':
+                        targetType = "Display";
+                        break;
+                    case 'property-constraintAttr':
+                        targetType = "Validation";
+                        break;
+                    case 'property-calculateAttr':
+                        targetType = "Calculation"
+                        break;
+                    default:
+                        targetType = "Expression Editor"
+                        break;
+                }
                 if (path.contains('casedb')) {
-                    window.analytics.usage("Case Reference", "Drag and Drop", "Calculation");
+                    window.analytics.usage("Case Reference", "Drag and Drop", targetType);
                 } else {
                     window.analytics.usage(
                         "Form Reference",
                         "Drag and Drop",
-                        "Calculation"
+                        targetType
                     );
                 }
             }
--- a/src/form.js
+++ b/src/form.js
@@ -960,8 +960,8 @@ define([
             this.vellum.afterSerialize();
             return value;
         },
-        isFormReference: function (path) {
-            return this.getMugByPath(path) ? true : false;
+        isCaseReference: function (path) {
+            return /instance\('casedb'\)/.test(path) ? true : false;
         }
     };
 
--- a/src/form.js
+++ b/src/form.js
@@ -959,6 +959,9 @@ define([
             var value = exporter.generateExportTSV(this);
             this.vellum.afterSerialize();
             return value;
+        },
+        isFormReference: function (path) {
+            return this.getMugByPath(path) ? true : false;
         }
     };
 
--- a/src/javaRosa.js
+++ b/src/javaRosa.js
@@ -1388,13 +1388,13 @@ define([
                     insertOutputRef(_this, target, path, mug);
                     if (window.analytics) {
                         window.analytics.usage("Output Value", "Drag and Drop");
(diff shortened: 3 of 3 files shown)


src/javaRosa.js
```diff
@@ -1393,6 +1393,9 @@ define([
             } else {
                 _this.__callOld();
             }
+            if (path.contains('casedb')) {
+                window.analytics.usage("Case Management", "Form Builder", "Click and Drag");
```
emord: Is this supposed to be only for labels or for labels and xpaths?

calellowitz: I think both. @krsdimagi or @amsagoff can correct if that is wrong

krsdimagi: Yep, for both.  Any time a case property is dragged out of the case property window.

emord: The general case should go here: https://github.com/dimagi/Vellum/blob/master/src/core.js#L757-L764

For output values it should go https://github.com/dimagi/Vellum/pull/625/files/cf5fbdbf288457b26a3cda2c3b33cb798eb90922#diff-e5b3eca8407b45b132e0147b8392192cR1390

Vellums uses a plugin structure and this function is overriding the one in core (and calls the old one (`_this.__callOld()`) if it's not an output value)

It also needs the `if (window.analytics)` wrapper so it doesn't fail in standalone. Not sure why we haven't made a helper for that yet

NoahCarnahan:
I think just one more thing outstanding:
Emord says you need to check for `windows.analytics` before accessing `usage()`

calellowitz:
@NoahCarnahan it looks to me like they are both in `if (window.analytics)` blocks. Do you see a place where they arent.

src/javaRosa.js
```diff
@@ -1388,6 +1388,9 @@ define([
                     insertOutputRef(_this, target, path, mug);
                     if (window.analytics) {
                         window.analytics.usage("Output Value", "Drag and Drop");
+                        if (path.contains('casedb')) {
+                            window.analytics.usage("Case Management", "Form Builder", "Click and Drag");
```
millerdev: Two different phrases are used to mean basically the same thing ("Drag and Drop" vs "Click and Drag"). Not sure if this is by design or if it could create two categories where one is preferred. Also, the arguments for the `usage` function are `category, action, label, value`. If I'm not mistaken action is "Drag and Drop" on line 1390 and "Form Builder" on line 1392—is that intended? @amsagoff may have input as well.

NoahCarnahan:
Gah, my bad @calellowitz, was only looking at the new lines.

emord:
Looks good pending @dimagi/product response to @millerdev comment

src/javaRosa.js
amsagoff: Good catch @millerdev, I forgot that I already had analytics for a similar event. Let's make the names consistent (so, "Drag and Drop")

emord:
Good pending tests

src/javaRosa.js
```diff
@@ -1388,6 +1388,9 @@ define([
                     insertOutputRef(_this, target, path, mug);
                     if (window.analytics) {
                         window.analytics.usage("Output Value", "Drag and Drop");
+                        if (path.contains('casedb')) {
+                            window.analytics.usage("Case Management", "Form Builder", "Drag and Drop");
+                        }
```
millerdev: The second part of my [previous comment](https://github.com/dimagi/Vellum/pull/625/files/21badeee3d5285ffb09b39f815bbcbed83197a51#r56562730) has not been addressed yet: the arguments for the `usage` function are `category, action, label, value`. If I'm not mistaken action is "Drag and Drop" on line 1390 and "Form Builder" on line 1392—is that intended?

millerdev:
Not sure what's up with the test failure. @orangejenny or @emord may have more insight.
```
Tracing dependencies for: main-components
{ [Error: ENOENT: no such file or directory, open '/home/travis/build/dimagi/Vellum/_build/bower_components/MediaUploader/swfobject.js']
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/home/travis/build/dimagi/Vellum/_build/bower_components/MediaUploader/swfobject.js',
  fileName: '/home/travis/build/dimagi/Vellum/_build/bower_components/MediaUploader/swfobject.js' }
Error: Error: Module loading did not complete for: main, vellum/uploader, file-uploader, swfobject
    at Function.build.checkForErrors (/home/travis/build/dimagi/Vellum/node_modules/requirejs/bin/r.js:31556:19)
```

millerdev:
I think the test failure may be related to dimagi/MediaUploader#16, which is now merged. Kicking travis.

calellowitz:
@millerdev The tracking properties for `category, action, and label`, were chosen based on https://docs.google.com/spreadsheets/d/1frMdFeznNcMAIyMW3pG3zes6mmY03UG67HyMUHXlb-s/edit?pref=2&pli=1#gid=0
@amsagoff can weigh in about the discrepancy between the old and the new versions and our departure from the semantic meanings of those terms but im hesitant to change that code without feedback from @dimagi/product if only because I dont have as deep an understanding of how they use google analytics

amsagoff:
Sorry I thought I had already commented but looks like I forgot to press "save"... @millerdev is right, the two should match. Let's make the new stuff consistent with the language already there, so we can change "click and drag" to "drag and drop" (I've changed it in the definition file)

calellowitz:
@amsagoff Updated that language. Still a question about the `action` vs `label`. In the old one the action is drag and drop while in the new one "drag and drop" is a label and the action is form builder. That is how it is described in the analytics sheet but I think @millerdev was concerned about the inconsistency

amsagoff:
I see what you're saying. It looks like I took two stabs at this in the
analytics doc and left one of them incomplete. Sorry to create confusion.

I do want form question / case property events structured the same way. I
have marked lines 87 and 88 as "deprecated" and re-specced on lines 101-102
here:
https://docs.google.com/spreadsheets/d/1frMdFeznNcMAIyMW3pG3zes6mmY03UG67HyMUHXlb-s/edit#gid=0

Let's go to voice if this isn't super clear, don't want to churn too much
on this PR

emord:
@calellowitz @amsagoff any update here?

amsagoff:
Cal, let me know if you need anything additional from me

calellowitz:
@millerdev @emord 
Had a conversation with @amsagoff and think we are going with this new spec. we are holding off on implementing tracking for autocomplete until it works with case properties too, and I couldn't find a way to determine the type of logic for drag and drop in the expression editor so we made that its own category.

src/core.js
```diff
@@ -755,11 +755,30 @@ define([
             }
 
             if (window.analytics) {
-                window.analytics.usage(
-                    "Question Reference",
-                    "Drag and Drop",
-                    _this.data.core.currentlyEditedProperty
```
emord: Is `currentlyEditedProperty` not set in the expression editor?

calellowitz: It might be, but the property isnt part of the new spec for what @dimagi/product wanted in ga. The new spec (in the google doc linked above) cares more about the target

src/core.js
```diff
@@ -755,11 +755,30 @@ define([
...
+                        targetType = "Validation";
+                        break;
+                    case 'property-calculateAttr':
+                        targetType = "Calculation";
+                        break;
+                    default:
+                        targetType = "Expression Editor";
+                        break;
+                }
+                if (path.contains('casedb')) {
```
millerdev: Looks like this will catch any path containing `casedb`, including something like `/data/not_in_my_casedb`. Seems like we might want to swap the branches here and do something like `path.startsWith("/data/")` to detect form references. Also should test this on reference to the current question, which uses a relative path of `.` (I can't remember if the relative path is passed in here or not).

src/javaRosa.js
```diff
@@ -1375,19 +1375,18 @@ define([
...
                     });
                     var e = window.event;
                     menu.css({'top': e.clientY, 'left': e.clientX}).show();
                 } else {
                     insertOutputRef(_this, target, path, mug);
-                    if (window.analytics) {
-                        window.analytics.usage("Output Value", "Drag and Drop");
+                }
+                if (window.analytics) {
+                    if (path.contains('casedb')) {
```
millerdev: same as above

src/core.js
millerdev: although the `/data/` might not always work after #616 is merged, but there might be a better way to do this check at that point. fyi @emord

calellowitz: if there is a larger set of complications to finding question ones (including possible breaking changes), how do you feel about using a fuller substring of the case path. something like `path.contains("instance('casedb')")` or `path.startsWith("instance('casedb')")` that will filter out most if not all false positives without having to deal with the different complexities of our question references.

millerdev: the problem with `instance('casedb')` is that both single and double quotes are allowed (not too hard to work around with a regexp) and that shorthand hashtag paths are coming (see the PR I linked above), which do not use that syntax or even contain the substring "casedb". Once hashtag paths become a thing I think we'll want utility functions for classifying paths rather than ad-hock pattern matching scattered throughout the codebase.

emord: This is purely for the databrowser drag/drop so I think instance('casedb') should be sufficient? `startswith('#case')` will probably be the best way after that pr, but I don't think this needs to worry about that

calellowitz: It looks like `mug` is undefined for case properties but it looks like not questions or answers. That could get around the issue of hashtags in paths and changing question paths. Is that a safe test to use or is there a special case where `mug` is not defined for non case properties.

millerdev: I wouldn't depend on it since it's not a precondition of this function that `mug` is defined for all dropped paths that are form properties.

For now I'd recommend implementing a function on the form that you can call like `_this.data.core.form.isFormReference(path)` and swap the branches in this `if`-block. In that function you can
```js
isFormReference: function (path) {
    return this.getMugByPath(path) ? true : false;
},
```
That way you can call the same function below in javaRosa.js and we can easily update it later if the criteria changes.

@emord wrote:
> This is purely for the databrowser drag/drop...

Isn't this function also called on drag/drop from the question tree?

calellowitz:
after further consideration, switched to looking for `instance('casedb')` because i believe drag and drop only uses single quotes and that is the same regex used at https://github.com/dimagi/Vellum/blob/master/src/richText.js#L383

calellowitz:
any idea whats going on with tests? still passing locally and failing here with seemingly unrelated output

src/form.js
```diff
@@ -959,6 +959,9 @@ define([
             var value = exporter.generateExportTSV(this);
             this.vellum.afterSerialize();
             return value;
+        },
+        isCaseReference: function (path) {
+            return /instance\('casedb'\)/.test(path) ? true : false;
```
millerdev: nitpicks: not sure if it matters in practice, but would be more robust to match ` /instance\(['"]casedb['"]\)/` in case double quotes were ever used. also ` ? true : false` is superfluous.

millerdev:
re: test failure, maybe try merging master?

millerdev:
nm. merging

src/core.js
emord: > Isn't this function also called on drag/drop from the question tree?
 
Yeah. I meant that checking for only single quotes was sufficient because the data browser will only drop single quotes. not double

PR #626, opened 2016-03-17
Fix accordion icons

orangejenny:
http://manage.dimagi.com/default.asp?221496


orangejenny:
Tests will pass once https://github.com/dimagi/MediaUploader/pull/16 is merged and bower is updated.

millerdev:
Merged dimagi/MediaUploader#16 and kicked the build.

PR #627, opened 2016-03-20
Add activity notification to keep session alive

millerdev:
http://manage.dimagi.com/default.asp?221617

@emord cc @proteusvacuum


czue:
sweet, thanks for doing this

PR #628, opened 2016-03-21
Undefined xmlns

emord:
http://manage.dimagi.com/default.asp?221525

Not adding new tests as it's already tested after changing to use `util.loadXML` and the check in the parser (tests fail between commits 2 and 3)

@millerdev @orangejenny 

cc: @biyeun

### DIFF
--- a/src/writer.js
+++ b/src/writer.js
@@ -67,7 +67,10 @@ define([
             jrm = "http://dev.commcarehq.org/jr/xforms";
         }
 
-        uuid = form.formUuid; //gets set at parse time/by UI
+        uuid = form.formUuid;
+        if(!uuid) {
+            uuid = "http://openrosa.org/formdesigner/" + util.generate_xmlns_uuid();
+        }
 
         xmlWriter.writeAttributeString("xmlns:jrm",jrm);
         xmlWriter.writeAttributeString("xmlns", uuid);
--- a/src/parser.js
+++ b/src/parser.js
@@ -151,7 +151,7 @@ define([
         form.formVersion = root.attr("version");
         form.formName = root.attr("name");
 
-        if (!form.formUuid) {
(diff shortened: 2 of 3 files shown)


src/writer.js
```diff
@@ -67,7 +67,10 @@ define([
             jrm = "http://dev.commcarehq.org/jr/xforms";
         }
 
-        uuid = form.formUuid; //gets set at parse time/by UI
```
emord: This comment was a lie

millerdev: I wonder when that comment became invalid? I'm pretty sure it used to be correct.

emord: it was mostly correct except when provided with an empty string. it goes back to major refactoring https://github.com/dimagi/Vellum/blame/master/src/parser.js#L40 so probably won't be fun trying to figure out when that happened

src/writer.js
```diff
@@ -67,7 +67,10 @@ define([
...
+        uuid = form.formUuid;
+        if(!uuid) {
+            uuid = "http://openrosa.org/formdesigner/" + util.generate_xmlns_uuid();
```
millerdev: Don't you need to write the value back to `form.formUuid`? Otherwise it will generate a new UUID each time the form is written. Might be worth making a test for that scenario.

src/writer.js
millerdev: oh man, one of those big PRs

src/writer.js
emord: Good point. Will do in separate PR. For this one I'm just bringing back old functionality https://github.com/dimagi/Vellum/commit/3bca6cb8d6c05ec25736ca3cad7d71311f9d8f69#diff-7dec243e7443d2aa55101241b38a8575L71

PR #629, opened 2016-03-22
HTML5 version of YUI uploader

orangejenny:
Drop flash-related code. Merge along with https://github.com/dimagi/MediaUploader/pull/17

@emord

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -36,7 +36,7 @@
         "jstree": "~3.2.1",
         "langcodes": "dimagi/langcodes",
         "XMLWriter": "dimagi/XMLWriter#master",
-        "MediaUploader": "dimagi/MediaUploader#retry-html5",
+        "MediaUploader": "dimagi/MediaUploader#master",
         "xpath": "dimagi/js-xpath#v0.0.2",
         "jsdiff": ">=1",
         "markdown-it": "~6.0.0"
--- a/bower.json
+++ b/bower.json
@@ -36,7 +36,7 @@
         "jstree": "~3.2.1",
         "langcodes": "dimagi/langcodes",
         "XMLWriter": "dimagi/XMLWriter#master",
-        "MediaUploader": "dimagi/MediaUploader#master",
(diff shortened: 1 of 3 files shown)


millerdev:
I see in [bower.js](https://github.com/dimagi/Vellum/pull/629/commits/5b9fd95ed36719798505fd30330a689cbb2eafcf#diff-0a08a7565aba4405282251491979bb6bR39) the we link to the `master` version of MediaUploader (I can't remember, but I may have even set it up that way). Thinking out loud here: would it make sense to use a commit hash ([e4c4ac7](https://github.com/dimagi/MediaUploader/pull/17/commits/e4c4ac7e54bb7e6b978c59edf0f2639f308d0a7f)) instead? Then we would have a record of exactly which version of MediaUploader is used by each version of Vellum.

emord:
@millerdev  That means we would have a different version for vellum than HQ almost every time we update MediaUploader

https://github.com/dimagi/MediaUploader/pull/16#issuecomment-199082818

millerdev:
lol, guess I'm looking at the issue from a different angle today, but I do seem to be contradicting what I said the other day. We could (should?) write a test for HQ that verifies that the versions of MediaUploader in the submodule and the embedded version of Vellum are the same.

orangejenny:
If bower is linked to a specific commit, updating MediaUploader will always require also updating Vellum (strictly speaking, maybe not always, but generally we're going to want media uploads to be the same in Vellum and HQ).

If bower is linked to master, updating MediaUploader in a way that requires Vellum changes becomes a small annoyance (making sure all changes are released together).

The first scenario happens more often than the second, so I'd rather keep linking to master.

emord:
I think the best solution is to make MediaUploader like jquery/underscore and let HQ serve the library and vellum uses HQ's version. I think that's out of scope for this though

millerdev:
> I think the best solution is to make MediaUploader like jquery/underscore and let HQ serve the library and vellum uses HQ's version. I think that's out of scope for this though

Yes, this! Thanks for articulating, @emord

PR #634, opened 2016-04-14
Inline video

emord:
http://manage.dimagi.com/default.asp?223608#1126669

image

It's a little weird that we use the itext type when displaying the types of media. never noticed until now. It'll take a bit of a refactor of how we use `SUPPORTED_MEDIA_TYPES` so it may be out of scope for this

@ctsims

### DIFF
--- a/src/javaRosa.js
+++ b/src/javaRosa.js
@@ -1160,7 +1160,7 @@ define([
         image: 'fa fa-photo',
         audio: 'fa fa-volume-up',
         video: 'fa fa-video-camera',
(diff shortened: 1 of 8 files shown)


emord:
Wasn't too sure about an icon. I'm using the play button now, but it's hard to use an icon to differentiate inline video vs video

millerdev:
Does this need QA, or is it ready to go live?

millerdev:
reviewed, good to merge

emord:
I think it's good to go

PR #635, opened 2016-04-18
Use <pre> for error message

emord:
Noticed this while in Moz. The old error wasn't useful at all, because the error message is already formatted. not sure when it was possible to have multiple error messages

Old:
selection_005

New:
selection_004

buddy: @proteusvacuum

### DIFF
--- a/src/templates/xpath_validation_errors.html
+++ b/src/templates/xpath_validation_errors.html
@@ -1,10 +1,7 @@
 <div class="alert alert-danger">
     <h4><i class="fa fa-warning"></i> Validation Failed</h4>
-    <p>Please fix all errors before leaving this page:</p>
-    <%=errors%>
-    <ul>
-    <% _.each(errors, function (error) { %>
-        <li><%=error%></li>
-    <% }); %>
-    </ul>
+    <p>Please fix the error before leaving this page:</p>
+    <pre>
+      <%=errors%>
+    </pre>
 </div>
--- a/src/expressionEditor.js
+++ b/src/expressionEditor.js
@@ -440,7 +440,7 @@ define([
                 } else {
                     getValidationSummary()
                         .html($(xpath_validation_errors({
-                            errors: results[1]
+                            errors: results[1].message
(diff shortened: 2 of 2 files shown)


src/templates/xpath_validation_errors.html
```diff
@@ -1,10 +1,7 @@
 <div class="alert alert-danger">
     <h4><i class="fa fa-warning"></i> Validation Failed</h4>
-    <p>Please fix all errors before leaving this page:</p>
-    <%=errors%>
-    <ul>
-    <% _.each(errors, function (error) { %>
-        <li><%=error%></li>
-    <% }); %>
-    </ul>
+    <p>Please fix the error before leaving this page:</p>
```
proteusvacuum: Should this be in a `{% trans %}` block?

emord: vellum doesn't currently have translations unfortunately. and I don't think we can rely on django to fix this because the templates are compiled into the vellum javascript code

proteusvacuum: :cry: :earth_africa:

src/templates/xpath_validation_errors.html
```diff
@@ -1,10 +1,7 @@
 <div class="alert alert-danger">
     <h4><i class="fa fa-warning"></i> Validation Failed</h4>
-    <p>Please fix all errors before leaving this page:</p>
-    <%=errors%>
-    <ul>
-    <% _.each(errors, function (error) { %>
-        <li><%=error%></li>
-    <% }); %>
```
millerdev: Did this code list the errors twice, or was this iteration for a different kind of error that's not printed by `<%=errors%>`?

emord: It listed them twice. You can see it in the screenshot I posted. I'm not actually sure when it was valid to have multiple errors. The parser will only return one

millerdev: :+1:

PR #636, opened 2016-04-25
Populate itemset element after insert

orangejenny:
http://manage.dimagi.com/default.asp?223314

Diff entire PR, not commit-by-commit.

### DIFF
--- a/src/itemset.js
+++ b/src/itemset.js
@@ -164,20 +164,19 @@ define([
 
     function afterDynamicSelectInsert(form, mug) {
         var sources = getDataSources(),
-            src = "",
-            nodeset = "",
-            choices = [];
+            newMug = form.createQuestion(mug, 'into', "Itemset", true);
         if (sources.length) {
-            src = sources[0].uri;
-            nodeset = "instance('" + sources[0].id + "')" + sources[0].path;
-            choices = datasourceWidgets.autocompleteChoices(sources, src);
+            var src = sources[0].uri,
+                nodeset = "instance('" + sources[0].id + "')" + sources[0].path,
+                choices = datasourceWidgets.autocompleteChoices(sources, src);
+            newMug = populateNodesetAttributes(newMug, choices);
+            newMug.p.filter = '';
+            mug.p.itemsetData = {
+                instance: form.parseInstance(
+                    nodeset, newMug, "itemsetData"),
+                nodeset: nodeset,
+            };
         }
-        var newMug = form.createQuestion(mug, 'into', "Itemset", true);
-        newMug = populateNodesetAttributes(newMug, choices);
-        newMug.p.filter = '';
-        newMug.p.itemsetData = {
-            nodeset: nodeset,
-        };
         return newMug;
     }
 
--- a/src/itemset.js
+++ b/src/itemset.js
@@ -44,6 +44,8 @@ define([
         writeControlLabel: false,
         writeControlRefAttr: null,
         writeCustomXML: function (xmlWriter, mug) {
+console.log("writeCustomXML");  // jls
+console.log(mug.__className + "'s mug.p.labelRef=" + mug.p.labelRef);
             var data = mug.p.itemsetData,
                 nodeset = data.nodeset,
                 filter = mug.p.filter,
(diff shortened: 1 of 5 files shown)


tests/copy-paste.js
```diff
@@ -685,13 +685,13 @@ define([
             paste([
                 ["id", "type", "labelItext:en-default", "itemsetData"],
                 ["/select", "SelectDynamic", "select",
-                 '[{"instance":null,"nodeset":"/items","labelRef":"@name","valueRef":"@id"}]'],
+                 '[{"instance":null,"nodeset":"/items","labelRef":"name","valueRef":"@id"}]'],
```
orangejenny: Not sure if this was a legitimate change to make. When I manually test copy & paste, it copies over the value & label correctly, but the test was failing to copy the labelRef correctly.

millerdev: this should probably not change, I suspect it's revealing a bug if this test is now failing. It may be a bug in test setup: may need to ensure data sources are loaded prior to running this test. My guess is that it's related to [`afterDynamicSelectInsert()`](https://github.com/dimagi/Vellum/blob/9a7b64b1e16d100b18c094fcc944471a13cfe7bb/src/itemset.js#L166) calling [`getDatasources()`](https://github.com/dimagi/Vellum/blob/9a7b64b1e16d100b18c094fcc944471a13cfe7bb/src/itemset.js#L359-L360), which returns asynchronously loaded data. It seems possible that the default value if the data sources are not yet loaded is `"name"`, and that's why `"@name"` is not pasted/copied correctly in this test (any value other than `"name"` will likely fail).

orangejenny: I reverted this change and dug into this some more, and I'm a bit confused about what this test should be doing. Stepping through this test, it looks like:
- `paste` calls `deserialize`
- In `deserialize`, because the given DynamicSelect mug doesn't have any children, `afterDynamicSelectInsert` is called
- `afterDynamicSelectInsert` creates a new itemset mug. The new mug doesn't have a label or value (is this the problem? should the new itemset be populated based off of the `{"instance":null,"nodeset":"/items","labelRef":"name","valueRef":"@id"}` supplied when the test calls `paste`?), so its value and label get set based on the first data source (happily, `getDataSources` does have the async data available when it's called).

So the @name/@id labelRef and valueRef are indeed being ignored, but I'm not sure which part of the code should be taking them into account. Thoughts?

I'm also a little confused because it doesn't look like vellum supports the exact use case of this test (copying a DynamicSelect without children). In the UI, adding a dynamic select automatically adds an itemset child that can't be deleted, so you can't copy/paste only the select.

emord: >  I'm not sure which part of the code should be taking them into account.

I think `afterDynamicSelectInsert` is what should be taking them into account. https://github.com/dimagi/Vellum/pull/636/files/9a7b64b1e16d100b18c094fcc944471a13cfe7bb#diff-f15dee15f58a804357f7fca0280de97dR374 does take it into account, but only if there are no data sources.

> copying a DynamicSelect without children

This is copying it with a child. The "child" in the form builder isn't actually a new question/mug, it just displays what's in `mug.p.itemsetData`

src/itemset.js
```diff
@@ -1,3 +1,16 @@
+/**
```
orangejenny: @millerdev This is my current understanding of itemsets, is it accurate?

orangejenny: Something I'd like to add but don't currently understand: why is the itemsetData attached to select questions an array, as in https://github.com/dimagi/Vellum/blob/master/tests/copy-paste.js#L688 ? Does vellum support multiple itemsets attached to a single question?

src/itemset.js
```diff
@@ -1,3 +1,16 @@
...
+ *  Its primary use is to populate data-driven select questions.
+ *
+ *  Itemsets contain
+ *      instance: which top-level data source contains the items
+ *      nodeset: path to apply to the instance to get at the desired items
+ *      valueRef: reference to apply to each node to get the item's value
+ *      labelRef: reference to apply to each node to get the item's display name
+ *
+ *  Dynamic select mugs typically have a child itemset mug. A dynamic select's
+ *  itsemset data may also be represented by an object stored in the mug's p.itemsetData.
```
millerdev: The dynamic variants of select widgets always have an itemset as a child mug (that's what makes them dynamic), and `itemsetMug.p.itemsetData` is where the itemset persistent state is stored.

src/itemset.js
millerdev: Yes, this is a pretty good summary of itemsets.

As for the array, I think it's because it was easiest to implement by iterating over the child mugs and serializing each of them; other parts of the code make sure there is only ever one. I don't think we have ever tested or formally supported a dynamic select question with multiple itemsets.

src/itemset.js
```diff
@@ -181,7 +213,7 @@ define([
             deserialize: function (data, key, mug) {
                 _.each(data[key], function (value, i) {
                     var children = mug.form.getChildren(mug),
-                        itemset = children[i] || afterDynamicSelectInsert(mug.form, mug),
+                        itemset = children[i] || afterDynamicSelectInsert(mug.form, mug, _.first(data.itemsetData)),
```
millerdev: This should use `value` instead of `_.first(data.itemsetData)`. However, I think the modification to `afterDynamicSelectInsert` (passing a third parameter) should not be necessary. Populating the `itemset` mug should be done by [line 219](https://github.com/dimagi/Vellum/pull/636/files/a4aec3fbb032e3265b241c0cc04a059664cf3194#diff-f15dee15f58a804357f7fca0280de97dR219) below. I'm guessing the bug is in `Itemset.spec.itemsetData.deserialize`.

orangejenny: > I'm guessing the bug is in `Itemset.spec.itemsetData.deserialize`.

Thank you - looks like there might have been an oversight in https://github.com/dimagi/Vellum/commit/e33600daf3eb3c1f417bc774b1ccbbbc20193fc9. Submitted a different fix.

PR #637, opened 2016-04-26
don't write #form to markdown

emord:
http://manage.dimagi.com/default.asp?225093

@orangejenny

### DIFF
--- a/tests/markdown.js
+++ b/tests/markdown.js
@@ -1,6 +1,7 @@
 define([
     'tests/options',
     'tests/utils',
+    'vellum/util',
     'chai',
     'jquery',
     'underscore',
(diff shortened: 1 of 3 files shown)


tests/markdown.js
```diff
@@ -10,10 +11,12 @@ define([
     'text!static/markdown/simple-markdown-no-chars.xml',
     'text!static/markdown/no-markdown.xml',
     'text!static/markdown/no-markdown-stars.xml',
-    'text!static/markdown/explicit-no-markdown.xml'
+    'text!static/markdown/explicit-no-markdown.xml',
+    'text!static/markdown/markdown-output-value.xml'
 ], function (
     options,
     util,
+    vellum_util,
```
millerdev: should be `vellumUtil`

emord: Looks like we use both conventions. tests/widgets uses vellumUtil but databrowser and javaRosa use vellum_util. Should `vellumUtil` be the canonical form?

millerdev: yes, to be consistent with the JS naming convention

millerdev: can rename them all in a separate PR if you want

orangejenny:
looks good, and @millerdev 's comment isn't blocking

PR #638, opened 2016-04-28
Advanced mode displaying #form even without flags on

orangejenny:
http://manage.dimagi.com/default.asp?225058

### DIFF
--- a/src/expressionEditor.js
+++ b/src/expressionEditor.js
@@ -355,8 +355,6 @@ define([
             } else {
                 showAdvancedMode(options.value);
             }
-
-            $div.find(".fd-xpath-editor-text").val(options.value);
         };
 
         // toggle simple/advanced mode


src/expressionEditor.js
```diff
@@ -355,8 +355,6 @@ define([
             } else {
                 showAdvancedMode(options.value);
             }
-
-            $div.find(".fd-xpath-editor-text").val(options.value);
```
orangejenny: It appears that this line is now redundant with `setExpression`.

millerdev:
I assume you tested locally. Writing tests for the expression editor is hard, but [it is possible](https://github.com/dimagi/Vellum/blob/3c2be45002270ef3f98fa156800911135ecb09cc/tests/widgets.js#L43-L64). Understandable if you looked at that and just gave up.

orangejenny:
I did test locally, but was admittedly lazy about writing a test for this. I'll take a look at adding one.

orangejenny:
@emord graciously wrote a test in https://github.com/dimagi/Vellum/pull/641, so we're good for test coverage here.

millerdev:
yay!

PR #639, opened 2016-04-28
Panel resize for undo delete

orangejenny:
http://manage.dimagi.com/default.asp?224963

@emord

### DIFF
--- a/src/undomanager.js
+++ b/src/undomanager.js
@@ -22,29 +22,36 @@ define([
         $('.fd-undo-container').append(UNDO_ALERT);
     }
 
-    function toggleAlert(undoStack) {
+    function toggleAlert(undoStack, vellum) {
         if (undoStack.length && !alertShown()) {
             createAlert();
         } else if (undoStack.length === 0 && alertShown()) {
-            $('.fd-undo-delete').alert('close');
+            $('.fd-undo-delete').remove();
+        }
+        if (vellum) {
+            vellum.adjustToWindow();
         }
     }
 
     function UndoManager() {
-        this.undoStack = [];
+        var _this = this;
+        _this.undoStack = [];
+        _this.vellum = undefined;
     }
 
     UndoManager.prototype = {
         resetUndo: function (mug, previousMug, position) {
             if (mug) {
                 this.undoStack = [[mug, previousMug, position]];
+                this.vellum = this.vellum || mug.form.vellum;
             } else {
                 this.undoStack = [];
             }
-            toggleAlert(this.undoStack);
+            toggleAlert(this.undoStack, this.vellum);
         },
         appendMug: function (mug, previousMug, position) {
             this.undoStack = this.undoStack.concat([[mug, previousMug, position]]);
+            this.vellum = this.vellum || mug.form.vellum;
             toggleAlert(this.undoStack);
         },
         undo: function () {
--- a/src/undomanager.js
+++ b/src/undomanager.js
@@ -2,10 +2,12 @@ define([
     'jquery',
     'underscore',
     'tpl!vellum/templates/undo_alert',
+    'vellum/window',
 ], function(
     $,
     _,
-    UNDO_ALERT
+    UNDO_ALERT,
+    windowManager
 ) {
     function alertShown() {
         var alert = $('.fd-undo-delete');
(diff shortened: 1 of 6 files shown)


orangejenny:
Don't actually have windows machine handy to test, but guessing this is due to the tree panel being too tall for the browser.

src/undomanager.js
```diff
@@ -46,6 +48,7 @@ define([
         appendMug: function (mug, previousMug, position) {
             this.undoStack = this.undoStack.concat([[mug, previousMug, position]]);
             toggleAlert(this.undoStack);
+            windowManager.adjustToWindow(mug.form.vellum);
```
millerdev: Another way to call this that would avoid the need to extract it to a module-level function is `mug.form.vellum.data.windowManager.adjustToWindow()`. Rather than (ab)using that long method path, I'd add a new core method `vellum.adjustToWindow()` that calls `this.data.windowManager.adjustToWindow()`.

And since `undomanager.appendMug()` is only called in one place, what do you think of pushing this call into [`form._addToUndoManager`](https://github.com/dimagi/Vellum/blob/6db614856293b222c5d3928d12a7a0429037524e/src/form.js#L888), like so:
```js
...
this.undomanager.appendMug(mug, previousSibling, position);
this.vellum.adjustToWindow()
...
```
The refactor mentioned at the top of this comment is still necessary, but that would eliminate some indirection.

orangejenny: yes, that sounds much better, will do

orangejenny:
> And since undomanager.appendMug() is only called in one place, what do you think of pushing this call into form._addToUndoManager

I did this and then undid it after adding logic to adjust the window when the alert is removed, not just when it's added - seemed cleaner to have one `adjustToWindow` call in undomanager.js than two calls in form.js. On the other hand, I don't love passing around a `vellum` in undomanager, I'm open to feedback on how things are organized now.

src/undomanager.js
```diff
@@ -22,29 +22,36 @@ define([
...
+        if (vellum) {
+            vellum.adjustToWindow();
         }
     }
 
     function UndoManager() {
-        this.undoStack = [];
+        var _this = this;
+        _this.undoStack = [];
+        _this.vellum = undefined;
```
emord: how about passing in the form instance to this constructor in src/form.js and then you could do `this.vellum = form.vellum` ?

orangejenny: yeah, i like that
https://github.com/dimagi/Vellum/pull/639/commits/13e30e30147eccd7ff6d6bd0a7e62e977f8bae3b

src/undomanager.js
```diff
@@ -22,16 +22,21 @@ define([
         $('.fd-undo-container').append(UNDO_ALERT);
     }
 
-    function toggleAlert(undoStack) {
+    function toggleAlert(undoStack, vellum) {
```
emord: Agreed that passing `vellum` through here does seem a little strange. Could change it to passing a generic `callback` function and have that function defined in `UndoManager`. Would be nice to also have this managed on a per form basis, but I don't think it's possible to do that with UI code like this

emord:
My comment is non blocking. This looks good to merge if you are happy with it

src/undomanager.js
millerdev: Another option would be to have `UndoManager` publish events (see [how](https://github.com/dimagi/Vellum/blob/deletion-panel-resize/src/form.js#L140) the form [does it](https://github.com/dimagi/Vellum/blob/deletion-panel-resize/src/form.js#L178-L181)), which would allow decoupling of UI code from `UndoManager`. The form could then hook up the relevant event(s) to call `vellum.adjustToWindow()` after creating the undo manager. Unfortunately the form is fairly tightly coupled with the UI at this point; ideally the form would not be interacting directly with the UI either since it is more of a model class.

If you want to dive deeper and do dependency injection, you could pass a fully configured and event-hooked undo manager into the form constructor, further decoupling the form from UI code. If you go this (DI) route, you might consider doing it in a separate PR. Currently the form instance is [created in parser.js](https://github.com/dimagi/Vellum/blob/deletion-panel-resize/src/parser.js#L35), which, while part of the vellum core, seems like a less than ideal place to be connecting model events with UI callbacks.

src/window.js
```diff
@@ -38,18 +38,18 @@ define([
...
-            this.adjustToWindow();
+            _this.adjustToWindow();
         },
-        adjustToWindow: function () {
+        adjustToWindow: function() {
             if (!this.$f.is(':visible')) {
                 return;
             }
-
+
```
millerdev: do you mind removing these whitespace-only modifications?

orangejenny: removed

src/undomanager.js
millerdev: Offlined with @orangejenny. She's going to work on the events suggestion in a separate PR.

PR #640, opened 2016-04-28
display #form in autocomplete

orangejenny:
http://manage.dimagi.com/default.asp?224965

@emord


millerdev:
Hmm, there are to no tests for `src/fuse.js`. How did that happen? We should add some.

emord:
@millerdev There weren't any direct tests for fuse, but it is tested indirectly through tests/atwho.js

millerdev:
Was going to suggest writing a test that replicates the bug fixed by this PR, but I guess it's pretty simple display logic so maybe not too important.

PR #641, opened 2016-04-28
normalize xpath for advanced expression editing

emord:
@orangejenny this should replace https://github.com/dimagi/Vellum/pull/638 I had fixed this the other day, but couldn't get wifi on my layover. I think the bug was double logged http://manage.dimagi.com/default.asp?225013#1133170 & http://manage.dimagi.com/default.asp?225058

### DIFF
--- a/src/expressionEditor.js
+++ b/src/expressionEditor.js
@@ -353,7 +353,11 @@ define([
                     $div.find('.fd-add-exp').click();
                 }
             } else {
-                showAdvancedMode(options.value);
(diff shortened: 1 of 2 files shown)


emord:
failing on phantomjs install. confirmed it passes locally except for jshint https://github.com/dimagi/Vellum/blob/hashtag-form-rerun/src/core.js#L74

orangejenny:
verified test passes locally

PR #644, opened 2016-04-29
support right to left text in labels

emord:
http://manage.dimagi.com/default.asp?224826#1136143

@NoahCarnahan says this looks right for hebrew.

### DIFF
--- a/src/javaRosa.js
+++ b/src/javaRosa.js
@@ -823,6 +823,8 @@ define([
             widgetClass = options.baseWidgetClass || widgets.richTextarea,
             widget, $input;
 
+        options.language = language;
+
         if (options.idSuffix) {
             id = id + options.idSuffix;
         }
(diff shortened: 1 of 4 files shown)


src/javaRosa.js
```diff
@@ -823,6 +823,8 @@ define([
             widgetClass = options.baseWidgetClass || widgets.richTextarea,
             widget, $input;
 
+        options.language = language;
```
millerdev: Seems like a strange place to be setting an option, given that both `language` and `options` are passed into this function. Is there a precedent for this?

emord: I think this is the closest use case: https://github.com/dimagi/Vellum/blob/master/src/javaRosa.js#L1152

I did it in this way so that it could be passed to the text and richText widgets, but I suppose it could have been done wherever this widget function was called

PR #646, opened 2016-05-02
fix uploader

emord:
http://manage.dimagi.com/default.asp?224561#assign_1_224561

Introduced here: https://github.com/dimagi/Vellum/pull/634

cc: @NoahCarnahan

### DIFF
--- a/src/uploader.js
+++ b/src/uploader.js
@@ -82,7 +82,7 @@ define([
         image: 'fd_hqimage',
         audio: 'fd_hqaudio',
         video: 'fd_hqvideo',
-        'video-inline': 'fd_hqvideo',
+        'video-inline': 'fd_hqInlineVideo',
         text:  'fd_hqtext',
     };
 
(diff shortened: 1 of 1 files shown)


src/uploader.js
```diff
@@ -82,7 +82,7 @@ define([
         image: 'fd_hqimage',
         audio: 'fd_hqaudio',
         video: 'fd_hqvideo',
-        'video-inline': 'fd_hqvideo',
+        'video-inline': 'fd_hqInlineVideo',
```
emord: Needs a different slug, or it opens two modals. one for video other for inline

src/uploader.js
```diff
@@ -295,10 +295,10 @@ define([
                         uploadUrl: uploadUrls.video,
                     }),
                     'video-inline': this.initUploadController({
-                        uploaderSlug: 'fd_hqvideo',
+                        uploaderSlug: 'fd_hqInlineVideo',
                         mediaType: 'video-inline',
                         sessionid: sessionid,
-                        uploadUrl: uploadUrls['video-inline'],
+                        uploadUrl: uploadUrls.video,
```
emord: Should use the same url as video as it's the same type.

PR #647, opened 2016-05-03
Fix lost changes

emord:
http://manage.dimagi.com/default.asp?217896#1093575
https://github.com/dimagi/Vellum/pull/617#issuecomment-186309022

this bug only shows up in the new editor. I've been sitting on this fix for a few days, because I haven't had a chance to bootstrap the expression editor tests that support rich text yet

@millerdev @gcapalbo

### DIFF
--- a/src/expressionEditor.js
+++ b/src/expressionEditor.js
@@ -233,6 +233,11 @@ define([
 
                 // add event handlers to validate the inputs
                 $expUI.find('.xpath-edit-node').on('keyup change', validateExpression);
+                if (options.mug.supportsRichText()) {
+                    $expUI.find('.xpath-edit-node').each(function () {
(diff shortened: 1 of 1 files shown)


orangejenny:
Curious what the bug was: the rich text editor was validating too frequently and this caused problems?

millerdev:
heh, I agree a test for this would be nice, but those tests are a mess.

emord:
> Curious what the bug was: the rich text editor was validating too frequently and this caused problems?

@orangejenny the bug isn't actually about validating the expresion, but `validateExpression` also sets a flag to warn the user when they have made changes to the expression in the expression editor, and that wasn't being set

orangejenny:
ah, makes sense, thanks

PR #648, opened 2016-05-03
External data panel always hidden

orangejenny:
@emord 

Introduced in https://github.com/dimagi/Vellum/pull/639

Would like to deploy this once it's merged. Chatted with @amsagoff and decided it isn't hotfix-worthy, but I'd rather not wait until tomorrow afternoon's deploy.


orangejenny:
tests passing now

PR #649, opened 2016-05-04
Events for UndoManager

orangejenny:
https://github.com/dimagi/Vellum/pull/639/files/13e30e30147eccd7ff6d6bd0a7e62e977f8bae3b#r61759913

@millerdev better?

Debating firing an event for when a question removal is undone, but no one would be listening for it, so it seemed like overkill.

### DIFF
--- a/src/undomanager.js
+++ b/src/undomanager.js
@@ -2,11 +2,13 @@ define([
     'jquery',
     'underscore',
     'tpl!vellum/templates/undo_alert',
+    'vellum/util',
 ], function(
     $,
     _,
-    UNDO_ALERT
-) {
+    UNDO_ALERT,
+    util,
+undefined) {
     function alertShown() {
         var alert = $('.fd-undo-delete');
         // creating alert uses classes "fade in", removing alert removes in
@@ -22,21 +24,20 @@ define([
         $('.fd-undo-container').append(UNDO_ALERT);
     }
 
-    function toggleAlert(undoStack, vellum) {
+    function toggleAlert(undoStack) {
         if (undoStack.length && !alertShown()) {
             createAlert();
         } else if (undoStack.length === 0 && alertShown()) {
             $('.fd-undo-delete').remove();
-        }
-        if (vellum) {
-            vellum.adjustToWindow();
         }
     }
 
     function UndoManager(form) {
         var _this = this;
         _this.undoStack = [];
-        _this.vellum = form.vellum;
+        _this.form = form;
+
+        util.eventuality(this);
     }
 
     UndoManager.prototype = {
@@ -46,13 +47,17 @@ define([
             } else {
                 this.undoStack = [];
             }
-            toggleAlert(this.undoStack, this.vellum);
+            toggleAlert(this.undoStack);
+            this.form.fire({
+                type: 'undo-reset',
+            });
         },
         appendMug: function (mug, previousMug, position) {
             this.undoStack = this.undoStack.concat([[mug, previousMug, position]]);
             toggleAlert(this.undoStack);
         },
         undo: function () {
+            var _this = this;
             _.each(this.undoStack, function(undo) {
                 var mug = undo[0],
                     sibling = undo[1],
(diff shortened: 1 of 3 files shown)


src/undomanager.js
```diff
@@ -2,11 +2,13 @@ define([
...
     'tpl!vellum/templates/undo_alert',
+    'vellum/util',
 ], function(
     $,
     _,
-    UNDO_ALERT
-) {
+    UNDO_ALERT,
+    util,
+undefined) {
```
emord: Vellum convention is to have templates at the end. Also I don't think that you need the `undefined` here.

src/undomanager.js
```diff
@@ -46,13 +47,17 @@ define([
...
+            this.form.fire({
+                type: 'undo-reset',
+            });
         },
         appendMug: function (mug, previousMug, position) {
             this.undoStack = this.undoStack.concat([[mug, previousMug, position]]);
             toggleAlert(this.undoStack);
         },
         undo: function () {
+            var _this = this;
```
emord: I think this is unneeded

emord:
Looks good to me. Agreed on not firing an event

src/undomanager.js
orangejenny: yes, fixed

src/undomanager.js
orangejenny: Re-ordered modules. The undefined allows for a trailing comma.

src/undomanager.js
```diff
@@ -1,12 +1,14 @@
...
+    'vellum/util',
     'tpl!vellum/templates/undo_alert',
 ], function(
     $,
     _,
-    UNDO_ALERT
-) {
+    util,
+    UNDO_ALERT,
+undefined) {
```
millerdev: I understand and agree with your appreciation for trailing commas, but I think this is aesthetically...not very pleasant. I'm -1 on this as a convention. Also, I think in modern browsers it won't, but might this possibly define `undefined` if too many arguments are passed to the function?

src/undomanager.js
```diff
@@ -22,21 +24,20 @@ define([
...
-        }
     }
 
     function UndoManager(form) {
         var _this = this;
         _this.undoStack = [];
-        _this.vellum = form.vellum;
+        _this.form = form;
+
+        util.eventuality(this);
```
millerdev: This is not needed since a form event is fired below. Although...

It seems like a violation of encapsulation for undo manager to fire a form event. On the other hand, it has the advantage of decoupling the form from the UI. I think this is a fine, if not the purest, solution.

FWIW here's what I originally had in mind:

form.js
```js
function Form (opts, vellum, mugTypes) {
    ...
    //make the object event aware
    util.eventuality(this);

    this.undomanager = new UndoManager();
    this.undomanager.on('undo-reset', vellum.adjustToWindow);  // this is simple, but...
    /* to really decouple the form from the UI (and use the vellum var as little as possible) we could instead
    // re-fire as a form event to be handled in the core
    this.undomanager.on('undo-reset', this.fire.bind(this));
    */
}
```
undomanager.js
```js
function UndoManager() {
    ...
    util.eventuality(this);
}
UndoManager.prototype = {
    resetUndo: function (...) {
        ...
        this.fire({type: 'undo-reset'});
    },
    ...
}
```

src/undomanager.js
orangejenny: I don't think there's a good way to compromise on aesthetics, so I'll defer to the current vellum conventions & drop it ;)

src/undomanager.js
orangejenny: I see what you mean. My thinking was basically that undoing a question removal is fairly similar to adding a question, which is handled by the form. But an undo manager could (maybe) handle events that aren't so form-related, and the actual `undo-reset` event is about the undo queue, not about the form. I think I'll rearrange this code.

orangejenny: (updated PR)

PR #650, opened 2016-05-05
instances for itext

orangejenny:
http://manage.dimagi.com/default.asp?222569

Will write a test, but first interested in knowing if this approach is acceptable.

### DIFF
--- a/src/javaRosa.js
+++ b/src/javaRosa.js
@@ -1814,10 +1814,6 @@ define([
                             }
                             writeValue(xmlWriter, val);
                             xmlWriter.writeEndElement();
-                            if (val.search(/#case\//) !== -1) {
-                                form_.addInstanceIfNotExists({ id: 'commcaresession' });
-                                form_.addInstanceIfNotExists({ id: 'casedb' });
-                            }
                         }
                         if (item.hasMarkdown && !this.data.core.form.noMarkdown) {
                             val = item.get('default', lang);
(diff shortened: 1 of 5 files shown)


emord:
I think this could work for the short term since it's kind of an edge case, but would be good to get itexts into the logic manager. It's also going to be painful changing over all of those xml file's instance locations.

I did a quick POC for getting output values into the logic manager
https://github.com/dimagi/Vellum/compare/hashtag-form-rerun...quick-ref-output

orangejenny:
> It's also going to be painful changing over all of those xml file's instance locations.

Hrm, I hadn't noticed that the instances get moved around. Which is probably not in line with the XForm spec.

src/javaRosa.js
```diff
@@ -1814,6 +1814,10 @@ define([
                             }
                             writeValue(xmlWriter, val);
                             xmlWriter.writeEndElement();
+                            if (val.search(/#case\//) !== -1) {
+                                form_.addInstanceIfNotExists({ id: 'commcaresession' });
+                                form_.addInstanceIfNotExists({ id: 'casedb' });
+                            }
```
millerdev: What if, instead of on write XML, we did something similar on itext change? The down-side is that simply loading and saving a form will not "fix" (add instances for) case references in itext. I'm not excited about moving the instances after the `<itext>` node either, but my excitement level aside, we should check if that's even supported if we really think that's going to happen.

millerdev:
> could work for the short term since it's kind of an edge case, but would be good to get itexts into the logic manager

my sentiments exactly

orangejenny:
superseded by https://github.com/dimagi/Vellum/pull/652

PR #651, opened 2016-05-05
Refresh visible data

emord:
http://manage.dimagi.com/default.asp?226125#1139553

@orangejenny 

cc: @gcapalbo

### DIFF
--- a/tests/databrowser.js
+++ b/tests/databrowser.js
@@ -307,9 +307,48 @@ define([
                 });
             });
 
+            describe("with loaded xml", function () {
+                beforeEach(function (done) {
+                    util.init({
+                        plugins: plugins,
+                        javaRosa: {langs: ['en']},
+                        core: {
+                            dataSourcesEndpoint: function (callback) {
+                                event.on("nodeError", function() {
+                                    callback(CASE_DATA);
+                                });
+                            },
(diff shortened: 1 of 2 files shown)


tests/databrowser.js
```diff
@@ -307,25 +309,55 @@ define([
...
+                                widget.input.promise.then(done);
+                            }
+                        }
+                    });
+                });
+
+                it("should not error for known properties", function() {
+                    assert.strictEqual(widget.getValue(), escapedDobProp);
+                    assert.lengthOf(widget.getControl().find('.label-datanode-unknown'), 0);
+                    event.fire("nodeError");
```
orangejenny: What are the two asserts after this verifying?

emord: Just that nothing has changed after data browser/external properties have been loaded. Probably not needed, but doesn't hurt either

orangejenny: picky, and i see you didn't invent this in this PR, but `nodeError` is a weird name if the purpose of the event is to force the external properties to load

emord: huh. I agree with you. I have no recollection of inventing that, but git doesn't lie https://github.com/dimagi/Vellum/commit/5b40c4c993df98588946a986af8996326b9ff97a will change it

PR #652, opened 2016-05-06
Quick ref output

emord:
@orangejenny  I think this could supersede https://github.com/dimagi/Vellum/pull/650 I figured out what was causing my tests to fail

I'm not 100% satisfied with this. Also thought about splitting `addReferences` out into something that handles javaRosa and one that handles xpath, but that also didn't seem like the best. 

@millerdev @gcapalbo


src/logic.js
```diff
@@ -265,6 +264,21 @@ define([
             });
             return messages;
         },
+        addReferences: function (mug, property, value) {
+            // get absolute paths from mug property's value
+            var isLabel = false,
+                _this = this;
+            if (!value && mug.p[property] && _.isFunction(mug.p[property].get)) {
```
orangejenny: is checking for a `get` function really the best/only way to determine a property is a label? seems roundabout/fragile.

src/logic.js
```diff
@@ -265,6 +264,21 @@ define([
             });
             return messages;
         },
+        addReferences: function (mug, property, value) {
+            // get absolute paths from mug property's value
+            var isLabel = false,
```
millerdev: Would this work? `var isLabel = (!value && mug.p[property] instanceof jr.ItextItem),`

millerdev: Better idea: add a function to the `ItextItem` type: `ItextItem.mapExpressions(func)`, which would encapsulate `$('<div>').append(mug.p[property].get()).find('output').map(func)`. This would be cleaner since `ItextItem` knows about its implementation details like multiple languages, output tags, etc. and those things need to be hard-coded into the logic manager. This design would allow any mug property to use this protocol to embed multiple expressions in an opaque property value.

If you do that, this line becomes
```js
var isLabel = (!value && mug.p[property] && _.isFunction(mug.p[property].mapExpressions)),
```

millerdev: update: fixed wording of above comment

emord: That's a sweet refactor https://github.com/dimagi/Vellum/pull/652/commits/455205c500a181168ccf2c672499a82dde0c72de

src/javaRosa.js
```diff
@@ -171,7 +171,14 @@ define([
             return Boolean(this.hasForm('default') || 
                            this.hasForm('long')    || 
                            this.hasForm('short'));
-        }
+        },
+        forEachLogicExpression: function (fn) {
+            var text = this.get(),
```
millerdev: This gets the value for the default language. What do you think about iterating over all languages?

millerdev: Might want to add a test with an expression in non-default language, but not in the default.

emord: https://github.com/dimagi/Vellum/pull/652/commits/cbbe9298271ec31077ddb27c9456d6ae405ce123

src/javaRosa/itext.js
```diff
@@ -135,7 +135,18 @@ define([
             return Boolean(this.hasForm('default') || 
                            this.hasForm('long')    || 
                            this.hasForm('short'));
-        }
+        },
+        forEachLogicExpression: function (fn) {
+            var forms = this.getForms(),
+                ret = _.map(forms, function(form) {
+                    return _.map(form.getOutputRefExpressions(), function (exprs, lang) {
```
millerdev: ha! `form.getOutputRefExpressions()` is perfect for this

tests/databrowser.js
```diff
@@ -200,6 +204,77 @@ define([
...
+                    });
+                    assert.equal(getInstanceId(mug.form, sessionUri), null);
+                    assert.equal(getInstanceId(mug.form, casedbUri), null);
+                    assert.equal(label.length, 1);
+                    util.findNode(dataTree, "dob").data.handleDrop(label);
+                });
+            });
+
+            describe("with two languages", function () {
+                beforeEach(function (done) {
```
millerdev: non-blocking: curious why you recently started using `beforeEach` instead of `before`? I know it resets more state to re-init vellum for every test, but it also takes more time. My previous best practice was to use `before` in cases where most vellum state was unlikely to be changed by the tests in the group, and then `util.loadXML("");` (which is fast) to load a fresh form for each test. Should I be thinking about things differently?

millerdev:
👍 on ✅

tests/databrowser.js
emord: not sure. it's possible I was using it to reduce some boilerplate in some tests and just got into the habit of using it. I quickly went through and changed all of the tests that don't require beforeEach https://github.com/dimagi/Vellum/pull/652/commits/a37daef0f0aadef713d6da634515b45fd1ec96b0

Other tests depend on it in some way (though some of them look easy to change to use util.loadxml("");)

millerdev: 👍

PR #653, opened 2016-05-06
Split javarosa

orangejenny:
Break javarosa into more manageable pieces:
```
src/javarosa.js (plugin only)
src/javarosa/
   itext.js
      ItextForm
      ItextItem
      ItextModel
   itextBlock.js
      itextConfigurableBlock
      itextLabelBlock
      itextMediaBlock
   itextWidget.js
      itextFormWidget
      iTextIDWidget
      itextLabelWidget
      itextMarkdownWidget
      itextMediaWidget
   util.js (a few of these could are only used in the plugin and could have stayed in javaRosa)
      ITEXT_PROPERTIES
      SUPPORTED_MEDIA_TYPES
      forEachItextItem
      generateItextXLS
      getDefaultItextRoot
      getDefaultItextId
      getItextItemsFromMugs
      getOutputRef
      insertOutputRef
      looksLikeMarkdown
      parseXLSItext
      warnOnCircularReference
      warnOnNonOutputableValue
```


src/javaRosa/itextWidget.js
```diff
@@ -0,0 +1,478 @@
...
+    'vellum/widgets',
+    'vellum/util',
+    'vellum/atwho',
+    'vellum/core'
+], function (
+    _,
+    $,
+    auto_box,
+    markdown_help,
+    jr_util,
```
millerdev: should use JS naming convention (`jrUtil`)

src/javaRosa/util.js
```diff
@@ -0,0 +1,277 @@
...
+        generateItextXLS: generateItextXLS,
+        getDefaultItextRoot: getDefaultItextRoot,
+        getDefaultItextId: getDefaultItextId,
+        getItextItemsFromMugs: getItextItemsFromMugs,
+        getOutputRef: getOutputRef,
+        insertOutputRef: insertOutputRef,
+        looksLikeMarkdown: looksLikeMarkdown,
+        parseXLSItext: parseXLSItext,
+        warnOnCircularReference: warnOnCircularReference,
+        warnOnNonOutputableValue: warnOnNonOutputableValue,
```
millerdev: `warnOnCircularReference` and `warnOnNonOutputableValue` appear to be only referenced by `insertOutputRef`, so they don't need to be exported in the module object.

src/javaRosa/itextWidget.js
millerdev: also noticed the same in javaRosa.js, but the diff is not available for comment on github

millerdev:
What do you think about moving `src/javaRosa.js` to `src/javaRosa/plugin.js`?

This is more food for thought/discussion than a well-reasoned recommendation. I'm on the fence, but wonder if it might be more intuitive for a new person reading the code, especially since `.js` extensions are excluded from module names in require.js module requirement lists.

orangejenny:
> What do you think about moving `src/javaRosa.js` to `src/javaRosa/plugin.js`?

I like calling it `plugin` and like having the plugin as the same directory as the rest of javaRosa, but would also like the plugin file to be visibly special in some way since it's the entry point. What would you think of `src/javaRosa/javaRosa.js`? Although the double `javaRosa` isn't exactly clear.

millerdev:
I think `src/javaRosa/plugin` would only be referenced in [one place](https://github.com/dimagi/Vellum/blob/split-javarosa/src/main.js#L167), so my vote is for `plugin`. I agree that "double `javaRosa` isn't exactly clear." In the end it will be up to you since you are the author :)

orangejenny:
> I think src/javaRosa/plugin would only be referenced in one place, so my vote is for plugin

That's a good point. Renamed.

PR #654, opened 2016-05-09
add mug property reference

orangejenny:
I've been dithering over making a wiki vs adding to the README vs adding more inline comments, decided a file-level comment was a reasonable compromise.


src/mugs.js
```diff
@@ -1,3 +1,37 @@
+/**
+ * A mug is a question, containing data, bind, and control elements.
+ *
+ * Main Properties
+ *  nodeID
+ *  label
+ *  readOnlyControl
+ *  itemsetData
```
millerdev: I think this one is specific to the itemsets plugin, so probably should not be listed here.

src/mugs.js
```diff
@@ -1,3 +1,37 @@
+/**
+ * A mug is a question, containing data, bind, and control elements.
+ *
+ * Main Properties
+ *  nodeID
+ *  label
```
millerdev: Might want to add [these](https://github.com/dimagi/Vellum/blob/1691e9bba92aebce08e8a28803bff82657003ce0/src/javaRosa.js#L348-L351) too since `mediaItext` is listed below. Technically they're part of the javaRosa plugin, but that plugin is so tightly coupled with the rest of vellum that I think it makes sense to include them here.

src/mugs.js
```diff
@@ -1,3 +1,37 @@
...
+ *  calculateAttr: Hidden questions only. The calculation that generates the question's value.
+ *  requiredAttr: Boolean. Whether or not user must enter a value for the question.
+ *  relevantAttr: Boolean expression that determines whether or not to display a question to the end user.
+ *  constraintAttr: Boolean expression that, if false, will prevent the user from proceeding past the question.
+ *  repeat_count: Repeat groups only. An integer expression that determines the number of groups to generate.
+ *
+ * Advanced Properties
+ *  dataSource
+ *  dataValue
+ *  defaultValue: An expression that will be assigned to a question before/until a user changes the value.
```
emord: Can you include the distinction between `dataValue` and `defaultValue` ? One is inserted into the data instance (I think `dataValue`, the other is a setValue (I think `defaultValue`). Also the one that inserts into the data instance deprecated

src/mugs.js
orangejenny: removed

src/mugs.js
orangejenny: Oh, I'd meant to add a file-level comment to `src/javaRosa/plugin.js` once the refactor was done, thanks for the reminder. Did that, and added the itext properties there.

src/mugs.js
emord: On second thought I think that's sufficient. using a setvalue is more of an implementation detail.

src/mugs.js
emord: > Technically they're part of the javaRosa plugin, but that plugin is so tightly coupled with the rest of vellum that I think it makes sense to include them here

side note: Should we stop assuming that someone could use vellum without javaRosa? We could get rid of label and hintLabel at least, maybe some other things?

emord:
:+1: from me. will let @millerdev comment on your last comments.

src/mugs.js
orangejenny: > Should we stop assuming that someone could use vellum without javaRosa?

I'd be in favor of that.

millerdev: I think that makes sense for practical reasons, but I'm a little torn.

I feel like we've been moving in a direction where we're defining better interfaces and protocols around mug properties and various plugin extension points. Labels (itext vs default) are one place where I feel like the interface is not quite there yet, but it's close. If we stop trying to keep javaRosa separate it may become even more tightly coupled with the rest of vellum, which is not ideal because it's a big chunk of code to digest, and having that be part of the core means there is that much more that needs to be understood to get the gist of how vellum works.

On the other hand, the `label` and `hintLabel` properties are not well tested, and I'm not even sure it's possible to run vellum with the javaRosa plugin disabled to fully test them. While that seems like an interesting goal, if for no other reason than having smaller XML snippets (no itext, etc.) for testing features that are orthogonal to javaRosa, I'm not sure it's practical.

Unless I've convinced either of you to come to my side (keep javaRosa as a plugin, separate from vellum core), I think I've been outvoted.

millerdev:
This is great! Thanks for adding these.

src/mugs.js
emord: I 100% agree javarosa should be kept as a plugin. I think the benefits of having it in a contained area outweigh anything else.

What I meant by my original comment was that things like `label` and `hintLabel` could be removed from src/mugs.js, and I assumed that there are probably some other things that could be removed as well, although I haven't checked.

That is unless we can see a future where vellum can be run without javarosa where a label would be needed

millerdev: I wonder how many forms exist that use `label` and `hintLabel`? (and actually need them)

I can't imagine a future where we're not using javaRosa and would want the code that implements `label` and `hintLabel`. I think the main reasons for keeping them is theoretical backward compatibility and plugin system purity.

emord: Ah. That sounds reasonable. Also looks like we do have some tests for `label` and `hintLabel` (mostly just `label` though)

PR #655, opened 2016-05-09
Parse/write jr:count hashtags

emord:
http://manage.dimagi.com/default.asp?226243#1141660

@orangejenny 

cc: @gcapalbo

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -306,6 +306,7 @@ define([
         }
 
         var form = mug.form,
+            vellumKey = key.replace(':', '__'),
             xpath_, hashtag;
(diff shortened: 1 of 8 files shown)


src/util.js
```diff
@@ -306,14 +306,15 @@ define([
         }
 
         var form = mug.form,
+            vellumKey = key.replace(':', '__'),
```
emord: Would be up for changing this. Not that concerned since we do control the vellum namespace, but probably not the most ideal situation

millerdev:
👍 on ✅

PR #657, opened 2016-05-11
remove appearance attributes for all questions that explicitly set it

emord:
http://manage.dimagi.com/default.asp?226528

`delete mug.p.apperance` doesn't work because `mug.p` is not an Object it is `MugProperties` instance.

cc @gcapalbo

### DIFF
--- a/tests/questionTypes.js
+++ b/tests/questionTypes.js
@@ -451,7 +451,7 @@ define([
 
             it("should remove appearance attribute from label", function () {
                 var from = "Trigger",
-                    to = "Text",
+                    to = "MSelect",
                     mug = setup(from, to);
                 call("changeMugType", mug, to);
                 mug = util.getMug(mug.p.nodeID);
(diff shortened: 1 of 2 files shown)


tests/questionTypes.js
```diff
@@ -451,7 +451,7 @@ define([
 
             it("should remove appearance attribute from label", function () {
                 var from = "Trigger",
-                    to = "Text",
+                    to = "MSelect",
```
millerdev: Curious why the tests needed to be changed? I assume they no longer passed, but I'm not sure why.

emord: Trigger -> MSelect wouldn't have passed before this change (that was the bug) and this tests what I thought it would test when I originally wrote it. I can change it to be a more generalized form of test to test more scenarios though

millerdev: Cool. More tests is usually a good thing. Up to you if you want to add them.

PR #658, opened 2016-05-11
remove dead comment

orangejenny:
doesn't seem that this is true anymore

@emord

### DIFF
--- a/src/javaRosa/itextBlock.js
+++ b/src/javaRosa/itextBlock.js
@@ -314,10 +314,6 @@ define([
             return _.intersection(block.activeForms, block.forms);
         };
 
-        // this used to be the form ID instead of 'data'.  Since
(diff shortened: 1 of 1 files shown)


src/javaRosa/itextBlock.js
```diff
@@ -314,10 +314,6 @@ define([
             return _.intersection(block.activeForms, block.forms);
         };
 
-        // this used to be the form ID instead of 'data'.  Since
-        // only hand-made forms will ever end up with a different
-        // ID (the ability to set it in the UI has been broken for
-        // a while), it seemed ok to make it just 'data'
```
emord: I think this is still true for the default media type `mug.form.getBasePath()` will return `/data/` in most forms, but I don't think this comment means much as it will probably never change back to form id (this has been around since "major refactoring")

PR #659, opened 2016-05-11
Document new text widget and reduce dependency on form

emord:
@orangejenny here's some more docs about the rich text widget.

It also decouples vellum's form implementation from the widget.

cc: @gcapalbo

### DIFF
--- a/src/richText.js
+++ b/src/richText.js
@@ -600,7 +600,7 @@ define([
             isValidHashtag: _.bind(form.isValidHashtag, form),
             getIconFromPath: function (path) {
                 var mug = form.getMugByPath(path);
-                return mug ? mug.options.icon : null;
+                return mug ? mug.getIcon() : null;
             },
             xpath: form.xpath,
         };
--- a/src/richText.js
+++ b/src/richText.js
@@ -145,8 +145,15 @@ define([
      * if the editor does not exist.
      *
      * @param input - editor jQuery HTML element.
-     * @param form - form object; used for creating bubbles.
-     * @param options - rich text bubble options.
+     * @param form - form object, functions used:
+     *    normalizeEscapedHashtag - transforms to escaped hashtag form
+     *    transform - transforms escaped hashtags
+     *    isValidHashtag - boolean if hashtag translation exists
+     *    getMugByPath - only used for mug.options.icon
+     *    xpath - xpath parser
+     * @param options -
+     *    rtl - use right to left text
+     *    isExpression - treat input as xpath or itext
      */
     var editor = function(input, form, options) {
         var wrapper = input.data("ckwrapper");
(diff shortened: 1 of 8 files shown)


src/richText.js
```diff
@@ -145,8 +113,17 @@ define([
      * if the editor does not exist.
      *
      * @param input - editor jQuery HTML element.
-     * @param form - form object; used for creating bubbles.
-     * @param options - rich text bubble options.
+     * @param form - form object, functions used:
```
emord: This could have a better name. It's really just a collection of required methods to pass in to the editor. All of them except `getIconFromPath` could have sensible defaults derived from a js-xpath instance

src/richText.js
```diff
@@ -588,10 +570,85 @@ define([
         };
     }
 
+    function createFormModelForEditor(form) {
```
emord: This is the connection between vellum form and form expected by ckeditor

tests/utils.js
```diff
@@ -173,6 +173,10 @@ define([
         if (opts.plugins) {
             vellum_options.plugins = opts.plugins;
         }
+        if (vellum_options && vellum_options.features &&
+            _.isUndefined(vellum_options.features.disable_popovers)) {
+            vellum_options.features.disable_popovers = true;
```
emord: This disables the bubble popovers unless a test explicitly wants them (only a couple of tests at the moment). Anecdotally  I had a test timing out before this (2 s) that went down to < .5s

millerdev: 👍

src/richText.js
millerdev: Agree, this either needs a different name, or it should be a real form instance. It seems like a recipe for confusion/bugs to be passing around things that look like forms but have different methods.

src/richText.js
millerdev: Why not add a `getIconFromPath` method to the real form object and eliminate this adaptor? If you think `getIconFromPath` is not useful in other contexts, then I would recommend making it a private utility function in this module with the interface `getIconFromPath(path, form)`.

src/richText.js
emord: This is now the form. One of the original goals for this was to totally remove refs of form from the widget so it was totally independent of vellum. I think it should be fairly trivial to do that now that it's fully documented though.

src/richText.js
emord: not sure why I originally decided to not add to form, but it is added now. https://github.com/dimagi/Vellum/pull/659/commits/bd8b70bea3c8c0b278ad3a49375f6e06e802dca1 & https://github.com/dimagi/Vellum/pull/659/commits/02e00b0f7cf5ae291d97cc366f616fdebd26a270

PR #663, opened 2016-05-13
Select question via url

emord:
I wrote support for this a year ago, but there was something in HQ that caused it not to work. I tried it out yesterday and it looks like it works now (assuming thanks to b3 migration), although we need a double # in the url. So to select question1 you could use a URL like commcarehq.org/a/..../source/##form/question1

There was some fighting with tests in here, where every other test would fail. I never quite got down to what was happening (I think jstree might not be initialized properly when using `util.loadXML`), but I was able to work around it

cc: @gcapalbo


gcapalbo:
(looks like build failed)

gcapalbo:
pinging @millerdev in case he wants to review as well

src/core.js
```diff
@@ -1159,7 +1160,7 @@ define([
     };
 
     fn.loadXML = function (formXML, options) {
-        var form, _this = this;
+        var form, _this = this, qId = window.location.hash.substr(1);
```
millerdev: What do you think of dropping the extra `#` from the URL? So it would become commcarehq.org/a/..../source/#form/question1 Then if `qId` is a non-empty string and it starts with `"form/"` then prepend `#` and load that question else ignore `qId`.

orangejenny:
This is a cool idea, and code looks reasonable. Any reason for using `commcarehq.org/a/..../source/##form/group1/question1` rather than `commcarehq.org/a/..../source/#group1/question1`? If there's a future world where `commcarehq.org/a/..../source/##case/property1` is meaningful, having the `#form` is worthwhile...but it does look weird.

orangejenny:
Displays `#null` when a choice is selected

screen shot 2016-05-16 at 9 21 12 am

src/core.js
```diff
@@ -1288,18 +1289,33 @@ define([
...
+        if (qId) {
+            var mug = _this.getMugByPath(qId);
+            if (mug) {
+                _this.selectSomethingOrHideProperties(true, mug.ufid);
+            } else {
+                _this.selectSomethingOrHideProperties(true);
+            }
+        } else {
+            _this.selectSomethingOrHideProperties(true);
+        }
```
millerdev: I think this block can be simplified to
```js
var mug = qId && _this.getMugByPath(qId) || undefined;
_this.selectSomethingOrHideProperties(true, mug && mug.ufid);
```

src/core.js
```diff
@@ -1288,18 +1289,33 @@ define([
...
         if (forceDeselect) {
             this.jstree('deselect_all');
         }
         // ensure something is selected if possible
         if (!this.jstree('get_selected').length) {
             // if there's any nodes in the tree, just select the first
-            var all_nodes = this.data.core.$tree.find("li");
-            if (all_nodes.length > 0) {
+            var all_nodes = this.data.core.$tree.find("li"),
+                selected = all_nodes.filter('[id= ' + ufid + ']');
```
millerdev: Does it matter if `ufid` is null or undefined? Seems like it might be better to skip the filter if `ufid` evaluates to false.

src/core.js
```diff
@@ -2082,6 +2098,7 @@ define([
 
     fn.handleMugRename = function (form, mug, newId, oldId, newPath, oldPath, oldParent) {
         form.handleMugRename(mug, newId, oldId, newPath, oldPath, oldParent);
+        window.location.hash = '#' + mug.hashtagPath;
```
millerdev: Does `mug.hashtagPath` not start with `#`? At any rate, see suggestion above about dropping the extra `#`.

millerdev:
build fail
```
  1) The copy-paste plugin with instances without src should properly copy:
     AssertionError: cut or copy mismatch 
===================================================================
--- 	actual
+++ 	expected
@@ -1,5 +1,5 @@
 Form Builder clip	version 1
 id	type	labelItext:en-default	appearance	calculateAttr	instances
 /score	Int	What was your score	null	null	null
 /output	DataBindOnly	null	null	instance('scores')/score[@high > #form/score][@low < #form/score]	"{""scores"":{""children"":""<score low=\""0.0\"" high=\""500.0\"">You're really bad</score><score low=\""500.0\"" high=\""99999999.0\"">You're really good</score>""}}"
-/result	Trigger	"<output value=""#form/output"" />"	minimal	null	null
+/result	Trigger	"<output value=""#form/output""></output>"	minimal	null	null
      at bower_components/chai/chai.js:210
      at bower_components/chai/chai.js:2192
      at assertEqual (tests/utils.js:112)
      at eq (tests/copy-paste.js:37)
      at tests/copy-paste.js:1082
```

emord:
Changed direct `window.location.hash` assignments to use the `window.history.pushstate` api (well supported IE > 10 https://developer.mozilla.org/en-US/docs/Web/API/History).

That failing test is a little confusing. if run by itself it passes. Also if I make that test run one test earlier they all pass. Might take a bit to dig into that one

millerdev:
looking good. ping me if you want another set of eyes on that test

emord:
This failing test was weird. vellum should explicitly change `<output></output>` to `<output />` https://github.com/dimagi/Vellum/blob/master/src/xml.js#L39

The xml it was loading was of format was of format `<output />` https://github.com/dimagi/Vellum/blob/master/tests/static/form/manual-instance-reference.xml#L26

It was comparing against `<output></output>` and passing?

@millerdev do you remember if there is something about copy paste that should change `<output />` to `<output></output>` ?

new buddy @snopoke

millerdev:
>  do you remember if there is something about copy paste that should change `<output />` to `<output></output>` ?

Nope, don't remember anything like that. Seems like that would be a bug.

millerdev:
👍 on ✅

emord:
:white_check_mark:

PR #667, opened 2016-05-24
Case management release

orangejenny:
opening for review

resolves http://manage.dimagi.com/default.asp?226552

### DIFF
--- a/src/atwho.js
+++ b/src/atwho.js
@@ -57,7 +57,7 @@ define([
      *                  category: sent to analytics
      *                  insertTpl: string to add to input when question is selected
      *                  property: sent to analytics
-     *                  useRichText: use rich text editor insert method
+     *                  noRichText: skip rich text editor insert method
      *                  outputValue: use output value in the template
      */
     that.questionAutocomplete = function ($input, mug, options) {
@@ -70,11 +70,11 @@ define([
             insertTpl: '${name}',
             property: '',
             outputValue: false,
-            useRichText: false,
+            noRichText: true,
             functionOverrides: {},
         });
 
-        if (options.useRichText) {
+        if (!options.noRichText) {
             options.insertTpl = '${name}';
             options.functionOverrides.insert = function(content, $li) {
                 // this references internal At.js object
@@ -138,7 +138,7 @@ define([
             };
 
             $input.atwho(_atWhoOptions('/data/'));
-            if (options.useRichText) {
+            if (!options.noRichText) {
                 $input.atwho(_atWhoOptions('#'));
             }
         }
--- a/src/atwho.js
+++ b/src/atwho.js
@@ -79,6 +79,7 @@ define([
             options.functionOverrides.insert = function(content, $li) {
                 // this references internal At.js object
                 this.query.el.remove();
+
                 richText.editor($input).insertExpression(content);
                 if (!this.$inputor.is(':focus')) {
                     this.$inputor.focus();
@@ -100,7 +101,9 @@ define([
                     callbacks: {
                         matcher: function(flag, subtext) {
                             var match, regexp;
-                            regexp = new RegExp('(\\s+|^)' + RegExp.escape(flag) + '([\\w_/]*)$', 'gi');
+                            // Match text that starts with the flag and then looks like a path.
+                            // CKEditor reserves the right to insert arbitrary zero-width spaces, so watch for those.
+                            regexp = new RegExp('([\\s\u200b]+|^)' + RegExp.escape(flag) + '([\\w_/]*)$', 'gi');
                             match = regexp.exec(subtext);
                             return match ? match[2] : null;
                         },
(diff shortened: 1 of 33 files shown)


src/saveToCase.js
```diff
@@ -61,7 +61,7 @@ define([
             var widget = widgets.normal(mug, options),
                 id = options.id,
                 internal_template = options.template;
-            options.noRichText = true;
+            options.noRichText = false;
```
emord: This should still be true?

src/widgets.js
```diff
@@ -445,7 +445,7 @@ define([
             path = options.widgetValuePath || options.path,
             id = options.id || 'property-' + path;
         widget.definition = mug.p.getDefinition(options.path);
-        options.noRichText = true;
+        options.noRichText = false;
```
emord: Should still be `true`

src/core.js
```diff
@@ -748,7 +748,7 @@ define([
         if (target) {
             // the .change fires the validation controls
             if (!widgets.util.getWidget(target, this).options.noRichText &&
-                ((!mug && _this.data.core.form.useRichText !== false &&
+                ((!mug && _this.data.core.form.noRichText !== false &&
```
emord: I'm not sure what I was thinking when I wrote this if statement, it took me a minute to get my head around.

What do you think about adding `noRichText` to the base widget? And it could be overwritten by `richText` widget. That way we never need to look at the form, mug or opts anymore

src/atwho.js
```diff
@@ -70,11 +70,11 @@ define([
             insertTpl: '${name}',
             property: '',
             outputValue: false,
-            useRichText: false,
+            noRichText: true,
             functionOverrides: {},
         });
 
-        if (options.useRichText) {
+        if (!options.noRichText) {
```
millerdev: not really a fan of the double-negative, but I suppose it was necessary because of default value, right?

src/mugs.js
```diff
@@ -480,7 +480,7 @@ define([
             return this.parentMug && this.parentMug.isInRepeat();
         },
         supportsRichText: function() {
-            return this.options.richText && this.form.useRichText !== false;
+            return this.options.richText && this.form.noRichText === false;
```
millerdev: why not `this.options.richText && !this.form.noRichText`

src/writer.js
```diff
@@ -88,7 +88,7 @@ define([
         xmlWriter.writeAttributeString( "xmlns:vellum", "http://commcarehq.org/xforms/vellum" );
         // todo: remove writeIgnoreRichText once bubbles is released
         // Here for future compatibility
-        if (form.writeIgnoreRichText && form.useRichText === false) {
+        if (form.writeIgnoreRichText && form.noRichText === true) {
```
millerdev: would this work? `if (form.writeIgnoreRichText && form.noRichText)`

emord:
```
  1) The data browser when loaded before the form should add ref on drag/drop:
  
  2) The data browser when loaded before the form should add parent ref on drag/drop:
  
  3) The data browser when loaded before the form should add recursive ref on drag/drop:
```

Possibly real failure. Able to consistently reproduce on your branch but not master.

src/atwho.js
orangejenny: agreed, and yes.

src/mugs.js
orangejenny: I wanted to keep the strict equality, though I don't know for sure that the strictness is necessary.

emord: Its possible that when loading an empty form it won't have a value for noRichText. You should be able to fix that by initializing it in `Form`. and then you would be able to get rid of the strict checks

src/saveToCase.js
orangejenny: careless search/replace, will fix

src/widgets.js
orangejenny: fixed

src/mugs.js
orangejenny: although, having a boolean flag use different semantics for `false` and `undefined` would be a bad idea anyway. i'll update this.

orangejenny: > Its possible that when loading an empty form it won't have a value for noRichText. You should be able to fix that by initializing it in Form. and then you would be able to get rid of the strict checks

Didn't see this. https://github.com/dimagi/Vellum/pull/667/commits/91ce41479d43620eb5e1260bd927ff91a714e54a feels not quite complete, because it's not checking for some form property being passed in...but it seems to work.

src/form.js
```diff
@@ -138,6 +138,7 @@ define([
         this.errors = [];
         this.question_counter = 1;
         this.xpath = escapedHashtags.parser(this.hashtagDictionary);
+        this.noRichText = !vellum.opts().features.rich_text;
```
millerdev: Can we please change this to `this.richText = !!vellum.opts().features.rich_text;` and then change all of the places that say `!form.noRichText` to `form.richText`?

src/core.js
orangejenny: https://github.com/dimagi/Vellum/pull/667/commits/ef2d597eda1c4a615e2d86111e6535e67b0b215c - got rid of the mug checks, but still need to check the form property

src/form.js
orangejenny: updated, should be more readable now.

millerdev:
`noRichText` -> `richText` ❤️ so much better. Thanks!

src/richText.js
```diff
@@ -154,12 +154,12 @@ define([
             throw new Error("input should reference exactly one element, " +
                             "got " + input.length);
         }
+        options = options || {};
         var NOTSET = {},
             newval = NOTSET,  // HACK work around async get/set
             editor = input.ckeditor({
                 contentsLangDirection: options.rtl ? 'rtl' : 'ltr',
             }).editor;
-        options = options || {};
```
millerdev: sneaky

millerdev:
👍 on ✅

PR #669, opened 2016-06-08
Make _.template calls 1.8-compatible

orangejenny:
It's backwards-compatible, and someday we'll upgrade to 1.8

@emord

### DIFF
--- a/src/richText.js
+++ b/src/richText.js
@@ -628,6 +628,7 @@ define([
     }
 
     return {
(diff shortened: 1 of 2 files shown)


millerdev:
👍 on ✅

orangejenny:
added test coverage

millerdev:
Test failed with
```
```
kicked...

PR #670, opened 2016-06-08
look at form instead of opts

emord:
@orangejenny will fix your `rtl` issue

### DIFF
--- a/src/javaRosa/util.js
+++ b/src/javaRosa/util.js
@@ -252,7 +252,8 @@ define([
     var insertOutputRef = function(vellum, target, path, mug, dateFormat) {
         var output = getOutputRef(path, dateFormat),
             form = vellum.data.core.form;
-        if ((!mug && vellum.opts().features.rich_text) || (mug && mug.supportsRichText())) {
+        if ((!mug && form.useRichText !== false) ||
+            (mug && mug.supportsRichText())) {
             richText.editor(target).insertOutput(output);
         } else {
             util.insertTextAtCursor(target, output, true);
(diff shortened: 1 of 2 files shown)


src/javaRosa/util.js
```diff
@@ -252,7 +252,8 @@ define([
     var insertOutputRef = function(vellum, target, path, mug, dateFormat) {
         var output = getOutputRef(path, dateFormat),
             form = vellum.data.core.form;
-        if ((!mug && vellum.opts().features.rich_text) || (mug && mug.supportsRichText())) {
+        if ((!mug && form.useRichText !== false) ||
```
orangejenny: Is it possible to initialize `form.useRichText` somewhere so we could get away from doing strict equality checks on booleans?

millerdev: yes please

millerdev: I thought someone had added that to the [form constructor](https://github.com/dimagi/Vellum/blob/bad7a8faa5b58cb27fcf5c2c2230d32c80378481/src/form.js#L111) a while ago. That seems like an ideal place for it.

millerdev:
gtm. not sure if you want to address @orangejenny's comment here or in a different PR

src/javaRosa/util.js
orangejenny: Sorry, original comment wasn't really relevant - forgot that https://github.com/dimagi/Vellum/pull/667/files#diff-62acef2590804164cb77b4b5aced00ebR141 initializes it but hasn't been merged yet.

PR #671, opened 2016-06-09
Bubble text when atwho match fails

orangejenny:
`#case/thing` will bubble
`#form/realquestion` will bubble
`#form/fakequestion` will not bubble
`#caze` will not bubble

@emord

### DIFF
--- a/src/atwho.js
+++ b/src/atwho.js
@@ -145,9 +145,9 @@ define([
                             // Gracelessly force focus back to input...otherwise, later code steals
                             // the focus, or, occasionally, gives focus to a different input.
                             var $inputor = this.$inputor;
-                            _.delay(function() {
+                            _.defer(function() {
                                 $inputor.focus();
-                            }, 100);
+                            });
 
                             return false;   // clear query
                         },
--- a/src/atwho.js
+++ b/src/atwho.js
@@ -79,6 +79,7 @@ define([
             options.functionOverrides.insert = function(content, $li) {
                 // this references internal At.js object
                 this.query.el.remove();
+
                 richText.editor($input).insertExpression(content);
                 if (!this.$inputor.is(':focus')) {
                     this.$inputor.focus();
@@ -140,7 +141,13 @@ define([
                                     options.functionOverrides.insert.call(this, content);
                                 }
                             }
-                            return false;   // allow user to keep typing
+
+                            var $inputor = this.$inputor;
+                            _.delay(function() {
+                                $inputor.focus();
+                            }, 100);
+
+                            return false;   // clear query
                         },
                     },
                     functionOverrides: options.functionOverrides,
--- a/src/atwho.js
+++ b/src/atwho.js
@@ -134,8 +134,9 @@ define([
                         },
                         afterMatchFailed: function(at, $el) {
                             if (options.useRichText) {
-                                var content = $el.html().trim();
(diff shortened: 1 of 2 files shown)


src/atwho.js
```diff
@@ -131,7 +131,16 @@ define([
                                                        options.property);
                             }
                             return value;
-                        }
+                        },
+                        afterMatchFailed: function(at, $el) {
+                            if (options.useRichText) {
+                                var content = $el.html().trim();
+                                if (richText.CASE_REF_REGEX.test(content) || form.isValidHashtag(content)) {
```
emord: I think `CASE_REF_REGEX` should include a something after the case.

typing #case/ causes the #case/ to be erased on space

orangejenny: thanks, fixed. but found that typing `#caze #case/thing` replaces the entire string with a bubble, instead of leaving `#caze` alone, so, not yet ready to merge.

orangejenny: fixed `#caze #case/thing`, although in a terrible way, so this is in theory ready to merge. working on tests.

orangejenny: no longer planning to add tests in this pr. did comment the terrible code.

src/atwho.js
```diff
@@ -131,7 +132,25 @@ define([
...
+                                    options.functionOverrides.insert.call(this, content);
+                                }
+                            }
+
+                            // Gracelessly force focus back to input...otherwise, later code steals
+                            // the focus, or, occasionally, gives focus to a different input.
+                            var $inputor = this.$inputor;
+                            _.delay(function() {
+                                $inputor.focus();
+                            }, 100);
```
millerdev: I assume this is 100ms, which seems like a pretty long time. Does it have to be that long? Would 10ms work?

millerdev:
👍 from me unless 10ms would work instead of 100ms

src/atwho.js
orangejenny: 10ms would likely be fine. I tend to use 100ms delays because it's approximately the longest time that you can wait without a human user perceiving a delay. Some background: https://www.nngroup.com/articles/response-times-3-important-limits/

czue: neat. :books: :bulb:

millerdev: ah, you're right. Personally I tend to go with shorter if possible. That way if there are multiple elements causing delays you can stack a bunch of delays without going over 100ms.

millerdev: although it begs the question: if 10ms would probably be fine, would 0 be fine too? Why not go as short as possible?

orangejenny: hmm, yeah, it might just be about running this code once the current call stack has cleared. i'll try that.

src/atwho.js
```diff
@@ -145,9 +145,9 @@ define([
                             // Gracelessly force focus back to input...otherwise, later code steals
                             // the focus, or, occasionally, gives focus to a different input.
                             var $inputor = this.$inputor;
-                            _.delay(function() {
+                            _.defer(function() {
```
emord: is this still necessary after https://github.com/dimagi/Vellum/pull/672 ?

orangejenny: no, although it does still need code to handle focus & selection correctly: https://github.com/dimagi/Vellum/pull/671/commits/2c292c49c9f2023318abf7b544becf8c4a3ebb5e

orangejenny:
tests passing

PR #672, opened 2016-06-13
Stop focus from jumping to different language's textarea

orangejenny:
@emord 

When bubbles get created, executing `createPopover` has been stealing focus.


millerdev:
👍 on ✅

orangejenny:
self-merging since tests pass

PR #673, opened 2016-06-14
Display data sources error text to end user

orangejenny:
https://github.com/dimagi/commcare-hq/pull/12004#issuecomment-225528836

Before
screen shot 2016-06-14 at 10 56 03 am

After
screen shot 2016-06-14 at 10 52 15 am

### DIFF
--- a/src/datasources.js
+++ b/src/datasources.js
@@ -95,13 +95,13 @@ define([
      *      have been loaded. This function should accept one argument,
      *      a list of data source objects.
      */
-    function getDataSources(callback) {
+    function getDataSources(successCallback, errorCallback) {
         if (dataCache) {
-            callback(dataCache);
+            successCallback(dataCache);
             return;
         }
         if (dataCallbacks) {
-            dataCallbacks.push(callback);
+            dataCallbacks.push(successCallback);
             return;
         }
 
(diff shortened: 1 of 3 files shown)


snopoke:
:+1: from me

emord:
This refers to a display condition in the app right? It might be confusing to say display condition here since questions have those. Is this error also visible on the app page?

orangejenny:
Hmm, yeah, I can see that being confusing. And it's not just form display conditions; a bad case list filter will also trigger this. How about `Your form uses an expression which refers to a case, but cases are not available. Please return to form settings and either...`? It's awkward that the error is displayed in vellum but to fix it you need to leave vellum.

emord:
Yeah, not a great solution here. Seems reasonable to notify them to return to the form/module settings. I don't think that anything in the formbuilder will cause that error at the moment.

You might be able to add a call to get_casedb_schema to the validate build call that uses the jar. then it would be visible on the form page. 

Obviously out of scope for this PR since it's an error from HQ

emord:
Looks good to merge from me unless you want to change the error message first

orangejenny:
@emord https://github.com/dimagi/commcare-hq/pull/12053

src/datasources.js
```diff
@@ -95,13 +95,13 @@ define([
      *      have been loaded. This function should accept one argument,
      *      a list of data source objects.
      */
-    function getDataSources(callback) {
+    function getDataSources(successCallback, errorCallback) {
```
millerdev: Docstring should be updated. Also, `errorCallback` will be ignored on all calls to this function except the first one. Something like the `dataCallbacks` list seems necessary to fix that.

PR #675, opened 2016-06-14
PR feedback

orangejenny:
https://github.com/dimagi/Vellum/pull/673/files/18ad819633187312792b90643426e31a06c55d6a#r67002157

@millerdev

### DIFF
--- a/src/datasources.js
+++ b/src/datasources.js
@@ -85,8 +85,8 @@ define([
 
     function reset() {
         dataCache = null;
-        dataCallbacks = null;
-        errorCallbacks = null;
+        dataCallbacks = [];
+        errorCallbacks = [];
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/datasources.js
```diff
@@ -128,11 +134,13 @@ define([
                     success: finish,
                     error: function (jqXHR, errorType, exc) {
                         finish([]);
-                        if (errorCallback) {
+                        _.each(errorCallbacks, function(callback) {
                             errorCallback(jqXHR, errorType, exc);
```
millerdev: I think this should be `callback(jqXHR, errorType, exc);`

Also, might want to check if `callback` is not undefined before calling it.

src/datasources.js
```diff
@@ -128,11 +134,13 @@ define([
...
-                        } else {
+                        });
+                        if (!errorCallbacks) {
```
millerdev: I think this should say `if (!_.some(errorCallbacks))`

src/datasources.js
orangejenny: oops, thanks. fixed.

src/datasources.js
orangejenny: updated so that the callbacks arrays are always arrays and never null

PR #677, opened 2016-06-15
Add video and video-inline to form export

millerdev:
http://manage.dimagi.com/default.asp?229818

@orangejenny


orangejenny:
Looks good, just waiting on tests.

PR #678, opened 2016-06-15
Allow unknown case properties

orangejenny:
http://manage.dimagi.com/default.asp?228035

Opening for review. Need to fix the failing tests and likely add some new tests.

### DIFF
--- a/src/form.js
+++ b/src/form.js
@@ -153,12 +153,19 @@ define([
 
     Form.prototype = {
         isValidHashtag: function(tag) {
-            return this.hashtagDictionary.hasOwnProperty(this.normalizeHashtag(tag));
+            tag = this.normalizeHashtag(tag);
+            return this.hashtagDictionary.hasOwnProperty(tag) && _.isString(this.hashtagDictionary[tag]);
         },
-        isValidHashtagPrefix: function(prefix) {
-            prefix = this.normalizeHashtag(prefix);
-            var lastSlashIndex = prefix.lastIndexOf("/");
-            return lastSlashIndex !== -1 && this.hashtagDictionary.hasOwnProperty(prefix.substring(0, lastSlashIndex + 1));
+        isValidHashtagPrefix: function(tag) {
+            tag = this.normalizeHashtag(tag);
+            return this.hashtagDictionary.hasOwnProperty(tag) && _.isFunction(this.hashtagDictionary[tag]);
+        },
+        hasValidHashtagPrefix: function(tag) {
+            tag = this.normalizeHashtag(tag);
+            var lastSlashIndex = tag.lastIndexOf("/");
+            return lastSlashIndex !== -1
+                && this.hashtagDictionary.hasOwnProperty(tag.substring(0, lastSlashIndex + 1))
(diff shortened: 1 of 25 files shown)


orangejenny:
It's a veritable rainbow.

screen shot 2016-06-15 at 6 09 50 pm

orangejenny:
Errors are now displayed in popovers rather than the yellow alert that used to show up at the bottom of the section:
screen shot 2016-06-15 at 6 12 05 pm

screen shot 2016-06-15 at 6 12 19 pm

src/widgets.js
```diff
@@ -749,8 +749,13 @@ define([
 
     function getMessages(mug, path) {
         var $messages = $(),
-            seen = {};
+            seen = {},
+            usesRichText = mug.form.vellum.opts().features.rich_text;
         mug.messages.each(path, function (msg) {
+            if (usesRichText && msg.key === "logic-bad-path-warning") {
```
emord: This hides errors for paths that aren't in the form and don't get bubbled so if you have use `/date/question` then teh question will have an error, but there's no warning on the input

src/form.js
```diff
@@ -154,18 +154,18 @@ define([
     Form.prototype = {
         isValidHashtag: function(tag) {
             tag = this.normalizeHashtag(tag);
-            return this.hashtagDictionary.hasOwnProperty(tag) && _.isString(this.hashtagDictionary[tag]);
+            return this.hashtagDictionary.hasOwnProperty(tag) && !_.isFunction(this.hashtagDictionary[tag]);
```
millerdev: Having two types of values in `hashtagDictionary` seems like it could cause confusion, given the difference in meaning if the value is a string or function. I'm not sure I understand this enough to suggest another idea, but one that comes to mind is to put the different types of things in separate dictionaries so they don't need to be filtered like this. Would that make sense, or would it make things worse?

emord: That seems clearer to me. Could have something like `knownHashtags` and `hashtagTransformations` ?

orangejenny: https://github.com/dimagi/Vellum/pull/678/commits/2ccf9b62153c5e0c1183584f32335b317be2a0d7

I don't entirely love this change, as it adds a good deal of repetitive code, but agree that the dictionary with mixed strings & functions is ugly.

src/widgets.js
orangejenny: Thought about this some more, and I'm inclined to just revert the changes that move warnings to popovers, don't think we thoroughly thought through that change.

@amsagoff it's awkward to move the "Unknown question" warnings into popovers because that same warning appears for certain non-bubbling expressions (specifically, absolute paths that start with something other than `/data/`). So right now, on production, you can get something like this:

pasted image at 2016_06_17 17_06

Seems weird to me that we'd move the `#form/question2` part of that error into a popover and leave the rest of it as-is. For the moment, I reverted the popover changes, so all unknown question errors will keep displaying in the yellow box, but let me know if you disagree.

millerdev:
👍 from me. Is this ready to merge?

orangejenny:
Yep, this is good to merge.

PR #679, opened 2016-06-16
Fix html input

emord:
@orangejenny fixes the html weirdness

cc: @NoahCarnahan

### DIFF
--- a/src/richText.js
+++ b/src/richText.js
@@ -497,6 +497,8 @@ define([
         var el = $('<div>').html(text),
             outputMapping = {},
             ret;
+        // replaces each bubble with a guid to be replaced later with the
+        // actual output value. needed so that < and > aren't escaped
         el.find('.label-datanode').replaceWith(function(index, output) {
             var uuid = util.get_guid();
             while (outputMapping.hasOwnProperty(uuid)) {
--- a/src/richText.js
+++ b/src/richText.js
@@ -60,7 +60,6 @@ define([
         },
         downcast: function(element) {
             element.setHtml(applyFormats($(element.getOuterHtml()).data()));
-            element.replaceWithChildren();
         },
         init: function() {
             // TODO: PR to ckeditor to make changing drag ui supported
@@ -103,7 +102,18 @@ define([
         }
     });
 
-    CKEDITOR.config.allowedContent = true;
+    CKEDITOR.config.allowedContent = {
+        span: {
+            classes: 'label,label-*',
+            attributes: 'data-output-value,data-value,data-date-format',
+        },
+        i: {
+            classes: 'fcc,fcc-*,fa,fa-*',
+        },
+        output: {
+            attributes: 'value',
+        },
+    };
     CKEDITOR.config.customConfig = '';
     CKEDITOR.config.title = false;
     CKEDITOR.config.extraPlugins = 'bubbles';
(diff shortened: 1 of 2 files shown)


src/richText.js
```diff
@@ -60,7 +60,6 @@ define([
         },
         downcast: function(element) {
             element.setHtml(applyFormats($(element.getOuterHtml()).data()));
-            element.replaceWithChildren();
```
emord: this never actually did anything

src/richText.js
```diff
@@ -103,7 +102,18 @@ define([
...
+        span: {
+            classes: 'label,label-*',
+            attributes: 'data-output-value,data-value,data-date-format',
+        },
+        i: {
+            classes: 'fcc,fcc-*,fa,fa-*',
+        },
+        output: {
+            attributes: 'value',
+        },
```
millerdev: Before CKEditor we did not do anything special to markup typed in labels (which can be handy when rendering as markdown). Does this have an impact on that?

emord: I'm not sure why markdown would matter here? Within ckeditor you only use plain text like `*bold*` or `**italics**` it's only rendered in html in a separate div using markdown-it

millerdev: I'm not sure if we support it, but [some flavors of markdown](https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#html) support inline HTML.

src/richText.js
```diff
@@ -484,9 +494,22 @@ define([
     }
 
     function unwrapBubbles(text) {
-        var el = $('<div>').html(text);
-        el.find('.label-datanode').children().unwrap();
-        return el.text();
+        var el = $('<div>').html(text),
+            outputMapping = {},
+            ret;
+        el.find('.label-datanode').replaceWith(function(index, output) {
```
orangejenny: explanatory comment would be nice, this isn't exactly self-documenting code

orangejenny:
Looks all right to me, leaving open because I"m not sure if @millerdev 's comments are resolved.

src/richText.js
emord: https://github.com/dimagi/Vellum/pull/679/commits/afeafd3a2db45897064e99495736d6c125dc1846

src/richText.js
emord: https://github.com/dimagi/Vellum/blob/7b6c26b3786f5e15dba7722114dc427c6ad185ac/src/markdown.js#L9 
https://confluence.dimagi.com/display/commcarepublic/Text+Formatting

we have a pretty strict whitelist on what we support. 

I also attempted to put html in a form on my phone and it hid the entire text entire text.

```text <b>bold</b>``` would output ```text ``` on a phone

emord:
expanding on

> fixes the html weirdness

This fixes these issues:

1) when typing < or > ckeditor attempts to interpret those as html tags unless you have a space in between them so `< b` would be `&lt; b` but `<b` would start to be a tag and not be replicated to the non default language properly

2) if you had text `<b>text</b>` it would be shown in the editor as **text**

src/richText.js
millerdev: fair enough

millerdev:
> expanding on
>
>> fixes the html weirdness
>
> This fixes these issues:

let the whack-a-mole games begin

PR #680, opened 2016-06-20
remove snapshot change events

emord:
@orangejenny Realized we never need the snapshot events, so there's no need to pass in a parameter.

cc: @NoahCarnahan

### DIFF
--- a/src/javaRosa/itextWidget.js
+++ b/src/javaRosa/itextWidget.js
@@ -305,6 +305,7 @@ define([
                 if (e.form === widget.form && e.itextType === widget.itextType) {
                     if (widget.getItextValue() === e.prevValue) {
                         // Make sure all the defaults keep in sync.
+                        // why doesn't setValue set the itext value?
                         widget.setItextValue(e.value);
                         widget.setValue(e.value);
                     }
--- a/tests/databrowser.js
+++ b/tests/databrowser.js
@@ -430,6 +430,7 @@ define([
 
             it("should error for unknown properties", function(done) {
                 widget.setValue(escapedDobProp);
+                widget.handleChange();
                 assert(!util.isTreeNodeValid(blue), "expected validation error");
                 assert.lengthOf(widget.getControl().find('.label-datanode-unknown'), 1);
                 event.fire("loadCaseData");
--- a/tests/exporter.js
+++ b/tests/exporter.js
@@ -65,29 +65,4 @@ define([
             );
         });
     });
-
-// TODO test with newlines (should they be preserved? old vellum did not)
-//    var TEST_XML_1 = util.xmlines('' +
-//    '<?xml version="1.0" encoding="UTF-8"?>\
-//    <h:html xmlns:h="http://www.w3.org/1999/xhtml" xmlns:orx="http://openrosa.org/jr/xforms" xmlns="http://www.w3.org/2002/xforms" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:jr="http://openrosa.org/javarosa" xmlns:vellum="http://commcarehq.org/xforms/vellum">\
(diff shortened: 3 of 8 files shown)


orangejenny:
good once tests pass

orangejenny:
test failures look real

src/javaRosa/itextWidget.js
```diff
@@ -305,6 +305,7 @@ define([
                 if (e.form === widget.form && e.itextType === widget.itextType) {
                     if (widget.getItextValue() === e.prevValue) {
                         // Make sure all the defaults keep in sync.
+                        // why doesn't setValue set the itext value?
```
emord: This seems weird, but it's also been here for a while, so didn't want to rock the boat

tests/databrowser.js
```diff
@@ -430,6 +430,7 @@ define([
 
             it("should error for unknown properties", function(done) {
                 widget.setValue(escapedDobProp);
+                widget.handleChange();
```
emord: I don't think `setValue` should call `handleChange` since it's what's used when initializing a widget, so when we use setValue in a test context like this, we should call it explictly. 

@millerdev & @orangejenny does this sound correct?

tests/exporter.js
```diff
@@ -65,29 +65,4 @@ define([
             );
         });
     });
-
-// TODO test with newlines (should they be preserved? old vellum did not)
-//    var TEST_XML_1 = util.xmlines('' +
```
emord: Removed xmlines. This is from at least almost 2 years ago, assuming if no one has asked to preserve newlines then they never will.

tests/javaRosa.js
```diff
@@ -243,7 +247,7 @@ define([
 
             util.assertXmlEqual(
                 call('createXML'),
-                util.xmlines(TEST_XML_2),
```
emord: `xmlines` was a holdover from when these xml "files" were in strings at the bottom of this file. I moved those to files a while ago, but didn't realize they were using this helper

tests/static/javaRosa/output-refs.xml
```diff
@@ -25,7 +25,7 @@
 						<value>first_question</value>
 					</text>
 					<text id="question2-label">
-						<value>question2</value>
+						<value><output value="/data/first_question" vellum:value="#form/first_question" /> <output value="/data/question11" /> <output value="/data/question1/b" /> <output value="/data/question1b" /></value>
```
emord: This should have always been here. It should change with the default language when setItextValue is called

src/javaRosa/itextWidget.js
orangejenny: This does seem weird, moreso after reading through these changes and seeing the various places where you have to call both `setValue` and `setItextValue` together. It seems odd that `setItextValue` is publicly exposed. Do tests fail if you move `setItextValue` inside `setValue`?

tests/databrowser.js
orangejenny: sounds logical

orangejenny:
Each of these changes looks reasonable, though I'm increasingly confused about why `setItextValue` is its own function instead of having `itextLabelWidget` override `setValue` itself. Investigating that a bit doesn't seem like a major additional refactor. (But feel free to push back, given that tests pass on this PR and also that this is basically an interrupt task that I kind of dumped on you.)

src/javaRosa/itextWidget.js
emord: that does raise an error here: https://github.com/dimagi/Vellum/blob/ecbd0fb30b35cc655e834408034437e4ff1a40f6/tests/javaRosa.js#L318

I'd have to look into it more deeply to understand why that happens, but thinking it's probably out of scope for this PR

PR #681, opened 2016-06-20
Update itext overview

orangejenny:
@emord


orangejenny:
no tests, canceled builds

PR #683, opened 2016-06-27
Warn on any question named case (and variants)

millerdev:
http://manage.dimagi.com/default.asp?229679

@orangejenny


orangejenny:
looks good, waiting on tests

millerdev:
✅

PR #684, opened 2016-06-27
Change hashtag syntax

emord:
@orangejenny should match the schema that we discussed.

https://github.com/dimagi/Vellum/pull/684/commits/ab7959ea67b274b0fa04837d15bb01a22ad765b9 main change

https://github.com/dimagi/Vellum/pull/684/commits/47b8b9af3cf1c05c7dd0530d596330de7d4d85a0 fixes the display when you have three levels of hierarchy, been around for a while

buddy @NoahCarnahan

### DIFF
--- a/tests/databrowser.js
+++ b/tests/databrowser.js
@@ -42,7 +42,7 @@ define([
                 "case_id": {
                     reference: {
                         source: "casedb",
-                        subset: "child",
+                        subset: "case",
                         key: "@case_id",
                     },
                 },
@@ -56,30 +56,29 @@ define([
                 name: {},
             },
             subsets: [{
-                id: "mother",
+                id: "parent",
+                name: "parent (mother)",
                 key: "@case_type",
                 structure: {
                     edd: {},
-                },
-                related: {
-                    "first-child": "child",
                 }
             }, {
-                id: "child",
+                id: "case",
+                name: "child",
                 key: "@case_type",
                 structure: {
                     dob: {},
                     invalid: {}
                 },
                 related: {
-                    parent: "mother",
+                    parent: "parent",
                 },
             }],
         }];
 
     describe("The data browser", function () {
-        var escapedDobProp = '`#case/child/dob`',
-            dobProp = '#case/child/dob',
+        var escapedDobProp = '`#case/dob`',
+            dobProp = '#case/dob',
             dataTree;
 
         function getInstanceId(form, src) {
@@ -143,7 +142,7 @@ define([
                     widget = util.getWidget('property-calculateAttr');
                 widget.input.promise.then(function () { 
                     editor.on('change', function() {
-                        assert.equal(mug.p.calculateAttr, "`#case/mother/edd`");
+                        assert.equal(mug.p.calculateAttr, "`#case/parent/edd`");
                         assert.equal(getInstanceId(mug.form, sessionUri), "commcaresession");
                         assert.equal(getInstanceId(mug.form, casedbUri), "casedb");
                         util.assertXmlEqual(call("createXML"), MOTHER_REF_XML,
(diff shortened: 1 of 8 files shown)


tests/databrowser.js
```diff
@@ -157,40 +156,13 @@ define([
                 });
             });
 
-            it("should add recursive ref on drag/drop", function(done) {
```
millerdev: Why did this test need to be removed? It should still be possible to represent recursive relationships with the datasources JSON.

emord: We want a flat list of all case relationships, so that would be impossible. I did the initial implementation poorly and currently it goes:

* child
* parent
  * grandparent

When it should have been

* child
* parent
* grandparent

emord: should add some validation to assert that we don't have recursive relationships, but I don't think that's a blocker and it's reasonable to assume HQ will deliver a valid schema.

orangejenny:
Looks all right to me.
`cd tests; pt "#case/\w+/"` shows a bunch of tests using the `#case/type/property` format. It's cool that they still pass, but they're a bit misleading now.

tests/options.js
```diff
@@ -23,16 +23,25 @@ define(["underscore"], function (_) {
...
                     edd: {},
                 },
                 related: {
-                    "first-child": "child",
+                    parent: "grandparent",
                 }
             }, {
-                id: "child",
+                id: "case",
+                name: "case (child)",
```
orangejenny: thought we were going to just use case type name for the base case type

tests/databrowser.js
millerdev: I think recursive relationships should be possible to represent in the schema. The tree should only show three levels: case, parent, granparent. Still sorta think this test should not have been removed, but probably not a big deal.

emord: Ah I see. I suppose they are still representable, but I didn't want to hardcode caes/parent/grandparent in vellum, so I made a generic flatten everything in this list method. It should be easy enough to change that if we decide recursive relationships are something we want hq to support later (all that would need to happen is to remove `flattenNode`)

emord:
> shows a bunch of tests using the #case/type/property format. It's cool that they still pass, but they're a bit misleading now.

I think those are just in the generic parsing tests. I don't know that that's misleading, since those aren't hardcoded in vellum, it still only looks at the options passed to it by HQ

tests/databrowser.js
millerdev: > if we decide recursive relationships are something we want hq to support later

I think recursive relationships are already supported, and I'm not sure we should break them in the schema. It's possible to create a case that has a parent case of the same type. Vellum test options previously had a slightly more complicated scenario where there was a recursive relationship between two case types:

mother > first-child (child) > mother > first-child (child) ...

This case graph should continue to be supported in the schema. How vellum represents that schema is an implementation detail in vellum. Now that I'm thinking this through I think I don't like how the schema code in HQ was changed.

orangejenny: > I think recursive relationships are already supported

They're supported in basic modules?

millerdev: What happens if vellum gets a schema with recursive relationships now? Will it go into an infinite loop trying to unwind them?

emord: No. the recursive relationship will not be displayed, but the rest of the case types will show up.

tested with (abbreviated):

```
            subsets: [{                                                                                                                                                                                     
                id: "grandparent",                                                                                                                                                                          
                name: "grandparent (household)",                                                                                                                                                            
            }, {                                                                                                                                                                                            
                id: "parent",                                                                                                                                                                               
                name: "parent (mother)",                                                                                                                                                                    
                related: {                                                                                                                                                                                  
                    parent: "grandparent",                                                                                                                                                                  
                    "first-child": "case",                                                                                                                                                                  
                }                                                                                                                                                                                           
            }, {                                                                                                                                                                                            
                id: "case",                                                                                                                                                                                 
                name: "case (child)",                                                                                                                                                                       
                related: {                                                                                                                                                                                  
                    parent: "parent",                                                                                                                                                                       
                },                                                                                                                                                                                          
            }]                                                                                                                                                                                              

```

millerdev: How does HQ know (when it's constructing the schema) that

child > mother > household

is an acceptable way to navigate to a grandparent (household), but

child > mother > first-child

is not?

millerdev: I'm still not sure about the question in my previous comment, but I think it probably doesn't matter since it seems to be an edge case.

No further comments from the peanut gallery for now.

emord:
merging with https://github.com/dimagi/commcare-hq/pull/12252

PR #685, opened 2016-06-27
Better fuse config

emord:
http://manage.dimagi.com/default.asp?229018

This should improve the fuzzy matching.

cc: @NoahCarnahan

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -31,7 +31,7 @@
         "requirejs-plugins": "~1.0.2",
         "At.js": "emord/At.js#override-insert",
         "Caret.js": "0.2.2",
-        "fuse.js": "~2.0.0",
+        "fuse.js": "emord/Fuse#token-option",
 
         "jstree": "~3.2.1",
         "langcodes": "dimagi/langcodes",
--- a/src/fuse.js
+++ b/src/fuse.js
@@ -10,7 +10,10 @@ define([
     util
 ) {
     var FUSE_CONFIG = {
-        keys: ['label', 'name', 'absolutePath', 'hashtagPath'],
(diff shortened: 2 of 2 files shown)


bower.json
```diff
@@ -31,7 +31,7 @@
         "requirejs-plugins": "~1.0.2",
         "At.js": "emord/At.js#override-insert",
         "Caret.js": "0.2.2",
-        "fuse.js": "~2.0.0",
+        "fuse.js": "emord/Fuse#token-option",
```
emord: https://github.com/krisk/Fuse/pull/93

src/fuse.js
```diff
@@ -10,7 +10,10 @@ define([
     util
 ) {
     var FUSE_CONFIG = {
-        keys: ['label', 'name', 'absolutePath', 'hashtagPath'],
+        keys: ['displayPath'],
```
emord: Removing `label` because it doesn't make much sense to be able to search by the label (itext) when starting with #form or #case (i think)

`name === absolutePath` `displayPath = mug.form.useRichText ? mug.hashtagPath : mug.absolutePath`  which looks like what we want to do

src/fuse.js
```diff
@@ -10,7 +10,10 @@ define([
...
+        tokenize: true,
+        tokenSeparator: /\/+/g,
+        threshold: 0.4,
```
emord: the default threshold is 0.6, but tokenization changes the scores that come back. (noticed because one of the atwho tests failed) I don't see any reason not to limit this more unless someone reports that it isn't returning the appropriate results

PR #687, opened 2016-07-01
Remove when

emord:
Saw this the other day. Pretty sure it isn't used anywhere anymore

cc @snopoke

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -43,7 +43,6 @@
     },
     "devDependencies": {
         "equivalent-xml-js": "mwhite/equivalent-xml-js#master",
-        "when": "~2.8.0",
         "chai": ">=1.5.0",
         "xrayquire": "requirejs/xrayquire"
     }


millerdev:
Merging despite test failure. I kicked it once already:
```
```
Maybe we should up the timeout on that test?

emord:
> Maybe we should up the timeout on that test?

Probably an ok short term solution.

I'd prefer to break that test up into manageable components, or at least get down to why that's just now failing so much more often. As a quick test I disabled ckeditor for that test and found that it only decreases the test from 100-300 ms at most locally. It's probably a bigger swing for travis, and I'm not sure if that test was already close to the limit before ckeditor

millerdev:
I think it had been in the 5K-8K ms range in the past. I agree, would be great to understand why the time is increasing now and fix the root cause rather than increasing the timeout. Not sure why I suggested increasing the timeout without first investigating root cause.

I also agree this test is a pain because it's so big. However, one advantage of that is that it's a sort of canary in our coal mine. When this thing starts failing we know we've done something that's making tests slower. Kinda nice to have around.

PR #689, opened 2016-07-06
Fix XML normalization of many output values

millerdev:
http://manage.dimagi.com/default.asp?231477

@emord cc @calellowitz

### DIFF
--- a/src/xml.js
+++ b/src/xml.js
@@ -32,14 +32,16 @@ define([
             value = inner ? $(value) : $("<div />").append(fixGTBug(value));
         }
         var xml = new XMLSerializer(),
+            xmlns = / xmlns:vellum="http:\/\/commcarehq.org\/xforms\/vellum"([ \/>])/g,
             wrapper = /^<([\w:.-]+)(?:\s+[\w:.-]+=(["'])[^]*?\2)*\s*(?:\/>|>([^]*)<\/\1>)$/g,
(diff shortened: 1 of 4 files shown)


src/xml.js
```diff
@@ -32,14 +32,16 @@ define([
             value = inner ? $(value) : $("<div />").append(fixGTBug(value));
         }
         var xml = new XMLSerializer(),
+            xmlns = / xmlns:vellum="http:\/\/commcarehq.org\/xforms\/vellum"([ \/>])/g,
             wrapper = /^<([\w:.-]+)(?:\s+[\w:.-]+=(["'])[^]*?\2)*\s*(?:\/>|>([^]*)<\/\1>)$/g,
-            emptytag = /<(([\w:.-]+)(?:\s+[\w:.-]+=(["'])[^]*?\3)*\s*)><\/\2>/g;
```
millerdev: This regex was the problem, it got exponentially slower as more `<output>`s were added to the XML.

Performance digression: I had a test that used the same label value from the original bug. The test failed in Chrome and Firefox, and the failure mode was a warning that the JS was not responding with the option to kill or wait. In phantomjs the test took ~2s to complete (slow); doubling the number of `<output>`s made it fail. I found that moving the `xmlns` find/replace before the `emptytag` find/replace made the test pass, but it still took a long time (this also reduced the phantomjs test time to ~800ms). With the new simpler regexp the test passes in Chrome (test time reported in browser is .02s), Firefox, and phantomjs (< 50ms), and I added 3x as many `<output>` tags for good measure.

PR #690, opened 2016-07-08
handle hashtags when writing intents

orangejenny:
http://manage.dimagi.com/default.asp?231804

@emord 

cc @kaapstorm


emord:
:+1: on :white_check_mark:

PR #691, opened 2016-07-20
Fix easy ref wording

emord:
@orangejenny from field/dev

> [10:56:11 AM] Jennifer Harlow: Oh, but "Use Raw References" was already checked.  But there were bubbles

buddy @NoahCarnahan


millerdev:
bleh. missed that tests hadn't run. Assume this one doesn't really matter.

PR #693, opened 2016-07-20
creating popovers is not undoable

emord:
http://manage.dimagi.com/default.asp?232527

@millerdev 

We use the undo plugin to gain access to a change event. If `canUndo !== false` then `execCommand` will send a change event which enables the save button

cc: @NoahCarnahan


millerdev:
👍 on ✅

emord:
:white_check_mark:

PR #695, opened 2016-07-21
Add whitespace constant

emord:
http://manage.dimagi.com/default.asp?232759

@millerdev I don't have a good test/explanation for this one. Just the only difference I noticed between selecting with mouse vs hitting enter

buddy @NoahCarnahan

### DIFF
--- a/src/richText.js
+++ b/src/richText.js
@@ -49,6 +49,8 @@ define([
     var CASE_REF_REGEX = /^\`?#case\//,
         FORM_REF_REGEX = /^\`?#form\//,
         REF_REGEX = /^\`?#(form|case)\//,
+        // not sure where this comes from... just grabbed from inspect element
+        CKEDITOR_WHITESPACE_CONSTANT = '​​​​​​​',
         bubbleWidgetDefinition = {
         template:
             '<span class="label label-datanode label-datanode-internal">' +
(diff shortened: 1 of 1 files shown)


src/richText.js
```diff
@@ -49,6 +49,8 @@ define([
     var CASE_REF_REGEX = /^\`?#case\//,
         FORM_REF_REGEX = /^\`?#form\//,
         REF_REGEX = /^\`?#(form|case)\//,
+        // not sure where this comes from... just grabbed from inspect element
+        CKEDITOR_WHITESPACE_CONSTANT = '​​​​​​​',
```
emord: This is 7 zero width spaces (http://www.fileformat.info/info/unicode/char/200B/index.htm). Can't use less or more or they show up in the xml. Guessing it's a separator in ckeditor, but I couldn't find it in their code.

src/richText.js
```diff
@@ -186,19 +188,21 @@ define([
...
+                    expression = $('<p>').append(output).html();
                 }
+                editor.insertHtml(expression + ' ' + CKEDITOR_WHITESPACE_CONSTANT);
             },
             insertOutput: function (xpath) {
                 if (options.isExpression) {
                     throw new Error("cannot insert output into expression editor");
                 }
-                editor.insertHtml(bubbleOutputs(xpath, form) + ' ');
+                editor.insertHtml(bubbleOutputs(xpath, form) + ' ' + CKEDITOR_WHITESPACE_CONSTANT);
```
millerdev: voodooblackmagic

millerdev:
👍 on ✅ ... I guess

PR #696, opened 2016-07-25
Revert "Add whitespace constant"

emord:
Reverts dimagi/Vellum#695

http://manage.dimagi.com/default.asp?232890
http://manage.dimagi.com/default.asp?232902

Didn't like this solution anyway, seems like it was a chrome-specific fix

@orangejenny @millerdev 

@NoahCarnahan


millerdev:
👍 on ✅

PR #698, opened 2016-07-29
Optimize logic manager

millerdev:
This brings the load time of the form in http://manage.dimagi.com/default.asp?232792 down from nearly 3 minutes to <15 seconds (on my local machine) by eliminating exponential complexity in the data browser `fixFormReferences` due to many errors in the form. It also further optimizes the logic manager by eliminating all places where it would inspect every reference known to the form builder. Probably easiest to review by commit.

@emord cc @biyeun

### DIFF
--- a/src/form.js
+++ b/src/form.js
@@ -730,7 +730,7 @@ define([
                 }
             }
             this._logicManager.updatePaths(updates);
-            this.fixBrokenReferences(mug);
+            this.fixBrokenReferences();
             // TODO make Item not a special case
             if (oldId && mug.__className !== "Choice") {
                 // update first child of old parent with matching conflicted nodeID
@@ -863,7 +863,7 @@ define([
             this.insertQuestion(mug, refMug, position, isInternal);
             // should we fix broken references when nodeID is auto-generated?
             //if (!mug.options.isControlOnly && !this.isLoadingXForm) {
-            //    this.fixBrokenReferences(mug);
+            //    this.fixBrokenReferences();
             //}
             return mug;
         },
@@ -917,7 +917,7 @@ define([
                 this.mugMap[this.normalizeHashtag(mug.hashtagPath)] = mug;
             }
         },
-        fixBrokenReferences: function (mug) {
+        fixBrokenReferences: function () {
             function updateReferences(mug) {
                 _this.updateLogicReferences(mug);
             }
--- a/src/logic.js
+++ b/src/logic.js
@@ -14,7 +14,8 @@ define([
             "filter",
             "defaultValue"
         ],
-        NO_SELF_REFERENCES = _.without(XPATH_REFERENCES, 'constraintAttr');
+        NO_SELF_REFERENCES = _.without(XPATH_REFERENCES, 'constraintAttr'),
+        CASE_REF_ID = "#case/";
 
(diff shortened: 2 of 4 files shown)


src/form.js
```diff
@@ -917,7 +917,7 @@ define([
                 this.mugMap[this.normalizeHashtag(mug.hashtagPath)] = mug;
             }
         },
-        fixBrokenReferences: function (mug) {
+        fixBrokenReferences: function () {
```
millerdev: Seems like it could make sense to re-introduce this as an parameter and make this function operate on expressions referencing this mug only if given. That seemed out of scope for these optimizations, so I didn't do it in this PR.

src/logic.js
```diff
@@ -392,10 +435,9 @@ define([
...
-            ret.preload = _.chain(this.all)
-                .filter(function(ref) {
-                    return isCaseReference(ref.path);
-                })
+            ret.preload = _.chain(this.reverse[CASE_REF_ID] || {})
+                .values()
+                .flatten(true)
                 .map(function(ref) {
                     var info = ref.path.split('/'),
                         prop = info[2];
```
millerdev: @emord: is this still correct now that case references look like `#case/property` instead of `#case/type/property`? Do we want something more like (python pseudo code) `ref.path.split("/", 1)[1]`?

src/form.js
emord: Damn. Can't believe I missed that that wasn't used...

src/logic.js
emord: Eek. looks like we missed that when we changed the syntax. I'll change it in a different pr

tests/main.js
```diff
@@ -174,7 +168,15 @@ requirejs(['jquery', 'jquery.vellum'], function ($) {
         $('#load-saved').click(function () {
             load(session.getItem("vellum.tests.main.lastSavedForm") || "");
         });
-        load(""); // load empty form on initial page load
+
+
+        // mocha.env is an object when invoked by mocha-phantomjs
+        if (mocha.env) {
```
emord: It's not clear to me how htis fixes a test time out.

emord:
Do you think it's worthwhile to add a test that has a lot of errors on it? Looks like this didn't do much to change the test run time

millerdev:
I think it requires a fairly big form (lots of questions) plus a lot of errors to make it take a long time, and I don't like the idea of how long a test that big might take to run, even if things are relatively "fast" (guessing it's still going to be near a 2s test). I'm not sure it's worth the potential trade-offs.

tests/main.js
millerdev: it defers `mocha.run()` until after the first form is fully loaded.

Previously tests were running sort of in parallel with the first form loading, and something, which I wasn't able to identify, is taking a long time (~4s) during that first load. When running all tests, that first group is normally
```
  The plugin manager
    ✓ should ignore unknown plugins (58ms)
```
And as you can see it's not a very long test, but the initialization was hitting the 2000ms timeout.

I said "in parallel" above, but actually "interleaved" is a better word. This interleaving of the first form loading at the same time that tests are running is caused by to use of `setTimeout` in the initialization code.

emord: Interesting. Thanks for the explanation

PR #700, opened 2016-07-29
Highlight label on tab

orangejenny:
http://manage.dimagi.com/default.asp?233084

Global event because ckeditor doesn't support `keyup`, only a `key` event that doesn't trigger unless the initial keypress happens while focus is already in the input.

I've played with this a bit, but it might not be a bad idea for someone other than me to pull up this branch, tab around, and see if anything unexpectedly annoying happens. I can ask @dimagi/product to do so if whoever reviews this doesn't.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -218,7 +218,7 @@ define([
 
         $(document).on('keyup', function (e) {
             if (e.which === 9 && $(e.target).hasClass("fd-textarea")) {
-                richText.editor($(e.target)).highlightContents();
+                richText.editor($(e.target)).highlight();
             }
         });
 
--- a/src/core.js
+++ b/src/core.js
@@ -205,6 +205,7 @@ define([
         }
         this.$f.addClass('formdesigner');
         this.$f.empty().append(main_template({ctrl: ctrl, alt: alt}));
+
         $(document).on("keydown", function (e) {
             var ctrlKey = (isMac && e.metaKey) || (!isMac && e.ctrlKey),
                 metaKey = (isMac && e.ctrlKey) || (!isMac && e.metaKey),
@@ -213,6 +214,12 @@ define([
                       (e.shiftKey ? "Shift+" : "") +
                       (metaKey ? "Meta+" : "") + e.keyCode;
             (hotkeys[key] || _.identity).call(_this, e);
+        });
+
+        $(document).on('keyup', function (e) {
+            if (e.which === 9 && $(e.target).hasClass("fd-textarea")) {
+                richText.editor($(e.target)).highlightContents();
+            }
         });
 
         this._init_toolbar();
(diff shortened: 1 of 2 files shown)


emord:
I thought we were only going to highlight if there were no references in the form?

image

orangejenny:
News to me...? It's not a pretty look, but seems a little odd that we'd change behavior for a non-obvious reason. That is, I don't think it'll be obvious to users why their text wasn't highlighted, given that you *can* highlight a bubble.

Then again, I think the main use case for this is likely initial question creation, when there won't be any bubbles, so I don't feel all that strongly about it.

The one case that's weird is if there's only a bubble in the label, this causes that bubble to be selected (in the ckeditor sense of "selected") and typing does nothing. Although, since we insert a space after every bubble, this shouldn't be happening often.

emord:
Agreed that having non-obvious differences in behavior is not great UX.

I think that the oddness was the original reason that this wasn't
implemented. but good for @dimagi/product to confirm what they would like
to see

amsagoff:
Let me make sure we are specifying the pain point that highlighting is
supposed to address, and then I'm hopeful we can find a better, more
targeted solution that will work for everyone.

Here's the flow:

1) User creates a question. The cursor is now in the question ID field with
all the text highlighted.

2) User types a question ID. The text in the label changes to reflect the
question ID. ~90% of the time, this is not what the user wants.

3) User tabs down to the Label field. They likely want to change it
wholesale. User then has to highlight all existing text, rewrite what they
want to see.

The user wants the text highlighted, but only immediately after question
creation. After they specify label text once, it is likely that they want
to edit it rather than totally rewrite it.

There are a lot of flaws in this process we could talk about...but I'm
generally pro highlighting if this is the first time the user is specifying
the Label text, and never highlighting again later (if that's possible).

I'm also up for meeting and brainstorming more creative solutions (teaser:
what if we put the ID field under the label field?)

orangejenny:
@amsagoff That problem flow sounds right.

I'm not keen to do something only the first time, because it's inconsistent and it's potentially ambiguous what "first time" means (only the first time they enter a label box? until the first time they click away from the question? until they change the label?).

Given that the label is ~never correct, what about removing the auto-inserted text altogether and using a placeholder instead? Could set the placeholder to match the id, or to something like "What is the value of question1?"

Probably easier to go to voice.

amsagoff:
yes let's

millerdev:
Another option that should be fairly easy to implement would be to select (highlight) the label text only if it is the same as the question ID. It would be a fairly good approximation for Amelia's proposal of only "highlighting if this is the first time" and it's consistent/simple to understand.

orangejenny:
@amsagoff and I offlined and decided to remove the default question labels altogether, since they aren't required and we expect users to change them. PR coming soon...

PR #702, opened 2016-08-01
Make tests/options.js part of build

orangejenny:
Would like to have access to the test version of vellum options from HQ pages.

@emord

### DIFF
--- a/build.js
+++ b/build.js
@@ -145,6 +145,16 @@
                 //'deferred-components', 
                 'local-deps'
             ]
+        },
+        {
+            create: true,
+            name: 'options',
+            include: [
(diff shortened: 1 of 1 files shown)


emord:
This is for the form builder demo in prelogin right? 

I don't think we would want this options file in there because it enables all plugins and feature flags (like save to case and supply questions). I think it's also likely that when we add something in that file we won't really think about it affecting that page (like when I was developing the rich text editor/hashtags).

Any reason this would be better than having a separate demo.json file in HQ?

orangejenny:
closing in favor of doing a more limited vellum config in prelogin

PR #703, opened 2016-08-05
Validate that choices have labels

orangejenny:
A choice without a label will cause a build error. Add that validation to vellum, so the problem gets surfaced earlier.

cc @benrudolph

### DIFF
--- a/src/javaRosa/plugin.js
+++ b/src/javaRosa/plugin.js
@@ -567,22 +567,9 @@ define([
             function itextValidator(property, name) {
                 return function (mug) {
                     var itext = mug.p[property],
-                        hasItext = false,
-                        languages = itext.itextModel.languages;
-                    if (itext) {
-                        _.each(['default', 'long', 'short'], function(form) {
-                            if (itext.hasForm(form)) {
-                                hasItext = true;
-                                if (_.find(languages, function(lang) {
-                                    return !itext.get('default', lang);
-                                })) {
-                                    hasItext = false;
-                                }
-                            }
-                        });
-                    }
+                        hasItext = itext && itext.hasHumanReadableItext();
                     if (!hasItext && mug.getPresence(property) === 'required') {
-                        if (languages.length === 1) {
+                        if (itext.itextModel.languages.length === 1) {
                             return name + " is required.";
                         } else {
                             return name + " is required for all languages.";
--- a/src/javaRosa/plugin.js
+++ b/src/javaRosa/plugin.js
@@ -567,9 +567,26 @@ define([
             function itextValidator(property, name) {
                 return function (mug) {
                     var itext = mug.p[property],
-                        hasItext = itext && itext.hasHumanReadableItext();
+                        hasItext = false,
+                        languages = itext.itextModel.languages;
+                    if (itext) {
+                        _.each(['default', 'long', 'short'], function(form) {
+                            if (itext.hasForm(form)) {
+                                hasItext = true;
+                                if (_.find(languages, function(lang) {
+                                    return !itext.get('default', lang);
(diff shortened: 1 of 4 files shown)


src/mugs.js
```diff
@@ -1464,6 +1464,22 @@ define([
                     return data.id && data.id.slice(data.id.lastIndexOf("/") + 1);
                 }
             },
+            labelItext: {
```
millerdev: itext-related code typically lives in the javaRosa plugin. What do you think of moving this [there](https://github.com/dimagi/Vellum/blob/6e234ddf7ee1a162e8a65da516fe549d60a2e752/src/javaRosa/plugin.js#L567-L584)? Also, does this `validationFunc` override the [default one for `labelItext`](https://github.com/dimagi/Vellum/blob/6e234ddf7ee1a162e8a65da516fe549d60a2e752/src/javaRosa/plugin.js#L802)?

millerdev: nevermind. I see this is for choices only. I suppose this is a reasonable place to put it.

millerdev:
Would be nice to have a test for the bug that prompted this fix.

src/mugs.js
orangejenny: Oh, interesting, I hadn't found the itext validation function. Reading that, I think this is a bug in that function, where `hasHumanReadableItext` only validates that the itext has at least one form, but it doesn't validate that there's actually text for the form. Added https://github.com/dimagi/Vellum/pull/703/commits/eccef134b8f472c7d7d8db0034229e3dbeac4d3f

orangejenny:
fair, will add test

src/javaRosa/plugin.js
```diff
@@ -567,9 +567,26 @@ define([
...
+                        languages = itext.itextModel.languages;
+                    if (itext) {
+                        _.each(['default', 'long', 'short'], function(form) {
+                            if (itext.hasForm(form)) {
+                                hasItext = true;
+                                if (_.find(languages, function(lang) {
+                                    return !itext.get('default', lang);
+                                })) {
+                                    hasItext = false;
+                                }
```
millerdev: I think the 6 lines above this comment can be reduced to
```js
hasItext = !_.find(languages, function(lang) {
    return !itext.get('default', lang);
});
```
However, that seems incorrect because it resets `hasItext` for each value of `['default', 'long', 'short']` where `itext.hasForm(form)` returns true. Seems like it should short circuit the outer loop on the first form with non-blank itext. Should it be this (now growing context to 8 lines above this and changing `_.each` to `_.some`)?
```js
hasItext = _.some(['default', 'long', 'short'], function(form) {
    return itext.hasForm(form) && !_.find(languages, function(lang) {
        return !itext.get('default', lang);
    });
});
```
I'm also in favor of moving this back into a method on the `ItextItem`. I found the name `hasHumanReadableItext` much easier to read than this double-nested conditional loop. Maybe the name needs to be changed?

millerdev: My suggestion above is still probably not correct. I think
```js
!_.find(languages, function(lang) {
    return !itext.get('default', lang);
});
```
should be
```js
_.find(languages, function(lang) {
    return itext.get('default', lang);
});
```

millerdev: Maybe it would make sense to add some unit tests for `ItextItem.hasHumanReadableItext` for various permutations of itext items with and without "human readable itext"?

orangejenny: Cleaned this up a bit: made it more concise, moved it back into ItextItem, added unit tests.

millerdev:
👍 on ✅

PR #704, opened 2016-08-05
Label placeholders

orangejenny:
Different approach to http://manage.dimagi.com/default.asp?233084 - don't supply default labels (except for choices, where a missing label causes a build error).

cc @benrudolph

### DIFF
--- a/tests/databrowser.js
+++ b/tests/databrowser.js
@@ -187,14 +187,14 @@ define([
                     editor = label.ckeditor().editor,
                     widget = util.getWidget('itext-en-label');
                 widget.input.promise.then(function () { 
-                    editor.on('change', function() {
+                    editor.on('change', _.debounce(function() {
                         assert.equal(mug.p.labelItext.get(), '<output value="#case/dob" />');
                         assert.equal(getInstanceId(mug.form, sessionUri), "commcaresession");
                         assert.equal(getInstanceId(mug.form, casedbUri), "casedb");
                         util.assertXmlEqual(call("createXML"), CHILD_REF_OUTPUT_VALUE_XML,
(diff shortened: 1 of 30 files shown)


millerdev:
👍 on ✅

tests/databrowser.js
```diff
@@ -187,14 +187,14 @@ define([
                     editor = label.ckeditor().editor,
                     widget = util.getWidget('itext-en-label');
                 widget.input.promise.then(function () { 
-                    editor.on('change', function() {
+                    editor.on('change', _.debounce(function() {
```
orangejenny: Adding the placeholder text means that the call to `insertHtml` now fires two change events (guessing one when the placeholder is removed and then one when the new text is added), and the asserts won't pass until the latter one. I don't love adding this timing, but it seemed less fragile than making the test expect exactly two change events...ckeditor docs say: `This event may thus in some cases be fired when no changes happen or may even get fired twice.` which makes me wary of relying on how many times it's fired.

orangejenny:
This turned out to involve a lot of test updates...should be good now, at least they pass locally.

emord:
This seems reasonable to me. 

Have you tested this through a full HQ run through? I know something like this came up before for groups and I recall that there was something non-obvious about the change with mobile or HQ's form validation (don't remember off the top of my head unfortunately).

 If this works through HQ builds and mobile, then looks good to merge

orangejenny:
> If this works through HQ builds and mobile, then looks good to merge

Fair, I had tested texts & selects but hadn't gone through groups. Just tested this with a bunch more questions types (basically everything except barcodes/callouts) and it's working in HQ and on mobile.

PR #705, opened 2016-08-08
allow dash in case property

emord:
@esoergel http://manage.dimagi.com/default.asp?234164

### DIFF
--- a/src/saveToCase.js
+++ b/src/saveToCase.js
@@ -172,7 +172,7 @@ define([
         };
 
     var CASE_XMLNS = "http://commcarehq.org/case/transaction/v2",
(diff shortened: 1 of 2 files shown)


millerdev:
👍 on ✅

millerdev:
Do you think it's worth adding a test for a case property with a dash to verify this actually works?

esoergel:
woo!  Thanks J$, and thanks for the quick turnaround.

emord:
Added/fixed some tests.

millerdev:
👍 on ✅

emord:
:white_check_mark:

PR #708, opened 2016-08-09
Fix save to case case id setvalue

emord:
This was tracking the same setvalue in two different places. once in the form, the other in the mug. this removes it from the form once it is tracked in the mug.

@czue


millerdev:
👍 on ✅

PR #709, opened 2016-08-10
Allow dash in atwho matching

emord:
Jeremy mentioned this to me offhand

@snopoke

### DIFF
--- a/src/atwho.js
+++ b/src/atwho.js
@@ -103,7 +103,7 @@ define([
                             var match, regexp;
                             // Match text that starts with the flag and then looks like a path.
                             // CKEditor reserves the right to insert arbitrary zero-width spaces, so watch for those.
-                            regexp = new RegExp('([\\s\u200b]+|^)' + RegExp.escape(flag) + '([\\w-/]*)$', 'gi');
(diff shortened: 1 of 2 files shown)


src/atwho.js
```diff
@@ -103,7 +103,7 @@ define([
                             var match, regexp;
                             // Match text that starts with the flag and then looks like a path.
                             // CKEditor reserves the right to insert arbitrary zero-width spaces, so watch for those.
-                            regexp = new RegExp('([\\s\u200b]+|^)' + RegExp.escape(flag) + '([\\w_/]*)$', 'gi');
+                            regexp = new RegExp('([\\s\u200b]+|^)' + RegExp.escape(flag) + '([\\w-/]*)$', 'gi');
```
emord: `\w` already contains `_`

millerdev: `\w` doesn't seem like a valid beginning value for a range, so I'm not sure what `\w-/` will do. I think it would be safer to put the dash at the end: `([\\w/-]*)$`.

A test would be nice.

emord: That did work, although I didn't delve into why. fixed

millerdev:
👍 on ✅

snopoke:
PR build failed but push build passed - not sure how important the failure is.

emord:
build failed on phantomjs installed, restarted the build

PR #710, opened 2016-08-10
Remove references in labels when deleting a question

emord:
http://manage.dimagi.com/default.asp?234197

The logic manager did not remove references contained in itext labels when the question was deleted.

cc @snopoke

### DIFF
--- a/src/logic.js
+++ b/src/logic.js
@@ -12,7 +12,13 @@ define([
             "dataParent",
             "repeat_count",
             "filter",
-            "defaultValue"
+            "defaultValue",
+        ],
+        LABEL_REFERENCES = [
+            "helpItext",
+            "constraintMsgItext",
+            "labelItext",
+            "hintItext",
         ],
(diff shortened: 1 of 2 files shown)


src/logic.js
```diff
@@ -12,7 +12,13 @@ define([
...
             "repeat_count",
             "filter",
-            "defaultValue"
+            "defaultValue",
+        ],
+        LABEL_REFERENCES = [
+            "helpItext",
+            "constraintMsgItext",
+            "labelItext",
+            "hintItext",
```
millerdev: Interesting. I wonder if we should add an API for plugins to register property names that may contain logic references? As is this couples the logic manager to javaRosa. Not a blocker, but something to think about.

millerdev:
👍 on ✅

src/logic.js
emord: Yeah. I thought about that for a bit since this is another thing that links javaRosa to core. Decided to punt on it since it would be more effort to try to route these through the plugin system for now.

PR #711, opened 2016-08-10
Fix cursor hiding after insert expression bubble

millerdev:
Second attempt at http://manage.dimagi.com/default.asp?232759 (first #695)

There are still strange editing behaviors related to whitespace after widgets (bubbles) in CKEditor. The most annoying to me is on backspace or left-arrow to remove or move within spaces following a bubble (sometimes the widget is selected instead of the conventional action of deleting or moving past the space before the cursor). Also selecting a different mug and then returning causes the ZWS chars to be removed or converted to regular spaces, which can cause a mildly inconsistent editing experience.

@emord

### DIFF
--- a/src/richText.js
+++ b/src/richText.js
@@ -49,6 +49,8 @@ define([
     var CASE_REF_REGEX = /^\`?#case\//,
         FORM_REF_REGEX = /^\`?#form\//,
         REF_REGEX = /^\`?#(form|case)\//,
+        // http://stackoverflow.com/a/16459606/10840
+        isWebkit = 'WebkitAppearance' in document.documentElement.style,
         bubbleWidgetDefinition = {
         template:
             '<span class="label label-datanode label-datanode-internal">' +
@@ -132,6 +134,14 @@ define([
      *        arguments are editor, ckwidget
      */
     var editor = function(input, form, options) {
+        var TRAILING_SPACE = " ";
+        if (isWebkit) {
+            // HACK use ZWS to fix cursor movement/hiding near bubble
+            TRAILING_SPACE = "\u200b ";
+        }
+        function insertHtmlWithSpace(content) {
+            editor.insertHtml(content + TRAILING_SPACE);
+        }
         var wrapper = input.data("ckwrapper");
         if (wrapper) {
             return wrapper;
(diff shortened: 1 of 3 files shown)


tests/richText.js
```diff
@@ -413,8 +413,8 @@ define([
             _.each([
                 ["one two", 3, "one/data/text two"],
                 ["one two", 4, "one /data/text two"],
-                ["one\n\ntwo", 3, "one/data/text\n\ntwo"],
-                ["one\n\ntwo", 4, "one\n/data/text\ntwo"],
+                ["one\n\ntwo", 3, "one/data/text \n\ntwo"],
+                ["one\n\ntwo", 4, "one\n/data/text \ntwo"],
```
millerdev: I always found it a bit odd that the extra space did not show up in these tests. Now it does, although it's inconsistent when editing a label, for example, because the trailing space will eventually get stripped in some cases after selecting the mug a couple times. I believe CKEditor or something in Vellum is collapsing whitespace, which seems like a separate bug.

src/richText.js
```diff
@@ -522,7 +532,7 @@ define([
         return html.replace(/<p>&nbsp;<\/p>/ig, "\n")
                    .replace(/<p>/ig,"")
                    .replace(/<\/p>/ig, "\n")
-                   .replace(/(&nbsp;|\xa0)/ig, " ")
+                   .replace(/(&nbsp;|\xa0|\u200b | \u200b|\u200b)/ig, " ")
```
millerdev: This is ugly, but I think better than leaving them in there.

src/richText.js
```diff
@@ -132,6 +134,14 @@ define([
      *        arguments are editor, ckwidget
      */
     var editor = function(input, form, options) {
+        var TRAILING_SPACE = " ";
+        if (isWebkit) {
+            // HACK use ZWS to fix cursor movement/hiding near bubble
+            TRAILING_SPACE = "\u200b ";
```
millerdev: I tried using the [7-ZWS constant](https://github.com/ckeditor/ckeditor-dev/blob/7fda187a7961d88cc162a2333ecbf821f2edc7d2/core/selection.js#L1147) (`CKEDITOR.dom.selection.FILLING_CHAR_SEQUENCE`) used by CKEditor, but it didn't work as expected. It just put 7 ZWS's in there and I could tell that they were there because I could move the cursor through or delete them (which feels like the cursor got stalled). Because of that I settled on one, which provides a better editing experience than 7.

emord:
Feels hacky, but I don't think there's a non-hacky way to do it. Did you test in FF and Chrome (or anything else we support)? If so, :+1:

millerdev:
Yes, I did test in FF to verify it didn't break there. The ZWS should only be added in Webkit browsers.

I think this is actually a bug in CKEditor, and I hope to eventually try fixing it there, along with other editing annoyances related to widgets, and submitting a patch.

PR #712, opened 2016-08-11
Fix loading case ids with same string

emord:
http://manage.dimagi.com/default.asp?234647#1200207

This wouldn't correctly load the case id if multiple save to case questions were using the the same value for the case_id. Example: create two cases in a form and use 'uuid()' to give them id, only one of them in vellum would have kept the value

@millerdev @snopoke

### DIFF
--- a/src/saveToCase.js
+++ b/src/saveToCase.js
@@ -632,7 +632,7 @@ define([
                 if (caseIdRegex.test(value.ref)) {
                     mug.p.case_id = value.value;
                     mug.form.dropSetValues(function(inner) {
-                        return value.value === inner.value;
+                        return value.ref === inner.ref;
                     });
                 }
             });
--- a/tests/saveToCase.js
+++ b/tests/saveToCase.js
@@ -194,7 +194,7 @@ define([
             var create1 = util.getMug("create1"),
                 create2 = util.getMug("create2");
             assert.equal(create1.p.case_id, "1");
-            assert.equal(create2.p.case_id, "2");
+            assert.equal(create2.p.case_id, "1");
             util.assertXmlEqual(call("createXML"), CREATE_2_PROPERTY_XML);
         });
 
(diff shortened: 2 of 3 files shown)


src/saveToCase.js
```diff
@@ -632,7 +632,7 @@ define([
                 if (caseIdRegex.test(value.ref)) {
                     mug.p.case_id = value.value;
                     mug.form.dropSetValues(function(inner) {
-                        return value.value === inner.value;
+                        return value.ref === inner.ref;
```
millerdev: Should this also be checking `value.event === inner.event` as well? `ref` and `event` are the two things I think of as making a `setvalue` node "unique" (although they need not actually be unique because ordering of `setvalue`s in the form is significant as well).

millerdev:
👍 on ✅ unless you want to check on `event` as well. Or that could be added in a separate PR.

src/saveToCase.js
emord: I'm going to punt on that for this PR, until I can think more about it.

First thoughts:

1) we don't guarantee ordering of setvalues (I believe they are in mug order, and then any extra ones get added to the end), which could be confusing if there's random setvalues. 
2) save_to_case is a (somewhat) supported question with a known case_id there should only be one in the form, anything else is a bug (although would be great to have a javascript `soft_assert` for this).

millerdev:
Build failed on timeouts. Kicked.
```

2) The rich text editor in vellum  "before each" hook for "should show the markdown preview":
```

millerdev:
> we don't guarantee ordering of setvalues (I believe they are in mug order, and then any extra ones get added to the end), which could be confusing if there's random setvalues. 

Small correction:

We sort of guarantee order of setvalues since the order is significant. However, we do not provide as much flexability in ordering as is possible when editing XML directly since we do not provide a way to change setvalue order independently of mug order. So the order we guarantee is the order of mugs in the form. We also try to preserve the relative order of setvalues that are not connected to a mug, but I think they are always put before mug setvalue nodes.

PR #713, opened 2016-08-11
Correctly parse case properties with dashes

emord:
@esoergel 

follow up to https://github.com/dimagi/Vellum/pull/705

cc @snopoke


millerdev:
👍 on ✅

millerdev:
"Vellum adds all question types and attributes" test timed out again. Wonder if we should up the timeout on that one since I've been seeing that one fail a lot lately. Kicked.

PR #714, opened 2016-08-12
Allow multimedia to fulfill itext requirement

orangejenny:
Followup for https://github.com/dimagi/Vellum/pull/703 : building an app requires that choices have labels, but their "labels" might be images, etc., not text.

@emord 

cc @biyeun


emord:
:+1: on :white_check_mark:

orangejenny:
tests pass; self-merging

PR #715, opened 2016-08-16
Change font size to the same as hq style

emord:
http://manage.dimagi.com/default.asp?235175#1202735

@orangejenny This is the default for textareas in bootstrap 3

@snopoke


orangejenny:
looks fine, no tests

PR #718, opened 2016-08-17
Remove parse fields

emord:
This was for the original spec of the printing widget. It was removed due to a bug in the uploader and then we switched to binary zebra print templates. I don't think it will ever be used again

@snopoke


orangejenny:
hooray for removing unused code

PR #719, opened 2016-08-18
fix dom leak

emord:
Only adds autocomplete to inputs when input is focused.

With easy references there is a dom leak for atwho. hint/validation/help/something itext's are hidden by default. ckeditor initializes these with `contenteditable=false` (known: https://github.com/dimagi/Vellum/blob/f6cbe1ff9ff40dcc330fa0705a7a040f569463ae/src/richText.js#L234-L235) this means that atwho was initializing, but never destroying those elements or the references they held unless each input was focused everytime a user changed a question.

@orangejenny @snopoke

### DIFF
--- a/src/atwho.js
+++ b/src/atwho.js
@@ -164,10 +164,12 @@ define([
                 };
             };
 
-            $input.atwho(_atWhoOptions('/data/'));
-            if (options.useRichText) {
-                $input.atwho(_atWhoOptions('#'));
-            }
+            $input.on('focus', function () {
+                $input.atwho(_atWhoOptions('/data/'));
+                if (options.useRichText) {
+                    $input.atwho(_atWhoOptions('#'));
+                }
+            });
         }
 
         addAtWhoToInput();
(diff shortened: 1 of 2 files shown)


src/atwho.js
```diff
@@ -164,10 +164,14 @@ define([
                 };
             };
 
-            $input.atwho(_atWhoOptions('/data/'));
-            if (options.useRichText) {
-                $input.atwho(_atWhoOptions('#'));
-            }
+            $input.on('focus', function () {
+                if (!$input.data('atwho')) {
```
orangejenny: `one` instead of `on` would do the same thing in less code, but that's just a matter of style

orangejenny:
good to merge when tests pass

src/atwho.js
emord: updated. I didn't know about `one`

snopoke:
tests failing

emord:
:white_check_mark: for some reason the push builds are way more prone to random timeouts. (this one was in atwho, not the 10s test)

PR #722, opened 2016-08-19
Move to phantomjs 2

emord:
before:
./test  166.22s user 1.45s system 97% cpu 2:51.17 total
after:
./test  114.96s user 1.21s system 105% cpu 1:50.41 total
:racehorse: 

`phantomjs-prebuilt` tries to be smart about using a phantomjs that is already on your system, so if you have one installed you might need to update it or uninstall it.

@millerdev I'm hoping this solves the timeout issues on recent prs

cc @snopoke


built.html
```diff
@@ -0,0 +1,67 @@
...
+    
+    <link href="tests/lib/font-awesome.css" type="text/css" rel="stylesheet" />
+    <link href="tests/lib/font-commcare.css" type="text/css" rel="stylesheet" />
+    <link href="tests/lib/hqstyle-core.css" type="text/css" rel="stylesheet"></link>
+    
+    <!-- Mocha and test-only dependencies -->
+    <!-- todo: call from tests/main.js -->
+    <link href="node_modules/mocha/mocha.css" type="text/css" rel="stylesheet"></link>
+    <script>
+      window.useBuilt = true;
```
emord: It wouldn't properly load a file with query string. I got fed up with it and just copied over index.html. not the best solution, but i just wanted it to work

.travis.yml
```diff
@@ -1,8 +1,22 @@
 language: node_js
 sudo: false
+
 node_js:
  - stable
-before_script:
- - phantomjs --version
-script: 
+
+cache:
```
emord: https://github.com/travis-ci/travis-ci/issues/3225#issuecomment-200965782

Should cache phantomjs for us so that we don't always fail on installing

orangejenny:
Agree this copying index.html into built.html isn't a good solution, but I'm also fairly frustrated with tests timing out.

millerdev:
🎉 !yay! 🎉

PR #720, opened 2016-08-18
Don't write unknown case hashtags to xml

emord:
Only write "valid" hashtags to the xml (where valid is a "known" hashtag conversion)

@orangejenny @snopoke

PR #723, opened 2016-08-22
parser references form instead of dictionary

emord:
http://manage.dimagi.com/default.asp?235588

introduced in https://github.com/dimagi/Vellum/pull/720/commits/52436344d4d1fd4effc0e6139c2993b9bce22f9e
hashtagDictionary changes addresses, so when the form's
hashtagdictionary changes, the parser's does not

cc @snopoke

### DIFF
--- a/src/xpath.js
+++ b/src/xpath.js
@@ -4,14 +4,14 @@ define([
     xpath
 ) {
     return {
-        makeXPathModels: function (hashtagToXPath, hashtagToTransformation) {
+        makeXPathModels: function (form) {
             return xpath.makeXPathModels({
                 isValidNamespace: function (namespace) {
                     return namespace === 'form' || namespace === 'case';
                 },
                 hashtagToXPath: function (hashtagExpr) {
-                    if (hashtagToXPath.hasOwnProperty(hashtagExpr)) {
-                        return hashtagToXPath[hashtagExpr];
+                    if (form.hashtagDictionary.hasOwnProperty(hashtagExpr)) {
+                        return form.hashtagDictionary[hashtagExpr];
                     }
 
                     // If full hashtag isn't recognized, remove the property name and check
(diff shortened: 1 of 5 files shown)


emord:
:white_check_mark: @orangejenny @millerdev

tests/escapedHashtags.js
```diff
@@ -63,7 +63,7 @@ define([
                     "#form/text1": "/data/text1",
                     "#form/text2": "/data/text2",
                 },
-                xpathParser = xpath.createParser(xpath.makeXPathModels(translationDict));
+                xpathParser = xpath.createParser(xpath.makeXPathModels({hashtagDictionary: translationDict}));
```
orangejenny: why doesn't this also need to pass a `hashtagTransformations`?

src/xpath.js
```diff
@@ -4,14 +4,14 @@ define([
     xpath
 ) {
     return {
-        makeXPathModels: function (hashtagToXPath, hashtagToTransformation) {
+        makeXPathModels: function (form) {
```
orangejenny: It's a little weird that this takes a `form` parameter that isn't necessarily a form, but rather an object that has properties for `hashtagDictionary` and `hashtagTransformations`.

orangejenny:
Minor comments, but overall I think this an improvement (beyond fixing the bug), having been confused/aggravated in the past by the multiple places hashtags are stored.

millerdev:
> so when the form's hashtagdictionary changes, the parser's does not

I'm guessing that you phrased that the opposite from what you intended to. Looks like the parser now holds a reference to the form, so if `form.hashtagDictionary` or `form.hashtagTransformations` changes then those changes also apply in the parser as well. Maybe I didn't understand your description of what you fixed?

Would be nice to have a test for the bug that's being fixed here. It doesn't look like you added one, but maybe there was already a test for it that was passing before, but should have failed?

PR #725, opened 2016-08-22
Move codemirror to bower

emord:
moves codemirror out of `lib/`

@orangejenny @millerdev 

cc @snopoke


bower.json
```diff
@@ -40,7 +40,8 @@
         "MediaUploader": "dimagi/MediaUploader#master",
         "xpath": "dimagi/js-xpath#v0.0.3",
         "jsdiff": ">=1",
-        "markdown-it": "~6.0.0"
+        "markdown-it": "~6.0.0",
+        "codemirror": "3.24.0"
```
emord: HQ uses 5.5.0, but https://github.com/dimagi/Vellum/blob/master/lib/codemirror/xml.js is not able to run that version (we don't use HQ's version)

orangejenny:
test failure looks real, need to update build.js

orangejenny:
good to merge once tests pass

PR #726, opened 2016-08-22
Bootstrap 3.3.7

orangejenny:
@emord The `should destroy popover after destroy` test is the one that's problematic. It's fine in the browser, but when run in the console, it seems to cause later tests to fail. Reordering it to be the last test in its `describe` block makes all tests pass. Which doesn't feel great, but I'm not enthused about spending more time on figuring out what's going wrong.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -33,7 +33,7 @@
         "Caret.js": "0.2.2",
         "fuse.js": "~2.3.0",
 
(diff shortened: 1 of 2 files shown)


millerdev:
ugly test fix, but cool upgrade

👍 on ✅

PR #727, opened 2016-08-22
jQuery 3

orangejenny:
Somewhat surprisingly, tests passed without any code changes.

Hold off on merge until HQ is also ready to upgrade, which looks like it's going to be more of an undertaking.

@emord

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -22,7 +22,7 @@
     ],
     "dependencies": {
         "underscore": "~1.6.0",
-        "jquery": "~1.11.1",
+        "jquery": "~3.1.0",
         "requirejs": "~2.1.11",
         "requirejs-text": "~2.0.12",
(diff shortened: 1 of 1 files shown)


millerdev:
Test failed on both test runs: The rich text editor in vellum popovers should destroy popover after destroy. Kicked.

orangejenny:
Hrm, I hope that test doesn't become a thorn in our side(s).

millerdev:
failed again. it's timing out

orangejenny:
Yeah, it's timing out rather consistently, will dig into it more.

Also realized that this breaks drag & drop, which our tests all fake at least a little bit.

orangejenny:
Closing this for now, will re-visit after we get on jQuery 2.

PR #728, opened 2016-08-23
fully fix save to case

emord:
http://manage.dimagi.com/default.asp?234191

cc @snopoke

### DIFF
--- a/src/saveToCase.js
+++ b/src/saveToCase.js
@@ -49,14 +49,6 @@ define([
             indexesCase(mug) || attachmentCase(mug);
     }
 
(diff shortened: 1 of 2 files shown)


src/saveToCase.js
```diff
@@ -541,7 +532,17 @@ define([
                     });
                 }
                 return $([]);
-            }
+            },
+            getSetValues: function(mug) {
```
emord: I think this interface came after save to case was implemented and I missed it in the previous PR. definitely the correct way to do this

snopoke:
:+1: on :white_check_mark:

PR #730, opened 2016-08-24
Fix null path in case references

millerdev:
http://manage.dimagi.com/default.asp?235138

@emord cc @esoergel


emord:
:+1: on :white_check_mark:

PR #731, opened 2016-08-24
Upgrade stuff

emord:
Some upgrades that don't seem to break anything

cc @czue

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
 
         "less": "~1.7.0",
         "csso": ">=1.3.11 <2.0.0",
-        "requirejs": "~2.1.14",
+        "requirejs": "~2.2.0",
 
         "grunt": "~1.0.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -25,7 +25,7 @@
 
         "less": "~1.7.0",
         "csso": ">=1.3.11 <2.0.0",
-        "requirejs": "~2.1.14",
+        "requirejs": "~2.2.0",
```
millerdev: interesting. do we need to have `requirejs` installed by both bower and npm?

millerdev:
👍 on ✅

package.json
emord: yeah for now at least. `make` uses the npm version as it's in npm's path, but all of our references in js and html point to bower_components. shouldn't be hard to make it only use the npm version though

PR #732, opened 2016-08-25
Restrict relationship in save to case

emord:
http://manage.dimagi.com/default.asp?235938#1207991

@czue


czue:
:boom: thanks! looks good on :white_check_mark:

PR #738, opened 2016-08-29
Move model iteration's setvalues into mug

emord:
@millerdev Uses the correct api for setvalues, and no longer puts them in the form

buddy @czue

### DIFF
--- a/src/modeliteration.js
+++ b/src/modeliteration.js
@@ -86,7 +86,6 @@ define([
             },
             getExtraDataAttributes: function (mug) {
                 // HACK must happen before <setvalue> and "other" <instance> elements are written
-                prepareForWrite(mug);
                 if (!mug.p.dataSource.idsQuery) {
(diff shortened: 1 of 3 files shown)


emord:
@amstone326 This should fix the model iteration setvalue issue you were running into

amstone326:
😄 😄

src/modeliteration.js
```diff
@@ -86,7 +86,6 @@ define([
             },
             getExtraDataAttributes: function (mug) {
                 // HACK must happen before <setvalue> and "other" <instance> elements are written
-                prepareForWrite(mug);
```
millerdev: Does the `HACK` comment above this line pertain to this line only or to the whole method? Seems like the former, in which case it should also be removed.

emord: oops. yeah, missed that

millerdev:
nice!

👍 on ✅

PR #739, opened 2016-08-30
.fd-column visibility

orangejenny:
http://manage.dimagi.com/default.asp?235825

The `_this.$f.find('.fd-content-right .fd-column').addClass('hide');` lines fix the bug, other changes are minor cleanup.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1478,7 +1478,7 @@ define([
         $props.animate({}, 200);
 
         this.showContentRight();
-        $props.hide();
+        $props.addClass("hide");
 
         this._setPropertiesMug(mug);
         var $content = this.$f.find(".fd-props-content").empty(),
@@ -1511,7 +1511,7 @@ define([
         mug.on("messages-changed", refreshMessages, null, "teardown-mug-properties");
         refreshMessages();
 
-        $props.show();
+        $props.removeClass("hide");
         this.adjustToWindow();
         this.$f.find('.fd-help a').fdHelp();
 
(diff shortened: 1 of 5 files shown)


src/core.js
```diff
@@ -1475,10 +1477,9 @@ define([
         this.$f.find('.fd-default-panel').addClass('hide');
 
         /* update display */
```
emord: nit: looks a little weird to have 

```
code
(whitespace)
comment
(whitespace)
code
```

emord:
Thanks for updating the whole file to one way of doing things. What's the reason to prefer `$element.addClass('hide')` to `$element.hide()`?

Test would also be nice (and you could take credit for test 1000 :) )

src/core.js
orangejenny: i suppose that comment wasn't adding value anyway

orangejenny:
I do like round numbers. Test added.

> What's the reason to prefer $element.addClass('hide') to $element.hide()?

I tend to avoid `show` because it unilaterally sets `display` to `block`, which isn't always desirable (most often tripped up by this when hiding and then re-showing something that was previously displayed `inline-block`). At some point jQuery fixed this (possibly https://jquery.com/upgrade-guide/3.0/#breaking-change-show-hide-and-toggle-methods-now-respect-more-stylesheet-changes) but 1.11 uses the the bad behavior.

emord:
Thanks for the explanation, makes a lot of sense (although maybe we won't need to worry in a couple of weeks :) )

:+1: on :white_check_mark:

PR #740, opened 2016-08-30
Don't require space before matching atwho

orangejenny:
http://manage.dimagi.com/default.asp?235907

Verified that this behaves reasonably if you type something like `/data/data/question` (where `/data/` is the at string).

@emord


emord:
Awesome. Looks like I was wrong on that ticket. (test would be good, although I'm not sure how feasible it is to do well)

:+1:  on :white_check_mark:

PR #741, opened 2016-08-31
Don't use ready 'event'

orangejenny:
Shot in the dark at http://manage.dimagi.com/default.asp?234272 ...wondering if the issue is that the onready handler that calls `adjustToWindow` isn't reliable. See https://jquery.com/upgrade-guide/3.0/#breaking-change-on-quot-ready-quot-fn-removed

@emord


emord:
:+1: on :white_check_mark:

orangejenny:
tests pass

PR #743, opened 2016-09-01
Fix run-on hashtag

millerdev:
http://manage.dimagi.com/default.asp?235899

The first commit fixes the bug, the rest are optimizations and refactoring/simplification. Probably best to review commits individually.

@emord

### DIFF
--- a/src/escapedHashtags.js
+++ b/src/escapedHashtags.js
@@ -36,40 +36,6 @@ define([
         DELIMITER = "`",
         ID_CHAR = /^[\w.-]/;
 
-    function toXPath(input, xpathParser) {
-        return transform(input, function(input) {
-            return xpathParser.parse(input).toXPath();
-        });
-    }
-
-    function toHashtag(input, xpathParser) {
-        return transform(input, function(input) {
-            return xpathParser.parse(input).toHashtag();
-        });
-    }
-
-    /*
-     * hopefully robust parser that transforms any input (whether hashtag,
-     * xpath, or escaped hashtag) into escaped hashtag used by internal structures
-     */
-    function toEscapedHashtag(escaper, input) {
-        if (!input) { return input; }
-
-        // TODO eliminate transform(input). We should always know whether the
-        // thing we're passing in is escaped or not (and we should only escape
-        // unescaped values). As is, this will normalize an expression that has
-        // both escaped and unescaped hashtags, which is invalid syntax,
-        // probably a bug, and should not be supported.
-        var hashtag = transform(input);
-
-        try {
-            return escaper.parse(hashtag).toHashtag();
-        } catch (err) {
-            // a bad xpath. let's just return the given input
-            return input;
-        }
-    }
-
     /*
      * transforms escaped hashtags based on transformFn
      * 
@@ -164,8 +130,7 @@ define([
         var xpathParser = xpath.createParser(xpath.makeXPathModels(hashtagInfo)),
             escapingModels = xpath.makeXPathModels(hashtagInfo, decorateHashtagger),
             escapingParser = xpath.createParser(escapingModels),
-            baseHashtagExpr = escapingModels.HashtagExpr,
-            escapeHashtags = _.partial(toEscapedHashtag, escapingParser);
+            baseHashtagExpr = escapingModels.HashtagExpr;
 
         escapingModels.HashtagExpr = function (definition) {
             baseHashtagExpr.call(this, definition);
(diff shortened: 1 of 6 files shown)


src/escapedHashtags.js
```diff
@@ -25,90 +25,16 @@
...
 ], function(
+    _,
     xpath
 ) {
     var OUTSIDE_HASHTAG = 0,
         INSIDE_HASHTAG = 1,
         DELIMITER = "`",
-        defaultParser = xpath.createParser(xpath.makeXPathModels({}, {}));
-
-    function toXPath(input, xpathParser) {
```
millerdev: This function had tests, but was not used by vellum.

src/escapedHashtags.js
```diff
@@ -25,90 +25,16 @@
...
-    function toEscapedHashtag(input, xpathParser) {
-        if (!input) { return input; }
-        xpathParser = xpathParser || defaultParser;
-
-        var hashtag = transform(input);
-
-        try {
-            var parsed = xpathParser.parse(hashtag);
-            parsed = xpathParser.parse(parsed.toHashtag());
-            return transformHashtags(parsed, xpathParser.models, function(input) {
```
millerdev: parse/render/parse/render simplified to a single parse/render: https://github.com/dimagi/Vellum/pull/743/files#diff-252bc0c8b00cd8556f4f3c24d84cd59fR154

src/escapedHashtags.js
```diff
@@ -186,16 +115,47 @@ define([
...
                     throw new Error("Invalid XPath");
                 }
-                var parsed = xpathParser.parse(toHashtag(input, xpathParser));
+                // TODO eliminate transform(input) here; should not be needed.
+                // transform should only be applied to data that is known to be
+                // escaped instead of every expression before parsing.
+                var parsed = xpathParser.parse(transform(input));
                 parsed.toEscapedHashtag = function() {
-                    return toEscapedHashtag(this.toHashtag(), xpathParser);
+                    var expr = this.toHashtag();
```
millerdev: Unfortunate that we still need to render here, then re-parse and render again below. Seems like there should be a way to pass something to `this.toHashtag()` to have it escape hashtags as it renders them (would require changes to the parser hashtag rendering code).

src/xpath.js
```diff
@@ -23,26 +29,20 @@ define([
...
-                        for (var key in hashtagConfig.hashtagDictionary) {
-                            if (hashtagConfig.hashtagDictionary.hasOwnProperty(key)) {
-                                if (hashtagConfig.hashtagDictionary[key] === xpathExpr)
-                                    return key;
-                            }
-                        }
-                        return null;
+                    var expr = xpath_.toXPath();
+                    if (hashtagInfo.invertedHashtagMap.hasOwnProperty(expr)) {
+                        return hashtagInfo.invertedHashtagMap[expr];
```
millerdev: Inverted hashtag map allows hash lookup rather than sequential search.

src/form.js
```diff
@@ -125,7 +125,8 @@ define([
 
         this.formName = 'New Form';
         this.mugMap = {};
-        this.hashtagDictionary = {};
+        this.hashtagMap = {};
+        this.invertedHashtagMap = {};
```
millerdev: Renamed `hashtagDictionary` to `hashtagMap` because `invertedHashtagDictionary` was way too long.

src/escapedHashtags.js
```diff
@@ -32,6 +32,7 @@ define([
     var OUTSIDE_HASHTAG = 0,
         INSIDE_HASHTAG = 1,
         DELIMITER = "`",
+        ID_CHAR = /^[\w.-]/,
```
orangejenny: Picky, but how about escaping the hyphen? It's easy to edit regexes with `[]` and not remember that the hyphen needs to go last if it isn't escaped.

millerdev: f1b282e

tests/escapedHashtags.js
```diff
@@ -58,33 +69,32 @@ define([
                     ["#form/text1", "`#form/text1`"],
                     ["/data/text1", "`#form/text1`"],
                     ["`#form/text1`", "`#form/text1`"],
+
+                    ["#form/text1 -1", "`#form/text1` - 1"],
+                    ["/data/text1 -1", "`#form/text1` - 1"],
+                    // ideally no change, but too hard right now (to much extra parsing going on)
```
emord: I think some of that extra parsing was due to attempting to be smart about how we stored the hashtag references in XML. For some reason we stored the hashtags unescaped when it was a valid xpath with known references, and escaped in the case of invalid xpaths. Do you remember the reason we decided to do that? Wouldn't storing escaped all the time be much simpler?

millerdev: Yes. We really only need to escape hashtags when we have an invalid xpath expression. Otherwise escaping is not necessary because the hashtags can be found with normal parsing. Using escaped hashtags all the time might be a bit simpler in vellum, although the escaping logic needs to be used in more places so this is debatable. However, it makes it harder for end-users to hand-edit XML and copy/pasted questions in Excel.

PR #744, opened 2016-09-01
Fix exponential slowdown in XML normalizer

millerdev:
http://manage.dimagi.com/default.asp?236305

Review commits individually.

@emord @orangejenny

### DIFF
--- a/tests/xml.js
+++ b/tests/xml.js
@@ -4,14 +4,16 @@ define([
     'chai',
     'vellum/xml',
     'text!static/xml/regexp-crashing-debug-itext.xml',
-    'text!static/xml/regexp-crashing-debug-itext-parsed.xml'
+    'text!static/xml/regexp-crashing-debug-itext-parsed.xml',
+    'text!static/xml/slow-regexp-itext.xml',
 ], function (
     $,
     _,
     chai,
     xml,
     REGEXP_CRASHING_DEBUG_ITEXT,
-    REGEXP_CRASHING_DEBUG_ITEXT_PARSED
+    REGEXP_CRASHING_DEBUG_ITEXT_PARSED,
+    SLOW_REGEXP_ITEXT
 ) {
     var assert = chai.assert;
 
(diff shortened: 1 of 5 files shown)


tests/xml.js
```diff
@@ -150,6 +152,18 @@ define([
             eq('<output attr=value />', '<output attr="value" />', false);
         });
 
+        it("should fail on escape unquoted attribute value with <", function () {
+            chai.expect(function () {
+                eq('<output value=2<3 />', '<output value="2&lt;3" />', false);
```
emord: What would `normalize` or `humanize` return? or is this just saying that they won't match each other?

millerdev: I don't know if that input is even valid XML for any parser, but it seems like it might be possible to interpret the first string as the second. However, it will not work with our `normalize` (or `humanize` either) because `fixGTBug` can't handle that input. I added these known edge case failures as documentation of things we know will not work.

millerdev: The result of `xml.normalize('<output value=2<3 />')` on phantomjs is

`<output value="2&lt;3"><!--3--></output>`

which is very strange indeed. It's even semantically equivalent XML, which is nice.

tests/xml.js
```diff
@@ -150,6 +152,18 @@ define([
...
+            }).to.throw(Error);
+        });
+
+        it("should fail on escape attribute value with />", function () {
+            chai.expect(function () {
+                eq('<output value="../>.." />', '<output value="..&gt;.." />', false);
```
millerdev: This input is really botched by `fixGTBug`:

`<output value="..&gt;&lt;/output&gt;.." />`

tests/xml.js
emord: Ah. I didn't realize this was an already known error. Should there be a comment on why/what we expect this to fail with an error? It looks like eq has two asserts that could error

millerdev: https://github.com/dimagi/Vellum/pull/748

phillipm:
Woohoo! Excited to see this fixed!

PR #745, opened 2016-09-02
Fix $ escaping

emord:
http://manage.dimagi.com/default.asp?236643#1212896

@millerdev 

buddy @czue

### DIFF
--- a/src/javaRosa/plugin.js
+++ b/src/javaRosa/plugin.js
@@ -397,7 +397,7 @@ define([
                     return '<output\\s*(ref|value)="' + expression + '"\\s*(\/|><\/output)>';
                 } else {
                     // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#Specifying_a_string_as_a_parameter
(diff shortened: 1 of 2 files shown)


src/javaRosa/plugin.js
```diff
@@ -396,7 +396,8 @@ define([
                     expression = RegExp.escape(expression);
                     return '<output\\s*(ref|value)="' + expression + '"\\s*(\/|><\/output)>';
                 } else {
-                    return '<output value="' + expression + '" />';
+                    // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#Specifying_a_string_as_a_parameter
+                    return '<output value="' + expression.replace('$', '$$$$') + '" />';
```
millerdev: This will only replace the first occurrence of `$`. Do you want to replace all occurrences? (use a regexp `/\$/g`)

emord: good point https://github.com/dimagi/Vellum/pull/745/commits/d8b141cee0745485ed004ff5771aac38d6fe76dc

PR #748, opened 2016-09-06
Add comments for known XML normalization failures

millerdev:
Follow-up on #744

@emord


emord:
Thanks for the follow up. Comments only, but will still wait on build

millerdev:
✅

PR #751, opened 2016-09-08
Fix expression editor save button inconsistencies

millerdev:
http://manage.dimagi.com/default.asp?236572

Save button is now
- Disabled when there are no changes.
- Green when expression has changed.

@emord @orangejenny

### DIFF
--- a/src/expressionEditor.js
+++ b/src/expressionEditor.js
@@ -21,7 +21,8 @@ define([
     function showXPathEditor($div, options) {
         var editorContent = $div,
             richTextOptions = {isExpression: true},
-            form = options.mug.form;
+            form = options.mug.form,
+            saveButton;
         options = _.defaults(options, {
             leftPlaceholder: "Drag question here",
             rightPlaceholder: "Drag question here",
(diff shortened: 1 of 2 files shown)


src/expressionEditor.js
```diff
@@ -430,16 +433,17 @@ define([
...
-            $xpathUI.find('.fd-xpath-editor-text').on('change keyup', function (){
+            $xpathUI.find('.fd-xpath-editor-text').on('change keyup', function () {
+                saveButton.removeClass("btn-default disabled").addClass("btn-success");
                 options.change();
             });
 
             var done = function (val) {
-                $div.find('.fd-xpath-editor').addClass("hide");
                 options.done(val);
+                editorContent.empty();
```
millerdev: empty instead of hide to prevent content flash on next expression edit.

src/expressionEditor.js
```diff
@@ -410,7 +412,8 @@ define([
                 tag: tag,
                 tagArgs: tagArgs,
             }));
-            editorContent.empty().append($xpathUI);
+            editorContent.html($xpathUI);
```
millerdev: [`html` is eqivalent to `empty().append`](http://stackoverflow.com/a/1903090/10840)

orangejenny:
nice improvement

PR #752, opened 2016-09-09
test invalid output values

emord:
http://manage.dimagi.com/default.asp?236968#1215641

buddy @proteusvacuum

### DIFF
--- a/src/javaRosa/util.js
+++ b/src/javaRosa/util.js
@@ -270,8 +270,14 @@ define([
                 text = $("<div />").append(text);
                 text.find('output').replaceWith(function() {
                     var $this = $(this),
-                        value = xpathParser.parse($this.attr('value') || $this.attr('ref'));
-                    $this.attr('value', value[functionName]());
+                        value = $this.attr('value') || $this.attr('ref'),
+                        parsedValue;
+                    try {
+                        parsedValue = xpathParser.parse(value);
+                        $this.attr('value', parsedValue[functionName]());
+                    } catch (e) {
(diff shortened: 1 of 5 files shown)


proteusvacuum:
Will this notify the user that there is an error?
(sorry, not that familiar with vellum code)

emord:
this is some bad ux, but we have never had any warnings about invalid xpaths except for in the advanced xpath editor. manual output values are a fairly advanced thing now that easy references exist anyways.

proteusvacuum:
👍

src/javaRosa/util.js
```diff
@@ -270,8 +270,14 @@ define([
...
                     var $this = $(this),
-                        value = xpathParser.parse($this.attr('value') || $this.attr('ref'));
-                    $this.attr('value', value[functionName]());
+                        value = $this.attr('value') || $this.attr('ref'),
+                        parsedValue;
+                    try {
+                        parsedValue = xpathParser.parse(value);
+                        $this.attr('value', parsedValue[functionName]());
+                    } catch (e) {
+                        $this.attr('value', value);
```
millerdev: Should this use the `#invalid/xpath ` prefix?

emord: I think that should only be added to attributes with a `vellum:` namespace correct? We could add a `vellum:value` to this output value, but I don't think that's necessary.

millerdev: I guess that means you can't use hashtags (like `#case/foo`) in typed output value expressions?

PR #756, opened 2016-09-15
upgrade jstree & fuse

emord:
because library upgrades are so fun.

I did a sanity check locally and everything looks ok


orangejenny:
atwho test timed out, going to assume that was a coincidence & kick build

emord:
:white_check_mark: 

> atwho test timed out

weird. the longest it took on the passing test was 117ms

PR #759, opened 2016-09-16
Fix datasources state

millerdev:
http://manage.dimagi.com/default.asp?237032

This bug was really hard to figure out and fix because of a timing issue between loading of the data browser and data sources. This is a major refactor of how the form object gets a list of valid hashtags from data sources. Hashtags and hashtag transformations are now available immediately when data sources are loaded rather than after the data browser is initialized. The data browser now does only one thing: display a tree of case properties, and it no longer mutates the form during its initialization phase.

A few other cleanup items were done to make it easier to think about the state of vellum in tests:
- data sources state is now local to a vellum instance rather than global
- the data browser no longer has global state.

Probably easiest to review commits individually. 00af9dd was a stab at making the commented-out test in it pass, but I got exhausted trying to figure out where hashtags should be used and where they should not be. Renaming questions is finicky and might even be broken in some cases when rich text is disabled (although that has not changed in this PR).

@emord

### DIFF
--- a/src/databrowser.js
+++ b/src/databrowser.js
@@ -4,7 +4,6 @@
 define([
     'jquery',
     'underscore',
-    'vellum/datasources',
     'vellum/util',
     'vellum/widgets',
     'vellum/window',
@@ -12,24 +11,13 @@ define([
 ], function (
     $,
     _,
-    datasources,
     util,
     widgets,
     window_,
     external_sources_tree
 ) {
     var fn = {},
-        DATABROWSER_HEIGHT = 0.33,
-        panelHeight,
-        handleError = function($container) {
-            return function(jqXHR, textStatus, errorThrown) {
-                if ($container && jqXHR.responseText) {
-                    $container.find(".fd-external-sources-error").removeClass("hide").text(jqXHR.responseText);
-                } else {
-                    window.console.log(util.formatExc(textStatus || errorThrown));
-                }
-            };
-        };
+        DATABROWSER_HEIGHT = 0.33;
 
     // plugin adds an item to the Tools menu when enabled
     $.vellum.plugin('databrowser', {}, {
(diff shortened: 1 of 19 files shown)


millerdev:
> Do you think that there will ever be a meaningful difference between `vellum` and `form`?

I think the meaningful difference right now is that (except for in tests) we have one vellum instance for the lifetime of the page, while a new `form` is created each time new XML is submitted (through the *Edit Source XML* dialog). Granted, there is only ever one *active* `form` instance at a time, but I still think it's a useful abstraction. The `vellum` instance holds values like data sources, UI elements, and other state that do not need to be re-fetched or rebuilt when a new `form` instance is created.

src/databrowser.js
```diff
@@ -38,6 +26,12 @@ define([
                 tree = this.$f.find(".fd-tree"),
                 pane = this.$f.find(".fd-accessory-pane"),
                 head, headHeight, searchBar, paneRatio;
+            vellum.data.core.databrowser = {
+                dataHashtags: {},
+                dataHashtagTransformations: {},
+                errorContainer: pane,
```
millerdev: I suppose it could be, but I'm not really interested in changing that in this PR. Doesn't seem too high priority.

src/itemset.js
```diff
@@ -375,16 +373,12 @@ define([
...
-        // HACK synchronously get asynchronously loaded data (if available)
-        var sources = [];
-        datasources.getDataSources(function (data) {
-            // will execute synchronously if data sources have been loaded
-            if (opts.dataSourcesFilter) {
-                data = opts.dataSourcesFilter(data);
-            }
-            sources = data;
-        });
+    function getDataSources(mug) {
```
millerdev: I thought about that, but decided against it because it adds duplication at every call site (`mug` -> `mug.form.vellum`). I suppose you could argue that [line 183](https://github.com/dimagi/Vellum/pull/759/files/00af9dd296c945644536b345d36c205ec6927a5c#diff-f15dee15f58a804357f7fca0280de97dR183) could have been simplified since `form` is local there, but it's not even used on [line 179](https://github.com/dimagi/Vellum/pull/759/files/00af9dd296c945644536b345d36c205ec6927a5c#diff-f15dee15f58a804357f7fca0280de97dR179) and I was hesitant to let the scope of changes in this PR creep too far.

PR #761, opened 2016-09-20
Fix HTML escaping in question tree

millerdev:
http://manage.dimagi.com/default.asp?236328

@orangejenny

### DIFF
--- a/src/javaRosa/util.js
+++ b/src/javaRosa/util.js
@@ -265,7 +265,7 @@ define([
     };
 
     function _outputToXPathOrHashtag (functionName) {
-        return function (text, xpathParser) {
+        return function (text, xpathParser, escape) {
             if (text) {
                 text = $("<div />").append(text);
                 text.find('output').replaceWith(function() {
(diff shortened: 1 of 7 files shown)


src/javaRosa/util.js
```diff
@@ -281,6 +281,11 @@ define([
                     return $this[0].outerHTML;
                 });
                 text = xml.normalize(text.html());
+                if (escape) {
+                    text = text.replace(/(<)|>/g, function (match, lt) {
+                        return lt ? "&lt;" : "&gt;";
+                    });
```
millerdev: `&` chars have already been converted to `&amp;` above on [line 270](https://github.com/dimagi/Vellum/pull/761/files#diff-16525243359147bd8661fbb97da09528R270)

orangejenny: I like doing clever things with regexes, but also think that doing two plain string replacements would be slightly more understandable.

orangejenny:
Thank you for grabbing this. Escaping-related bugs are rotten.

millerdev:
✅

src/javaRosa/util.js
millerdev: Replacing multiple occurrences of a string in JS requires a regexp (with the `g` flag).

orangejenny: aha, makes sense

PR #766, opened 2016-10-03
Slash data

emord:
@millerdev saw you had a comment on slack. I didn't change the save to case thing as that is a feature flag and I haven't looked into if HQ always uses `/data/meta/timeEnd` or if it changes it to `/rootid/meta/timeEnd`

### DIFF
--- a/src/form.js
+++ b/src/form.js
@@ -134,6 +134,8 @@ define([
         this._updateHashtags();
 
         this.tree = new Tree('data', 'control');
+        // initalize #form as /data
+        // if data is not root node, it will be changed later
         this.addHashtag('#form', '/data');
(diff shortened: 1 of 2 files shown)


src/form.js
```diff
@@ -134,6 +134,8 @@ define([
         this._updateHashtags();
 
         this.tree = new Tree('data', 'control');
+        // initalize #form as /data
+        // if data is not root node, it will be changed later
```
millerdev: Can we add a test to verify that this is true?

emord: I think there may already be one, although it may not be an explicit test. I'll do some digging later to find one/make a new one

PR #767, opened 2016-10-05
hide undo after adding a question

emord:
This fixes half of http://manage.dimagi.com/default.asp?238232, and will probably solve most of the cases of this bug popping up again.

At the moment, it's a little unclear to me how best to have the undo manager check the node id for conflicting nodes. Right now that code is in `moveMug` and undo uses `insertMug` to insert. I want to think a little harder before moving/duplicating that code into `insertMug` or changing the way the undo puts mugs into its queue

@millerdev @orangejenny buddy @czue

### DIFF
--- a/src/form.js
+++ b/src/form.js
@@ -154,6 +154,8 @@ define([
         util.eventuality(this);
         this.on('form-load-finished', function() {
             _this.fuse = new Fuse(_this);
+        }).on('question-create', function () {
+            _this.undomanager.resetUndo();
         });
         this.disconnectDataSources = function () {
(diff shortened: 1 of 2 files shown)


emord:
I think it's better from a UX perspective. It seems pretty weird to have changes in the question tree, but keep around an "undo" action that is from a long time ago. Without this change this happens:

1. create question1 and question2
2. delete question1
3. create question3
4. hitting undo recreates question1 instead of deleting question3

Preferably undo would be more robust, but it's not a trivial problem to solve in vellum at the moment

PR #768, opened 2016-10-05
Add explicit test for #form transform

emord:
@millerdev as promised https://github.com/dimagi/Vellum/pull/766/files#r81555334

buddy @czue


millerdev:
Kicked build, which failed with
```
  1) atwho without rich text should not show mugs without absolutePath:
```

millerdev:
thanks, @emord!

PR #770, opened 2016-10-06
get rid of double scroll bar in full screen mode

orangejenny:
http://manage.dimagi.com/default.asp?236113

### DIFF
--- a/src/less-style/structure.less
+++ b/src/less-style/structure.less
@@ -1,6 +1,10 @@
 // Styling for the vellum container and its main content blocks: question tree, question properties panel, etc.
 @import "lib/main";
 
+body.vellum-full-screen {
+    overflow-y: hidden;
+}
+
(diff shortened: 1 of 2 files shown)


src/less-style/structure.less
```diff
@@ -1,6 +1,10 @@
 // Styling for the vellum container and its main content blocks: question tree, question properties panel, etc.
 @import "lib/main";
 
+body.vellum-full-screen {
+    overflow-y: hidden;
```
millerdev: gutsy move - hope there's nothing important in the overflow 🙏

orangejenny: Yeah, this does make me a little nervous, but I think it's logical since fullscreen makes vellum fixed position and screen height anyway. I'm pretty sure the below-the-fold content isn't really content, it's just that a bunch of elements in HQ's layout have their height explicitly set taller than the screen.

PR #772, opened 2016-10-07
Update codemirror

emord:
Upgrades codemirror to newest version. This should allow us to share codemirror with HQ later. which will be useful since the app manager now loads that.

buddy @czue


src/main.js
```diff
@@ -30,13 +30,16 @@ requirejs.config({
             name: 'json',
             location: '../bower_components/requirejs-plugins/src',
             main: 'json.js'
-        }
+        },
+        {
+            name: "codemirror",
```
emord: http://codemirror.net/doc/manual.html#modloader

czue:
build is failing, not sure if related

emord:
Hmm. yeah it's somehow related to the build changes I made.The tests pass on the unminified version of vellum, but not once it's minified, but it's not obvious to me why these tests fail as there's no dependency for code mirror there.

@orangejenny @millerdev any requirejs insight as to why this would occur?

src/core.js
```diff
@@ -56,14 +59,14 @@ define([
...
     jrUtil,
+    codemirror,
     debug
 ) {
     
     // Load these modules in the background after all runtime dependencies have
     // been resolved, since they're not needed initially.
     setTimeout(function () {
         require([
-            'codemirror',
```
millerdev: Was it not possible to load the `codemirror` dependencies here rather than above in the main code block?

src/main.js
```diff
@@ -30,13 +30,16 @@ requirejs.config({
...
+            location: "../bower_components/codemirror",
+            main: "lib/codemirror",
+        },
     ],
     paths: {
         'vellum': '.',
 
-        'codemirror': '../lib/codemirror/xml',
-        'codemirrorBase': '../bower_components/codemirror/lib/codemirror',
```
millerdev: Oh, didn't see your comment before I posted my review. My guess is that you need to replace these removed lines with something similar, but pointing to the new version.

bower.json
```diff
@@ -41,7 +41,7 @@
         "xpath": "dimagi/js-xpath#v0.0.3",
         "jsdiff": ">=1",
         "markdown-it": "~7.0.0",
-        "codemirror": "3.24.0"
+        "codemirror": "5.19.0"
```
emord: 5.20.2 is out, but the owners have taken a stance against bower https://github.com/codemirror/CodeMirror/issues/4328

index.html
```diff
@@ -7,6 +7,7 @@
     <link href="tests/lib/font-awesome.css" type="text/css" rel="stylesheet" />
     <link href="tests/lib/font-commcare.css" type="text/css" rel="stylesheet" />
     <link href="tests/lib/hqstyle-core.css" type="text/css" rel="stylesheet"></link>
+    <link href="bower_components/codemirror/lib/codemirror.css" type="text/css" rel="stylesheet" />
```
emord: This is already included in the build file, so it will get minified with other package's css. Previously this lived in src/main.js, but since we can no longer use paths for codemirror I needed to pull this out

src/core.js
emord: Yep. I was able to once I figured out the other issues with the build https://github.com/dimagi/Vellum/pull/772/commits/b3bca8a4936b7404da12579f4e26cc38de18423e

bower.json
millerdev: well that's sad :(

PR #773, opened 2016-10-11
Fix lookup table menu

millerdev:
http://manage.dimagi.com/default.asp?238911

Includes https://github.com/ichord/At.js/pull/438 via [millerdev/At.js#master](https://github.com/millerdev/At.js/tree/master)

A lot of this is cleanup that I did while troubleshooting the bug. I had been putting off blaming the problem on At.js until I eliminated everything else that could have caused the issue.

Best to review each commit individually.

@emord

### DIFF
--- a/src/richText.js
+++ b/src/richText.js
@@ -196,6 +196,7 @@ define([
                     },
                     noSnapshot: true,
                 });
+                return wrapper;
             },
             insertExpression: function (xpath) {
                 if (options.isExpression) {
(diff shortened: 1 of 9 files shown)


src/richText.js
```diff
@@ -205,19 +206,27 @@ define([
...
             },
             insertOutput: function (xpath) {
                 if (options.isExpression) {
                     throw new Error("cannot insert output into expression editor");
                 }
                 insertHtmlWithSpace(bubbleOutputs(xpath, form));
+                return wrapper;
+            },
+            change: function () {
+                editor.fire("saveSnapshot");
```
emord: Why did you choose `saveSnapshot` instead of `change` ? https://github.com/dimagi/Vellum/pull/680

PR #774, opened 2016-10-11
Multiple modals

orangejenny:
http://manage.dimagi.com/default.asp?238845

Closing one modal and then immediately opening another one doesn't play nicely with the fade transition. This looks like an issue with bootstrap, but an easy one to work around by turning off the transition in the case where we're closing a modal because we're about to open another one.

@millerdev

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -702,6 +702,7 @@ define([
             });
         }
         if (immediate) {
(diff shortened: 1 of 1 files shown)


millerdev:
wow, that was quick! thanks!

PR #778, opened 2016-10-26
jQuery 3

orangejenny:
Opening for tests. Merge when https://github.com/dimagi/commcare-hq/pull/13736 is ready.


orangejenny:
QA passed for https://github.com/dimagi/commcare-hq/pull/13736, so this is ready to merge. There's no technical reason they have to be deployed at the same time, but we might as well.

PR #779, opened 2016-10-27
Highlight easy reference in tree

orangejenny:
Experiment discussed with @amsagoff  So that we can determine whether or not to keep this as a feature, it adds events that record when people a) see the popover and b) click on the link in the popover.

jeqrnmkopo

cc @calellowitz

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -224,6 +224,7 @@ define([
                 var nodeTop = $node.position().top;
                 $scrollable.scrollTop($node.position().top - $scrollable.position().top);
             }
+            console.log("fire click events");
         });
 
         this._init_toolbar();
--- a/src/core.js
+++ b/src/core.js
@@ -224,7 +224,10 @@ define([
                 var nodeTop = $node.position().top;
                 $scrollable.scrollTop($node.position().top - $scrollable.position().top);
             }
-            console.log("fire click events");
+            if (window.analytics) {
+                window.analytics.usage("Form Builder", "Clicked link to show in tree");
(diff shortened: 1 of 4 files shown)


czue:
now that we have "go to declaration" we just need to add "find usages" :smiling_imp:

src/richText.js
```diff
@@ -649,12 +649,31 @@ define([
                             title: '<h3>' + util.escape(displayId) + '</h3>' +
                                    '<div class="text-muted">' + util.escape(widget.mug.form.normalizeHashtag(xpath)) + '</div>',
                             html: true,
-                            content: '<p>' + labelText.text() + '</p>',
+                            content: '<p>' + labelText.text() + '</p>' +
```
emord: This may be large enough for an underscore template

src/core.js
```diff
@@ -215,6 +215,20 @@ define([
...
 
+        $(document).on('click', '.jstree-hover', function(e) {
+            e.preventDefault();
+            _this.jstree("hover_node", $(this).data("ufid"));
+            var $node = $(".jstree-hovered");
+            if ($node.length) {
+                var $scrollable = $node.closest(".fd-scrollable");
+                $scrollable.scrollTop($node.position().top - $scrollable.position().top);
+            }
+            if (window.analytics) {
```
emord: not immediately actionable, but it feels like these analytics calls are being littered across the code. I wonder if its worth it to have it in a plugin or as an `eventuality` that we can call?

PR #782, opened 2016-10-28
update image in readme

emord:
Same form, just updated with new UI.

now with more panes but same number of scroll bars


orangejenny:
so much prettier

PR #784, opened 2016-10-28
Move full screen button into tools menu

orangejenny:
Don't think this is important enough to be always visible, and user testing has pushed me to want to reduce clutter in vellum. @dimagi/product feel free to register an objection.

before
screen shot 2016-10-28 at 5 58 43 pm

after
screen shot 2016-10-28 at 5 57 44 pm
screen shot 2016-10-28 at 5 57 57 pm

cc @calellowitz

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -168,29 +168,6 @@ define([
             unsavedMessage: 'Are you sure you want to exit? All unsaved changes will be lost!',
             csrftoken: _this.opts().csrftoken
         });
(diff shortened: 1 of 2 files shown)


millerdev:
Firefox and Chrome both say "Enter Full Screen" although I'd be onboard with "Full screen" and "Exit full screen" (despite it being full browser and not full screen).

orangejenny:
updated to "Enter/Exit Full Screen"

PR #785, opened 2016-10-28
Make collapsing decisions sticky

orangejenny:
In question properties, collapse all sections by default except for Basic. Then, whenever someone collapses or expands a section, remember that decision and respect it in the future.

Quick and dirty and testless, since the failure mode of this is not a big deal.

cc @calellowitz

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -423,6 +423,13 @@ define([
                 }
             });
         });
+
+        this.$f.on('show.bs.collapse hide.bs.collapse', function(e) {
+            var $target = $(e.target),
+                section = $target.parent().find("legend").text().trim(),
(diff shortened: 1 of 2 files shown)


orangejenny:
@emord

orangejenny:
fyi @dimagi/product

orangejenny:
> Does this need sign off from @dimagi/product ?

Wouldn't hurt. I don't think Amelia ever weighed in on the slack conversation.

amsagoff:
how does "remembering the decision" work? Is it per-session?

orangejenny:
It's remembered forever, but is specific to the user's browser.

amsagoff:
Nice!

PR #786, opened 2016-11-01
Fix copy bubble from label to expression input

millerdev:
http://manage.dimagi.com/default.asp?237816

Includes a major refactor of escaped hashtags: they're now only used inside the rich text widget and when writing invalid expressions in XML. Best to review commits individually 🐠 the commit messages are descriptive.

@emord

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1159,6 +1159,7 @@ define([
                 }
                 hidePageSpinner();
             } catch (e) {
+                window.console.log(util.formatExc(e));
                 // hack: don't display the whole invalid XML block if it
                 // was a parse error
                 var msg = e.toString();
--- a/tests/richText.js
+++ b/tests/richText.js
@@ -104,25 +104,21 @@ define([
     function externalIcon () { return icon('fcc-fd-case-property'); }
     function externalUnknownIcon () { return icon('fa-exclamation-triangle'); }
 
-    function bubbleSpan(xpath, internal, output) {
+    function makeBubble(xpath, dispValue, icon, internal, output) {
         var span = $('<span>').addClass('label label-datanode').attr({
             'data-value': xpath,
-            'data-output-value': output,
+            'data-output-value': output || false,
         });
-        if (internal) {
+        if (internal && !_.isString(internal)) {
             span.addClass('label-datanode-internal');
-        } else if (form.isValidHashtag(xpath)) {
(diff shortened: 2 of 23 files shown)


tests/richText.js
```diff
@@ -406,44 +431,80 @@ define([
...
-                            editor.insertExpression('/data/text');
-                            assert.equal(editor.getValue(), result);
-                        } finally {
-                            options.isExpression = false;
+                    exprEditor.setValue(expr, function () {
+                        assert.equal(exprEditor.getValue(), expr);
+                        exprEditor.select(i);
+                        exprEditor.insertExpression('#form/text');
+                        if (expr === "one two" || i === 3) {
+                            // BUG for some reason ckeditor.getData() replaces
```
emord: I think this may be a "feature" of ckeditor. It's intended to be an html editor, not a text editor so spaces aren't considered significant.

src/core.js
```diff
@@ -1159,6 +1159,7 @@ define([
                 }
                 hidePageSpinner();
             } catch (e) {
+                window.console.log(util.formatExc(e));
```
emord: unsure if you meant to keep this in here, although I don't think it's a bad idea to have

emord:
Nothing blocking on this PR, so going to go ahead and merge

src/core.js
millerdev: Yes, intended to keep it. Otherwise it's hard to debug errors in form loading logic.

millerdev:
> How difficult do you think it would be to have the date formatting popup show when you copy from an xpath to a label? It will currently not add any formatting so it will just show the raw number in the label, which could be confusing.

This is true, it could be confusing. Probably a good idea to show the date format in the popover. I'll make a separate PR for that.

tests/richText.js
emord: Agreed that if a "text editor" removed whitespace it would be classified as a bug.

I think the problem is that ckeditor/contenteditable divs shouldn't be thought of as a "text editor" and more like a "word processor". By introducing Easy References, I think we got away from the notion of a "text editor". Not to say that that's what is the best, but that it's a limitation of moving away from pure text.

Doing a quick test with display conditions, it seems that inside a string, the extra whitespaces are kept between loads, so I don't think it's a functional issue (shouldn't affect the evaluation of the xpath), just a visual one?

PR #787, opened 2016-11-02
find usages poc

emord:
Been a while since I PID'd something in vellum

@dannyroberts This is from your field report. Pushed it into the tools menu with other advanced features.

Obviously could be improved a number of ways, but wanted to do this in a way that didn't add any new dependencies and I don't think we have any libraries to display this better at the moment.

I'd also like to be able to do it per question, but don't have a way to do it now without cluttering the UI even more

image

cc: @dimagi/product @orangejenny @millerdev 
buddy @esoergel


dannyroberts:
neat :)

orangejenny:
This is glorious. Brace yourself for scope creep.

src/core.js
```diff
@@ -710,6 +716,61 @@ define([
...
 
+    fn.findUsages = function () {
+        var _this = this,
+            $modal = this.generateNewModal("Use of each question", []),
+            $modalBody = $modal.find('.modal-body'),
+            form = _this.data.core.form,
+            logicManager = form._logicManager,
+            $table = $('<table>'),
+            $thead = $('<thead>'),
+            propertyToPerson = {
```
orangejenny: Can the human-readable names for these be pulled out of the mug specs?

orangejenny:
How about doing one row for `Label` rather than one row per language?
screen shot 2016-11-02 at 5 56 39 pm

orangejenny:
> I'd also like to be able to do it per question, but don't have a way to do it now without cluttering the UI even more

To trigger the modal? I could see a button likely buried in `Logic` or `Advanced`. Not ideal, but it would allow per-question display, which I suspect would be a big improvement.

screen shot 2016-11-02 at 6 18 28 pm

Not blocking, the current implementation seems like a reasonable first step.

millerdev:
> I could see a button likely buried in Logic or Advanced. Not ideal, but it would allow per-question display, which I suspect would be a big improvement.

I second this idea. Logic seems like a logical place 👅

millerdev:
This is cool! I had imagined it taking a lot more code.

czue:
sweet! i bet @proteusvacuum and @phillipm will be excited too

src/core.js
emord: https://github.com/dimagi/Vellum/pull/787/commits/22aa904631cc2405b791d974d33cbfc849157e2d

emord:
Moved to an underscore template so now even less code :) 

I think this is good to go. Everyone seems to be on board for the first pass. Next obvious step would be to do this on specific question. I imagine it will look something like https://github.com/dimagi/Vellum/commit/3d6811b352a138d4b51d18170b1c181b445fafe6 but it needs a widget to support a button to use @orangejenny ui.

amsagoff:
I like the idea of releasing in tools and tracking usage - if users are
loving it we can make it more prominent. Have we added analytics to this
yet?

emord:
@amsagoff analytics added

phillipm:
Ooooo, very exciting!!
Can you click on the row and be taken to the usage?!

proteusvacuum:
Nice!

emord:
> Can you click on the row and be taken to the usage?!

@phillipm sounds like another good improvement, but that's not part of this PR

PR #788, opened 2016-11-02
Move question ID into logic section

orangejenny:
@amsagoff green-lighted this

screen shot 2016-11-02 at 5 41 58 pm

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1976,10 +1976,8 @@ define([
                 properties: this.getMainProperties(),
                 help: {
                     title: "Basic",
-                    text: "<p>The <strong>Question ID</strong> is an internal identifier for a question. " +
-                        "It does not appear on the phone. It is the name of the question in data exports.</p>" +
-                        "<p>The <strong>Label</strong> is text that appears in the application. " +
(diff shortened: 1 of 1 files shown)


amsagoff:
Hold off on merge til we've figured out the docs - but I managed to get the
scope of the doc change down, so we can move forward on this in general!

amsagoff:
Also checking - does ID get auto-filled?

orangejenny:
It's initially set to `question1`, `question2`, etc. Nothing updates it after that.

czue:
have we thought about, and decided that we don't care about the segment of people that might run into trouble analyzing their data if they don't realize that they should set proper question IDs ahead of time? just wondering if we should get more input on that question before releasing this.

amsagoff:
I don't think we've properly thought through the implications of defaulting
question IDs to something non-meaningful. Jenny, let's regroup on this
offline?

orangejenny:
Sure, sounds good.

amsagoff:
(i think this is being addressed through form builder onboarding now)

orangejenny:
yeah, closing this

PR #789, opened 2016-11-02
Icons for tools menu

orangejenny:
The tools menu is getting lengthy, especially with https://github.com/dimagi/Vellum/pull/784 and https://github.com/dimagi/Vellum/pull/787  And we apparently like names that start with "E" or "F" which makes it harder to skim.

before
screen shot 2016-11-02 at 6 56 48 pm

after
screen shot 2016-11-02 at 6 54 18 pm

Marking don't merge because in standalone I don't see all of the images and would like to figure out why that is:
screen shot 2016-11-02 at 6 54 10 pm


millerdev:
Excellent! Good to go once you get the standalone issue fixed.

orangejenny:
Updated fonts to fix standalone, this is good to go.

PR #791, opened 2016-11-03
Je/mug specific find usage

emord:
Would get merged after #787 

Used @orangejenny UI from that PR but didn't want to tack it on as we no one had approved it explicitly.

image

### DIFF
--- a/src/widgets.js
+++ b/src/widgets.js
@@ -387,6 +387,34 @@ define([
         return widget;
     };
 
+    var button = function (mug, options) {
+        var widget = normal(mug, options),
+            disabled = options.disabled || false;
+        widget.input = $("<button />")
+            .addClass('btn btn-default')
+            .attr('type', 'button')
+            .attr("name", widget.id)
+            .attr("id", widget.id)
+            .prop('disabled', disabled)
+            .html(options.buttonContent);
+
+        widget.setValue = function (value) { };
+
+        widget.getValue = function() { return null; };
+
+        return widget;
+    };
+
+    var findUsagesButton = function (mug, options) {
+        var widget = button(mug, options);
+
+        widget.input.click(function() {
+            mug.form.vellum.findUsages(mug.hashtagPath);
+        });
+
+        return widget;
+    };
+
     var xPath = function (mug, options) {
         options.disableNativeSpellChecker = true;
         var widget = richInput(mug, options),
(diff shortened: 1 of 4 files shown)


src/widgets.js
```diff
@@ -387,6 +387,37 @@ define([
...
+
+        return widget;
+    };
+
+    var findUsagesButton = function (mug, options) {
+        var widget = button(mug, options);
+
+        widget.input.click(function() {
+            if (window.analytics) {
+                window.analytics.usage('Form Builder', 'Logic', 'Find Usages');
```
emord: @amsagoff analytics for if you click find usages within the question

amsagoff:
In general, I'd like to be more confident of the value and usability before
making it prominent in the UI. So we might want to put it through some
product processes or organize this as a "test" with an end date if we want
to release it here.

emord:
@amsagoff not sure if this was obvious, but it's slightly different than the one in the tools menu. This would only find usages of this particular question, which I think could be more useful than the one hidden in the menu. Do you think putting this in advanced would be better? or maybe something like:

image

src/widgets.js
```diff
@@ -387,6 +387,37 @@ define([
...
+            .html(options.buttonContent);
+
+        widget.setValue = function (value) { };
+
+        widget.getValue = function() { return null; };
+
+        return widget;
+    };
+
+    var findUsagesButton = function (mug, options) {
```
orangejenny: Surprised to see the button's action inside of the widget, I'd expected something like a generic button widget that accepts both a `buttonAction` function and `buttonContent`. On the other hand, there's not much point making this widget more generic until we have more than one use case for it.

orangejenny:
Code looks reasonable.

amsagoff:
I like Emord's idea of putting it in the menu settings for now. Or we can
put it on the main screen, but would want to frame it as a test (similar to
the "show in tree" test you're running, Jenny)

orangejenny:
Modal layout could be simpler:
screen shot 2016-11-07 at 9 36 28 am

emord:
@orangejenny good idea https://github.com/dimagi/Vellum/pull/791/commits/410a3ab6c6bb1e5a42d135ca6157cb27e9742a35

millerdev:
Looks good. @emord, are you planning to put links in the specific mug usage template as in https://github.com/dimagi/Vellum/pull/795?

What do you think of inserting the specific mug usage table inline (show/hide on clicking the button) rather than in a modal? The advantage being that the tree would not be obscured by the modal.

emord:
Doubt this will get merged in current state and I think most of the functionality is in the current implementation of find usages now. Closing out

PR #792, opened 2016-11-03
Show date format in popover

millerdev:
As suggested in https://github.com/dimagi/Vellum/pull/786

@emord


millerdev:
@dimagi/product 

no formatting on the first one
screenshot 2016-11-03 14 19 21

screenshot 2016-11-03 14 19 44

screenshot 2016-11-03 14 20 02

emord:
Originally I was thinking about maybe having an example in the popover? I think it would take me a bit to figure out what (ddd, mmm d, yyy) means. Would be extra nice if we could say something like "Today's date with this format: Thu Nov 3, 2016" or something, but not sure what would be the best place to put that kind of thing.

That said this makes dates a little more transparent to the end user so its fine with me

amsagoff:
To Emord's point, I can't figure out what (ddd, mmm, d, yyy) means. And now
I have "mmmbop" stuck in my head...

I wonder whether we're trying to get too fancy here, especially with live
preview coming?

millerdev:
> Originally I was thinking about maybe having an example in the popover?

The menu you choose from when dragging/dropping a date into a label shows examples. Adding more text to the popover seems like it might be too noisy. I think that's good enough:
screenshot 2016-11-03 14 28 33

millerdev:
Update:

screenshot 2016-11-03 17 10 29
screenshot 2016-11-03 17 21 48
Click format link in popover to choose new date format:
screenshot 2016-11-03 17 11 57

Non-date bubbles do not show date format selector:
screenshot 2016-11-03 17 12 20

emord:
Will let @orangejenny or @dimagi/product  take over merging this for UI/UX

Code looks good

orangejenny:
I like the link that lets you change the format from the popup, but don't think `d/m/yy` is comprehensible, would replace it with an example date.

> I wonder whether we're trying to get too fancy here

Maybe. I'm going to leave this to @millerdev and @amsagoff to negotiate.

amsagoff:
Did this come up for any specific reason/based on any feedback? I wonder if
there's a case of YAGNI happening...

emord:
@amsagoff I think I just noticed that once you selected a format there was no way to view that without diving into xml. I don't think there have been any users that reported it, it just seemed very opaque to me

millerdev:
> don't think `d/m/yy` is comprehensible, would replace it with an example date.

The examples are in the menu after dragging a date into a label or when clicking the format in the popover. Are you saying those are not enough?

millerdev:
@amsagoff I think this is a fairly non-intrusive feature, and it seems necessary since the format is not easily known or changed, and most likely not what you want, when copying a bubble from an expression to a label. It's also hard to know the format of a date in a label created by someone else or a long time ago.

I can imagine removing it once we have the ability to edit the expression inside the bubble, but that seems a while off yet.

orangejenny:
> The examples are in the menu after dragging a date into a label or when clicking the format in the popover. Are you saying those are not enough?

Yes, I think people will forget between the time they drag a date in (if they even read it then) and the time they see it in the popover.

millerdev:
@orangejenny I see what you're getting at, but I think putting today's date or a hard-coded date in there would be worse. You can easily click the link in the popover to find out what it means and see examples. Clicking the link to learn more does not commit you to changing the format.

orangejenny:
Hmm...I do agree that putting in an example still isn't exactly clear enough, and adding more text gets long/cluttered, and the explanation is easy to access. Don't think I have a better but still concise suggestion. I'm okay with it as-is.

amsagoff:
Went to voice with Daniel, this is good to go from product perspective (didn't realize this now included the ability to change the date formatting - very cool!)

millerdev:
resolved conflicts. wait for tests

PR #794, opened 2016-11-04
Je/refactor find usages

emord:
https://github.com/dimagi/Vellum/commit/3d6811b352a138d4b51d18170b1c181b445fafe6#commitcomment-19687675

classic @millerdev using refactoring as a ruse to make it easier to test :P 

@esoergel

### DIFF
--- a/src/logic.js
+++ b/src/logic.js
@@ -517,18 +517,18 @@ define([
             var _this = this,
                 form = _this.form,
                 tableData = {},
-                refs = _this.reverse;
+                formRefs = _this.reverse;
 
-            _.each(refs, function (value, key) {
-                var usedMug = form.getMugByUFID(key),
+            _.each(formRefs, function (refsToUsedMug, usedMugUfid) {
+                var usedMug = form.getMugByUFID(usedMugUfid),
                     mugReferences = {};
                 if (path && path !== usedMug.hashtagPath) {
                     return;
                 }
-                _.each(value, function (value, key) {
-                    _.each(value, function (value) {
-                        var usedInMug = form.getMugByUFID(key),
-                            readablePropName = usedInMug.spec[value.property].lstring;
+                _.each(refsToUsedMug, function (refs, usedInMugUfid) {
+                    _.each(refs, function (ref) {
+                        var usedInMug = form.getMugByUFID(usedInMugUfid),
+                            readablePropName = usedInMug.spec[ref.property].lstring;
                         mugReferences[usedInMug.hashtagPath] = readablePropName;
                     });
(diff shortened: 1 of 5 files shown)


src/logic.js
```diff
@@ -501,7 +501,46 @@ define([
...
+            _.each(refs, function (value, key) {
+                var usedMug = form.getMugByUFID(key),
+                    mugReferences = {};
+                if (ufidOrPath && ufidOrPath === usedMug.hashtagPath) {
+                    return;
+                }
+                _.each(value, function (value, key) {
+                    _.each(value, function (value) {
+                        var usedInMug = form.getMugByUFID(key),
+                            readablePropName = usedInMug.spec[value.property].lstring;
```
emord: Arguably should return `value.property` and the modal code should change it to a readable name, but I think this is good enough

millerdev:
what else do you use refactoring for?

src/logic.js
```diff
@@ -501,7 +501,42 @@ define([
...
+                refs = _this.reverse;
+
+            _.each(refs, function (value, key) {
+                var usedMug = form.getMugByUFID(key),
+                    mugReferences = {};
+                if (path && path !== usedMug.hashtagPath) {
+                    return;
+                }
+                _.each(value, function (value, key) {
+                    _.each(value, function (value) {
```
millerdev: using `value` ~~twice~~ thrice in nested scopes like this makes it hard to follow. More descriptive names would be nice. Also, same for `key`.

tests/logic.js
```diff
@@ -204,6 +204,33 @@ define([
...
+            before(function (done) {
+                form = util.loadXML(TEST_XML_1);
+                done();
+            });
+
+            it("should return dictionary of usages", function () {
+                var expectedOutput = {
+                        "#form/question1": {"#form/question2": "Display Condition"}
+                    };
+                assert.deepEqual(form.findUsages(), expectedOutput);
```
millerdev: Maybe load a form with more usages so the expected output of this test is not the same as the one that filters on a path?

src/logic.js
emord: slightly better naming although I think it will still take me a bit to figure out what's going on when I come back to it https://github.com/dimagi/Vellum/pull/794/commits/e476b06bea4390dd0bbbf13b901ef20c6aec3108

tests/logic.js
emord: https://github.com/dimagi/Vellum/pull/794/commits/bf04c1c0cdf20408ec7c69c508401c1da7908a8d

PR #795, opened 2016-11-04
Follow click on table to reference

emord:
@phillipm 

https://github.com/dimagi/Vellum/pull/787#issuecomment-258323576

This enables clicking on one of the references displayed in the find usages table and going to the question that references that.

buddy @esoergel

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -750,8 +750,8 @@ define([
 
         $modalBody.append($(find_usages({tableData: tableData})));
 
-        $modalBody.find('.usage-row').click(function() {
-            var gotomug = $(this).find('.used-in-mug').text();
+        $modalBody.find('.link-to-question').click(function() {
+            var gotomug = $(this).text();
             $modal.modal('hide');
             _this.setCurrentMug(form.getMugByPath(gotomug));
         });
--- a/src/core.js
+++ b/src/core.js
@@ -751,9 +751,10 @@ define([
         $modalBody.append($(find_usages({tableData: tableData})));
 
         $modalBody.find('.link-to-question').click(function() {
-            var gotomug = $(this).text();
+            var goToMug = $(this).text();
             $modal.modal('hide');
-            _this.setCurrentMug(form.getMugByPath(gotomug));
+            _this.setCurrentMug(form.getMugByPath(goToMug));
+            return false;
         });
 
         $modal.modal('show');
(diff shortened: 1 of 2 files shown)


src/templates/find_usages.html
```diff
@@ -1,4 +1,4 @@
-<table class="table table-condensed">
+<table class="table table-hover">
```
orangejenny: Adding `table-hover` seems fine, but why get rid of `table-condensed`?

src/core.js
```diff
@@ -750,6 +750,12 @@ define([
 
         $modalBody.append($(find_usages({tableData: tableData})));
 
+        $modalBody.find('.usage-row').click(function() {
```
orangejenny: How about making the question path into a link, instead of having the entire row be a click target? Would make it more obvious that clicking will make something happen.

src/templates/find_usages.html
emord: it felt too compact. It seemed good for when you couldn't interact with the table, but it's nice to have a bit of spacing when you're clicking on something.

That said I don't have a strong feeling on this at all and will do whatever you think is best

orangejenny: I forgot the table will be a lot smaller because it's question-specific now. Yeah, leave as is, sure.

src/core.js
emord: https://github.com/dimagi/Vellum/pull/795/commits/2d4edd7eb457265db821f6117f2b0bda53f0817f

src/core.js
```diff
@@ -750,6 +750,12 @@ define([
 
         $modalBody.append($(find_usages({tableData: tableData})));
 
+        $modalBody.find('.link-to-question').click(function() {
+            var gotomug = $(this).text();
```
orangejenny: gotomug => goToMug

phillipm:
👏 👏 👏

src/core.js
```diff
@@ -750,6 +750,12 @@ define([
 
         $modalBody.append($(find_usages({tableData: tableData})));
 
+        $modalBody.find('.link-to-question').click(function() {
```
millerdev: This event handler function should return false. Otherwise I think it adds a `#` to the URL in the address bar, which is annoying.

src/core.js
emord: https://github.com/dimagi/Vellum/pull/795/commits/feeadcb8b5d329d3a1cf3af0526f2b3e512c0fe9

PR #796, opened 2016-11-04
Move analytics to its own file

emord:
Moving out analytics into its own file. Having it sprinkled throughout the code was annoying me.

@millerdev @orangejenny 

buddy @esoergel

### DIFF
--- a/src/analytics.js
+++ b/src/analytics.js
@@ -1,0 +1,100 @@
+define([], function () {
+    function _workflow(message) {
+        if (window.analytics) {
+            window.analytics.workflow(message);
+        }
+    }
+
+    function _usage(label, group, message) {
+        if (window.analytics) {
+            window.analytics.usage(label, group, message);
+        }
+    }
+
+    function _fbUsage(group, message) {
+        _usage("Form Builder", group, message);
+    }
+
+    function clickedSave() {
+        _workflow("Clicked Save in the form builder");
+    }
+
+    function clickedGoToQuestion() {
+        _fbUsage("Form Builder", "Clicked link to show in tree");
+        _workflow("Clicked on easy reference popover's link to show in tree");
+    }
+
+    function clickedToolsMenu(itemName) {
+        _fbUsage("Tools", itemName);
+    }
+
+    function clickedCopyPaste() {
+        _usage("Copy Paste", "Copy Button");
+        _workflow("Clicked Copy Button in form builder");
+    }
+
+    function fullScreenMode(formid) {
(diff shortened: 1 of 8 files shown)


emord:
Honestly not convinced that this is the best refactor (+162/-93), just an annoyance. open to doing something differently or just closing the PR

src/analytics.js
```diff
@@ -0,0 +1,100 @@
...
+
+    function _fbUsage(group, message) {
+        _usage("Form Builder", group, message);
+    }
+
+    function clickedSave() {
+        _workflow("Clicked Save in the form builder");
+    }
+
+    function clickedGoToQuestion() {
```
millerdev: I didn't look closely at how many times each of these functions is referenced, but it looks like once for most of them. I'd be happy with this module having three functions: `workflow`, `usage`, and `fbUsage`, which eliminate a bit of boilerplate, making most analytics calls a one or two liner instead of three or four.

emord: Thanks. I like that much more. Got a little too ahead of myself trying to extract them all into functions. There were only a couple that were duplicated across files

orangejenny:
Second commit is much better than the first.

Now that the code is cleaner, I'm wondering why we don't always use `fbUsage` instead of `usage`. It looks like we use `fbUsage` except for some of the easy references analytics, where the category is then form/case reference...and also copy paste...which is a bit convoluted. Would be nice to either get rid of `usage` or be able to clearly state when to use `usage` versus `fbUsage`. Not blocking this PR, just thinking aloud.

src/core.js
```diff
@@ -202,10 +202,8 @@ define([
                 var $scrollable = $node.closest(".fd-scrollable");
                 $scrollable.scrollTop($node.position().top - $scrollable.position().top);
             }
-            if (window.analytics) {
-                window.analytics.usage("Form Builder", "Clicked link to show in tree");
-                window.analytics.workflow("Clicked on easy reference popover's link to show in tree");
-            }
+            analytics.fbUsage("Form Builder", "Clicked link to show in tree");
```
millerdev: Should be either
```js
analytics.fbUsage("Clicked link to show in tree");
```
or
```js
analytics.usage("Form Builder", "Clicked link to show in tree");
```
right?

millerdev:
> I'm wondering why we don't always use `fbUsage` instead of `usage`.

Yeah, I agree it would be nice to merge `fbUsage` and `usage` into a single function (named `usage`, does what `fbUsage` does). However, I'm guessing that might mess up @dimagi/product's stats since it would put some stats in new buckets?

PR #798, opened 2016-11-08
find usages should be full screen

emord:
@czue bug you mentioned

image

That's pretty ugly, so making find usages full screen for now

buddy @esoergel

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -565,7 +565,7 @@ define([
                          "to your form. Press 'Update Source' to save changes, or 'Close' to cancel."
         }));
 
-        $modal.addClass('fd-source-modal fd-full-screen-modal')
+        $modal.addClass('fd-full-screen-modal')
             .find('.modal-body')
             .removeClass('form-horizontal').removeClass('form')
             .html($updateForm);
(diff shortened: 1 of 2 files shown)


src/less-style/modals.less
```diff
@@ -37,7 +37,7 @@
 }
 
 // Full-screen modal for displaying form XML
-.fd-source-modal {
+.fd-full-screen-modal {
```
orangejenny: Doesn't this rename mean you can remove `fd-source-modal` from `core.js`? I don't see any other usages of `fd-source-modal`.

src/core.js
```diff
@@ -725,6 +725,7 @@ define([
             form = _this.data.core.form,
             tableData = form.findUsages();
 
+        $modal.addClass('fd-find-usages-modal fd-full-screen-modal');
```
orangejenny: What's the purpose of the `fd-find-usages-modal` class?

czue:
thanks J$

src/less-style/modals.less
emord: I removed those. I initially had them separately and hesitated to remove something descriptive

orangejenny: sorry, forgot I was looking at individual commits

PR #799, opened 2016-11-10
Fix drag/drop case property with rich text off

millerdev:
Noticed while trying to reproduce http://manage.dimagi.com/default.asp?241253

@emord cc @nickpell

### DIFF
--- a/src/databrowser.js
+++ b/src/databrowser.js
@@ -121,8 +121,7 @@ define([
 
     function dataTreeJson(data, vellum) {
         function treeNode(node) {
-            var path = isRichText && node.hashtag || node.xpath,
-                getNodes = function () { return _.map(node.getNodes(), treeNode); },
+            var getNodes = function () { return _.map(node.getNodes(), treeNode); },
                 children = node.recursive ? true : getNodes();
             return {
                 text: node.name,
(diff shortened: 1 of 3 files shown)


src/databrowser.js
```diff
@@ -132,13 +131,14 @@ define([
...
+                    handleDrop: _.partial(handleDrop, node, node.sourceInfo),
                     getNodes: node.recursive ? getNodes : null,
                 }
             };
         }
-        function handleDrop(path, info, target) {
-            var widget = widgets.util.getWidget(target, vellum);
+        function handleDrop(node, info, target) {
+            var widget = widgets.util.getWidget(target, vellum),
+                path = widget.mug.form.richText && node.hashtag || node.xpath;
```
emord: It's possible I wrote this code and totally forgot but what's the advantage of `... && ... || ...` over ` ... ? ... : ...` ?

millerdev: I'm guessing I wrote that without even thinking (`x and y or z` used to be the idiomatic way to do an inline conditional in Python, not without its gotchas (`true && "" || "wrong" === "wrong"`), so not recommended). Fixed in 1daecca

millerdev: I guess in this case if `node.hashtag` happened to be empty it would fall back to `node.xpath`, which is probably what you want, although that syntax is still probably too tricky to use for that intent.

millerdev:
Realized I may have pushed the limit there since I merged after adding more code since this was approved, but figured it was a small enough change and I want to run `vellum-to-hq` to get this out for Jen tomorrow.

PR #800, opened 2016-11-10
Bump jQuery 3.1.0 to 3.1.1

orangejenny:
HQ is already on 3.1.1

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -22,7 +22,7 @@
     ],
     "dependencies": {
         "underscore": "~1.8.3",
-        "jquery": "~3.1.0",
+        "jquery": "~3.1.1",
         "requirejs": "~2.3.1",
         "requirejs-text": "~2.0.12",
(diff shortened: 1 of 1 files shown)


bower.json
```diff
@@ -22,7 +22,7 @@
     ],
     "dependencies": {
         "underscore": "~1.8.3",
-        "jquery": "~3.1.0",
+        "jquery": "~3.1.1",
```
millerdev: fwiw, `~3.1.0` matches `3.1.1`, and any other patch-level version in the `3.1` series ([SO answer that covers the meaning of `~`](http://stackoverflow.com/a/20412378/10840)). So this PR isn't strictly necessary unless 3.1.1 is required, which I assume it is not.

orangejenny: I did not know that (or, really, anything else about semantic versioning). Just read through the link, thank you for pointing it out :)

PR #802, opened 2016-11-11
Contextual title for expression editor

orangejenny:
@emord 

I'd expected to do this by tweaking a template, and then learned that the expression editor has a generic outer template, and making it question-specific looked like a bear.

screen shot 2016-11-11 at 3 55 13 pm


proteusvacuum:
😻

millerdev:
Love this change!

PR #805, opened 2016-11-14
Load case hashtags on enable easy references

millerdev:
http://manage.dimagi.com/default.asp?241253

@emord or @orangejenny


src/form.js
```diff
@@ -568,6 +569,10 @@ define([
         },
         setAttr: function (slug, val) {
             this[slug] = val;
+            if (slug === "richText" && val && !this.hasCaseHashtags) {
```
emord: Feels weird having `"richText"` hanging out here. May be worth pulling out into a constant since it's in writer, parser and form? (maybe we should have a legit constant file too?)

PR #806, opened 2016-11-14
search find usages

emord:
image
 
buddy @esoergel

### DIFF
--- a/src/atwho.js
+++ b/src/atwho.js
@@ -109,7 +109,7 @@ define([
         });
 
         if (options.useRichText) {
-            options.insertTpl = '${name}';
+            options.insertTpl = '${hashtagPath}';
             options.functionOverrides.insert = function(content, $li) {
                 // this references internal At.js object
                 this.query.el.remove();
--- a/src/core.js
+++ b/src/core.js
@@ -735,6 +735,9 @@ define([
         $modalBody.append($('<div>').append($('<input id="find-usages-search">').attr('type', 'text')));
         $modalBody.append($(find_usages({tableData: tableData})));
 
+        this._resizeFullScreenModal($modal);
+        $modal.modal('show');
+
         $modalBody.find('.link-to-question').click(function() {
             var goToMug = $(this).text();
             $modal.modal('hide');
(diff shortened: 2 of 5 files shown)


src/atwho.js
```diff
@@ -109,7 +109,7 @@ define([
         });
 
         if (options.useRichText) {
-            options.insertTpl = '${name}';
+            options.insertTpl = '${hashtagPath}';
```
emord: This was interesting. `fuse.js` returns `name` as the absolute path (`/data/....`) so when using rich text it was insterting /data/ and the rich text code was converting that to hashtags.

src/templates/find_usages.html
```diff
@@ -1,4 +1,4 @@
-<table class="table table-hover">
+<table class="table table-hover find-usages-table">
```
orangejenny: this can be removed, i think

src/core.js
```diff
@@ -729,17 +731,45 @@ define([
             tableData = form.findUsages();
 
         $modal.addClass('fd-full-screen-modal');
+        $modalBody.append($('<div>').append($('<label for="find-usages-search">').text("Search")));
+        $modalBody.append($('<div>').append($('<input id="find-usages-search">').attr('type', 'text')));
```
millerdev: Can the search box be styled more like the data browser search box, with a magnifying glass icon to the left of the search box rather than a *Search* label above it?

src/atwho.js
millerdev: I didn't understand that, so I'm not sure what to take away. Is there some cleanup that could be done make this less messy?

emord: I think there is some clean up that could be done. https://github.com/dimagi/Vellum/blob/master/src/fuse.js#L67-L78 by using `${name}` here it was inserting `/data/question` instead of  `#form/question` and it was getting converted to hashtag later, which seems silly.

I don't quite remember my thinking here, but it probably had to do with trying to support easy references as a feature flag at first. I think this could be changed to assume easy references instead of not. I'm not sure how deep that will go, so not going to look into it right now

src/core.js
emord: https://github.com/dimagi/Vellum/pull/806/commits/0909f13a37c91d6de39541e54542046e7345d2ad

PR #807, opened 2016-11-14
Add context to question properties window

orangejenny:
screen shot 2016-11-14 at 3 27 24 pm

@millerdev

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1294,7 +1294,10 @@ define([
             _this.refreshMugName(e.mug);
             _this.toggleConstraintItext(e.mug);
         }).on('change-display-language', function (e) {
-            _this.refreshMugName(_this.getCurrentlySelectedMug());
+            var mug = _this.getCurrentlySelectedMug();
+            if (mug) {
+                _this.refreshMugName(mug);
+            }
         }).on('mug-property-change', function (e) {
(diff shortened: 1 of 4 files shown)


src/core.js
```diff
@@ -1528,6 +1528,8 @@ define([
 
         this.showContentRight();
         $props.addClass("hide");
+        var questionName = mug.getDisplayName(this.data.core.currentItextDisplayLanguage);
```
orangejenny: I lean towards id only because it's simpler, but I don't feel strongly about it.

Updated to change on label change...I'd been lazy about that, but yeah, it's a better experience.

orangejenny:
Realized this looks weird with markdown - the plain markdown syntax is displayed - but it's the same behavior as the question tree.

src/javaRosa/plugin.js
```diff
@@ -192,8 +192,8 @@ define([
                 }
             });
 
-            if (form && form.vellum.getCurrentlySelectedMug()) {
-                form.vellum.getCurrentlySelectedMug().fire({
+            if (form) {
```
emord: Nice I like this change. I wonder when this would happen without a form object though

millerdev:
Love it! Even does the right thing for questions with huge amounts of label text, newlines, et al.

PR #808, opened 2016-11-15
Auto-generate question id based on label

orangejenny:
k8kli4j6wm

The id is generated the first time you click away from the question, unless you've already manually entered an id. The question id field does **not** update live as you type - I experimented with this and ran into problems with the id validation running repeatedly. Also experimented with generating the id on blur instead of on question change, but the blur/focus code is a little fragile already (the initial focus is delayed because it needs to wait for ckeditor to initialize, which makes it difficult to know when the first blur is). I moved the id into `Logic` when updating it as you type didn't work out - it just sits there blank, and I'd rather discourage people from entering an id. Moderate/advanced users can still tab into it as long as they have the logic section open.

@emord / @millerdev / @amsagoff how does this sound? I'm certainly open to digging into this more if the experience described about sounds problematic. Also happy to put on staging if you want to play with it.

Changes to make:
- [x] ~~On HQ, new questions display as "undefined" in the question tree~~ Does't seem to be happening anymore.
- [x] ~~If the label is blank, display just the icon in the question tree, rather than defaulting to the question ID~~ Not sure this is a good idea or worth the effort, will talk with Amelia about it.
- [x] Visual changes to question id input: try dropping the current label, and doing an input with "Question ID" prepended and a light opacity.
- [x] PR feedback below


src/core.js
```diff
@@ -1149,9 +1149,26 @@ define([
...
 
+    // Suggest a node ID, based on the mug's label
+    fn.nodeIDFromLabel = function(mug) {
+        var suggestedID = this.getMugDisplayName(mug) || "";
+        suggestedID = $("<div/>").html(suggestedID).text();     // strip any HTML (i.e., bubbles)
+        suggestedID = suggestedID.toLowerCase();
+        suggestedID = suggestedID.trim();
+        suggestedID = suggestedID.replace(/\s+/g, '_');         // collapse whitespace & replace with underscores
+        suggestedID = suggestedID.replace(/[^\w\-]/g, '');      // strip illegal characters
+        suggestedID = suggestedID.substring(0, 75);             // no exceedingly long IDs
```
emord: may be worthwhile to filter out "stop words" in the future so that these are more useful. probably not necessary in this PR though

src/core.js
```diff
@@ -2002,10 +2027,8 @@ define([
                 properties: this.getMainProperties(),
                 help: {
                     title: "Basic",
-                    text: "<p>The <strong>Question ID</strong> is an internal identifier for a question. " +
-                        "It does not appear on the phone. It is the name of the question in data exports.</p>" +
-                        "<p>The <strong>Label</strong> is text that appears in the application. " +
-                        "This text will not appear in data exports.</p> ",
+                    text: "<p>The <strong>Label</strong> is text that appears in the application. " +
```
emord: For a later PR these help texts seem strange. I think they should be on the question property and not on the accordion

src/core.js
```diff
@@ -1378,9 +1395,17 @@ define([
             }
             analytics.workflow("Added question in form builder");
             mug = _this.data.core.form.createQuestion(foo.mug, foo.position, qType);
-            var $firstInput = _this.$f.find(".fd-question-properties input:text:visible:first");
-            if ($firstInput.length) {
-                $firstInput.focus().select();
+
+            // Focus on first input, which might be a normal input or a rich text input
+            var $firstGroup = _this.$f.find(".fd-question-properties .form-group:first");
```
emord: This doesn't have the `visible` selector. What happens if the basic accordion is collapsed?

orangejenny: Nothing gets focused. I think that's okay, not worried about people who have basic collapsed and other sections opened.

src/core.js
orangejenny: yeah, been meaning to do that, will do another PR

src/core.js
orangejenny: i like that idea. agreed on not doing it right now.

amsagoff:
I'll take you up on your offer to put this on staging :) Seems like a good
idea to test out with some FMs and make sure it's not too disruptive, which
I can take lead on.

Thanks!

orangejenny:
@amsagoff deployed to staging

orangejenny:
Noting that in HQ, adding a new question makes the question tree display "undefined":
screen shot 2016-11-15 at 3 16 45 pm

src/richText.js
```diff
@@ -232,6 +232,15 @@ define([
                 editor.fire("saveSnapshot");
                 return wrapper;
             },
+            focus: function() {
+                if (editor.status === "ready") {
+                    editor.focus();
+                } else {
+                    editor.on('instanceReady', function () {
```
millerdev: Wonder if it would make sense to set a flag indicating that the editor will be focused when it's ready and do not create another event handler if it is set? That would prevent creating a pile of `instanceReady` even handlers if `focus()` is called many times before the instance is ready.

tests/javaRosa.js
```diff
@@ -366,8 +366,8 @@ define([
 
         it("itext widget should not overwrite label with question id", function () {
             util.loadXML("");
-            var q1 = util.addQuestion("Text");
-            util.addQuestion("Text");
+            var q1 = util.addQuestion("Text", "question1");
+            util.addQuestion("Text", "question2");
```
millerdev: Given the video in the PR description, I would expect these IDs to default to the values specified here. Is it now required to always pass a nodeID to `addQuestion()`? If so, what do you think of modifying `util.addQuestion` to get the next default question ID with `form.generate_question_id()` if not specified?

src/richText.js
orangejenny: Good point. Updated to turn the listener off before adding it in https://github.com/dimagi/Vellum/pull/808/commits/eded8b679928103b330b18592074bc0ba19dea1b (just because it's less code than adding a flag).

tests/javaRosa.js
orangejenny: > Is it now required to always pass a nodeID to `addQuestion()`?

Not exactly, but it's harder to predict when the id will be generated. It does make sense that 99% of the time tests will want the id to be generated by the time `addQuestion` returns. Updated the `addQuestion` in `tests/utils` to always generate an id.

orangejenny:
Closing in favor of https://github.com/dimagi/Vellum/pull/840

PR #809, opened 2016-11-15
Associate question ID help with question ID instead of section

orangejenny:
@emord 

Left the label help at the section level, rather than have it repeated for every language.


emord:
Looks good to me. I imagine this will cause conflicts with your other PR

orangejenny:
yep

PR #810, opened 2016-11-15
Fix print question type template collision

millerdev:
Fixes:
- http://manage.dimagi.com/default.asp?229441
- http://manage.dimagi.com/default.asp?220337

This implements most of the items in the [proposed solution](http://manage.dimagi.com/default.asp?229441#BugEvent.1172875). Things not implemented:

- Itext MM paths are still assigned prior to upload, although the path changes each time a new file is uploaded.
- The button for uploading new MM content still changes its title as it did before.
- The new *Replace ... for all linked questions* button is not implemented.

Once this is merged each multimedia upload done in the form builder will create a unique new `jr://commcare/...` path, which means that the media map for a given app could grow to be larger than it would have previously if many media items are uploaded and then re-uploaded later. In other words, I think older abandoned media items will still be present in the media map because HQ does not clear them out if they are no longer referenced (except when uploading the app to the Exchange, I think). @dimagi/team-commcare-hq will this cause problems in HQ (I think not)? @ctsims will it cause problems on mobile?

Review commits individually 🐠 

@orangejenny or @emord

### DIFF
--- a/src/intentManager.js
+++ b/src/intentManager.js
@@ -61,8 +61,8 @@ define([
         return widget;
     }
 
-    function mediaWidget(mug, options) {
-        var widget = widgets.media(mug, options);
+    function printTemplateWidget(mug, options) {
+        var widget = widgets.abstractMediaWidget(mug, options);
         widget.getBaseMediaPath = function () {
             return "jr://file/commcare/text/" + mug.p.nodeID;
         };
@@ -312,7 +312,7 @@ define([
             docTemplate: {
                 lstring: 'Document Template',
                 visibility: 'visible',
-                widget: mediaWidget,
+                widget: printTemplateWidget,
             },
             androidIntentAppId: { visibility: 'hidden' },
         }
--- a/src/intentManager.js
+++ b/src/intentManager.js
@@ -151,7 +151,7 @@ define([
     }
 
     function syncMugWithIntent (tags, mug) {
-        // called when initializing a mug from a parsed form
+        // called when initializing a new mug or when parsing a form
         if (mug.__className === "AndroidIntent" ||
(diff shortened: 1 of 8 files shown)


src/intentManager.js
```diff
@@ -61,6 +61,14 @@ define([
         return widget;
     }
 
+    function mediaWidget(mug, options) {
```
emord: Should this be called `printTemplateWidget` since it already inherits from a `media` widget?

millerdev: good idea 0a5a52b

ctsims:
Daniel,

Does that mean that multimedia that is no longer used will get sent to the
phone still? That sounds like a pretty big hit. A lot of apps change
multimedia _frequently_ during iteration, so if the multimedia never clears
unused elements I would definitely consider it an issue.

millerdev:
@ctsims 
> Does that mean that multimedia that is no longer used will get sent to the phone still?

I don't know, that's why I am asking. Does the phone download every multimedia item that has ever been created, or only multimedia that is referenced by the app? If the former, then it seems like phones could be downloading more media than they need, even now without this change. For example, what if a question is added to a form, multimedia is uploaded for the question, then the question is deleted. Is the multimedia for the question still downloaded by the phone even though it's no longer referenced?

millerdev:
Update: clarified wording on previous comment.

ctsims:
Gotcha,

The thing that controls whether multimedia makes it down to the phone is
whether it is included in the media_suite.xml file that is in the app
package, so as long as that is getting cleaned up properly things will work.

millerdev:
@ctsims I verified that no-longer-referenced images are not included in media_suite.xml

this is good to merge

PR #811, opened 2016-11-16
Do not set form.hasCaseHashtags prematurely

millerdev:
http://manage.dimagi.com/default.asp?241253

@orangejenny or @emord

### DIFF
--- a/src/form.js
+++ b/src/form.js
@@ -179,7 +179,7 @@ define([
                 form.hashtagMap = _.clone(vellum.datasources.getHashtagMap({}));
                 form.invertedHashtagMap = _.invert(form.hashtagMap);
                 form.hashtagTransformations = vellum.datasources.getHashtagTransforms({});
-                form.hasCaseHashtags = true;
+                form.hasCaseHashtags = vellum.datasources.isReady();
             } else {
                 form.hashtagMap = {};
                 form.invertedHashtagMap = {};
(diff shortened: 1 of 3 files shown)


tests/datasources.js
```diff
@@ -203,13 +203,6 @@ define([
                 assert.equal(data.find("option:selected").text(), "outer - inner");
             });
 
-            /*
-            TODO make this test pass
-            Should not lose xpath expression when rich text is off.
```
millerdev: This is unrelated to the PR, but was fixed a while ago.

src/form.js
```diff
@@ -179,7 +179,7 @@ define([
                 form.hashtagMap = _.clone(vellum.datasources.getHashtagMap({}));
                 form.invertedHashtagMap = _.invert(form.hashtagMap);
                 form.hashtagTransformations = vellum.datasources.getHashtagTransforms({});
-                form.hasCaseHashtags = true;
+                form.hasCaseHashtags = vellum.datasources.isReady();
```
orangejenny: Is it worthwhile/possible to flip this to true once the datasources are ready?

millerdev: Yes, that's what happens now (see [line 140](https://github.com/dimagi/Vellum/blob/542a4e27dabf10b715645a64601a965df240208f/src/form.js#L140)) when rich text is enabled. If rich text is disabled then it won't be flipped until rich text is enabled.

PR #813, opened 2016-11-18
Keep question properties head updated when current language is _ids

orangejenny:
http://manage.dimagi.com/default.asp?241935

cc @czue


orangejenny:
Daniel's looking into a different approach to this; don't merge.

PR #814, opened 2016-11-18
Refresh mug name on display mug properties

millerdev:
http://manage.dimagi.com/default.asp?241935

Plus some tech debt.

@orangejenny

### DIFF
--- a/tests/javaRosa.js
+++ b/tests/javaRosa.js
@@ -395,6 +395,22 @@ define([
             assert.equal($(treeSelector).text(), "hindi");
         });
 
+        it("should not fire lost of events for every mug on change label", function () {
+            var form = util.loadXML(""),
+                names = {q1: 0, q2: 0};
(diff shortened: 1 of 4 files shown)


tests/javaRosa.js
```diff
@@ -395,6 +395,22 @@ define([
...
+                names[event.mug.p.nodeID]++;
+            });
+            util.paste([
+                ["id", "type"],
+                ["/q1", "Text"],
+                ["/q2", "Text"],
+            ]);
+            util.clickQuestion("q2");
+            $("[name='itext-en-label']").val('english').change();
+            assert.deepEqual(names, {q1: 0, q2: 1});
```
millerdev: Prior to this change the number of events fired was 9 per question. I wonder if this will affect the performance of large forms (in a good way).

orangejenny:
looks good to me!

PR #815, opened 2016-11-21
Don't reload on enter find usages

emord:
http://manage.dimagi.com/default.asp?242019 @proteusvacuum

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -746,7 +746,7 @@ define([
             return false;
         });
 
-        $modalBody.find('#findUsagesSearch').on('keyup keypress inserted.atwho', function (e) {
+        $modalBody.find('#findUsagesSearch').on('keypress inserted.atwho', function (e) {
             var searchKey = $.trim(this.value),
                 filteredData = {};
             if (!searchKey) {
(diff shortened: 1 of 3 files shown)


src/core.js
```diff
@@ -746,7 +746,7 @@ define([
             return false;
         });
 
-        $modalBody.find('#findUsagesSearch').on('keyup inserted.atwho', function () {
+        $modalBody.find('#findUsagesSearch').on('keyup keypress inserted.atwho', function (e) {
```
orangejenny: Might as well remove the `keyup` rather than firing two events for each character.

orangejenny:
Looks fine, made a couple picky comments.

orangejenny:
Thought about this for another minute...it'd be cleaner / more semantic to make the `form` in `find_usages_search.html` a div, since it isn't really a form, and that would fix this problem without adding js. Then would just need to tweak the `.fd-search` styles in `forms.less` to target `.form-inline` instead of `form` elements.

emord:
Yeah. it looks nicer in code now. I originally did this as that's how the search on the case properties tree works. I attempted to change that too, but changing it changed the css somehow, so leaving that for another day

src/core.js
emord: Gah. I was on the wrong branch and accidentally pushed `debounce` to master https://github.com/dimagi/Vellum/commit/65cccf5d3ccc676cedc05b8d89247b3e3293fb68 I'm not going to worry about reverting it right now as it's a small change, but that should also be reviewed

orangejenny:
changes look good

PR #816, opened 2016-11-21
add timeout to debounce

emord:
https://github.com/dimagi/Vellum/commit/65cccf5d3ccc676cedc05b8d89247b3e3293fb68

luckily default is 0

@orangejenny


orangejenny:
to quibble, it defaults to undefined, but the difference between zero & undefined doesn't end up mattering.

PR #817, opened 2016-11-21
New CKEditor with improved nav keys on widgets

millerdev:
https://github.com/ckeditor/ckeditor-dev/compare/4.5.7...millerdev:widget-nav-4.5.7

https://github.com/ckeditor/ckeditor-dev/pull/304

This build was created with the following procedure:

- clone https://github.com/millerdev/ckeditor-dev and `cd` into `ckeditor-dev`
- checkout `widget-nav-4.5.7` branch
- download the minimalist skin (v1.0) and unzip/place it in the `./skins` directory.
- download confighelper plugin (v1.8.3) and unzip/place it in the `./plugins` directory
- copy `vellum/lib/ckeditor/build-config.js` into `./dev/builder`
- build: `./dev/builder/build.sh --no-zip --no-tar`
- copy `./dev/builder/release/ckeditor/ckeditor.js` into `vellum/lib/ckeditor/`


emord:
From your other PR:

> Move to right if key is HOME/LEFT/UP 

move to right if it's left/up? shouldn't it be the other way around?

Also why did you decide to do this based off of 4.5.7 instead of 4.5.11? is that what we're on right now?

can you update https://github.com/dimagi/Vellum/blob/master/lib/ckeditor/dimagi-README.md

millerdev:
> move to right if it's left/up? shouldn't it be the other way around?

nope. It moves the real selection to the right so the subsequent (real) key handler includes the widget in the selection (if any) (see [comment](https://github.com/ckeditor/ckeditor-dev/pull/304/files#diff-ae237139107d046c67dc8622d25bace9R385)). All of the new key combinations either include SHIFT (for changing the selection) or will move the cursor to the beginning or end of the line in which case the initial direction doesn't really matter.

> Also why did you decide to do this based off of 4.5.7 instead of 4.5.11? is that what we're on right now?

Oh, I thought we were on 4.5.7, at least that's the most recent version in [lib/ckeditor/CHANGES.md](https://github.com/dimagi/Vellum/blob/fddd4b47fbdc58ea9cc39dda95228066cf3d2741/lib/ckeditor/CHANGES.md). I'd be happy to update to a more recent version if that's what we've been using or if we think that's a better idea.

emord:
Ah. thanks that explanation makes sense.

Yeah. we're probably on 4.5.7 right now. I wasn't sure if we wanted to go ahead and upgrade to the latest minor version. there's also 4.6.0, but I haven't looked into that

millerdev:
Given that the other files in the build output did not look exactly like what was already in vellum/lib/ckeditor, I'd prefer to stay at 4.5.7 for now since I only copied over the ckeditor.js file. I think it's a good idea to look at 4.6 and upgrade to that, especially if my PR gets merged soon.

PR #818, opened 2016-11-23
better naming

emord:
@proteusvacuum http://manage.dimagi.com/default.asp?242247

### DIFF
--- a/src/templates/widget_index_case.html
+++ b/src/templates/widget_index_case.html
@@ -5,7 +5,7 @@
     <div class="form-group">
       <div class="col-sm-11">
         <input class="fd-index-property-name form-control"
-               value="<%-k%>" type="text" placeholder="Case Name"/>
+               value="<%-k%>" type="text" placeholder="Relationship Identifier (e.g. parent or host)"/>
       </div>
       <div class="col-sm-1">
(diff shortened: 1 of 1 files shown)


src/templates/widget_index_case.html
```diff
@@ -16,19 +16,19 @@
     <div class="form-group">
       <div class="col-sm-11">
         <input class="fd-index-property-source form-control"
-               value="<%-v['calculate']%>" type="text" placeholder="Calculation" />
+               value="<%-v['calculate']%>" type="text" placeholder="Referenced Case ID" />
```
proteusvacuum: It's confusing that this gets saved to `'calculate'`. Just verifying that when this makes an index block it makes something like this:
```xml
<index>
     <{relationship identifier} case_type={case_type} relationship={relationship}>{Referenced case ID}</ {relationship identifier}>
</index>
```

emord: Yeah. it uses a bind instead of saving in place. It's technically the same as a "calculate condition" from a hidden value, but I don't see a time when you wouldn't want it to be a "referenced case id"

proteusvacuum: Oh I see, that makes sense. Thanks for the explanation.

PR #821, opened 2016-11-29
Fix CKEditor skipping many spaces

millerdev:
Took a long time to track this down. `pre-wrap` caused a string of normal spaces (char 32), rather than alternating space/`$nbsp;`, to be entered (at least in Chrome), which was a problem because normal spaces are skipped by CKEditor when navigating toward a widget with arrow keys or backspace/del.

This fix was tested in Firefox and Chrome.

@emord

### DIFF
--- a/src/less-style/inputs.less
+++ b/src/less-style/inputs.less
@@ -44,7 +44,7 @@
     height: 4.5em;
     overflow: auto;
     resize: both;
-    white-space: pre-wrap;
+    white-space: normal;
     word-break: break-word;
 
     &.fd-xpath-editor-text {


millerdev:
> not sure why it's not needed anymore

Not sure I understand. Are you assuming `white-space: normal` is the same as not specifying a `white-space` rule at all? I'm pretty sure that's not the case. When I removed the rule entirely it did not wrap long lines, but when changing it to `normal` it wraps. Both removing it entirely and `normal` fix the issue in CKEditor.

emord:
Ah. thanks for clarification. That's what I was assuming.

PR #826, opened 2016-12-07
User properties

millerdev:
Add support for new data sources schema elements that allow user properties to be specified and referenced with hashtag expressions (`#user/property`). It also removes magic/ambiguity from the data sources schema. Finally, all hard-coded references to `#case` hashtags have been removed since the data sources schema may define arbitrary hashtag names.

All changes here should be entirely backward-compatible with the existing data sources schema provided by HQ. Until HQ is updated to provide a new schema containing user properties, the only end-user visible change is c51fd69.

Spec: https://docs.google.com/document/d/1kPVwZketyWl6aRo1ZNCM3EngkrTsgmFbnBXAv-FThKU

Review commits individually 🐠 

@emord @orangejenny

### DIFF
--- a/src/databrowser.js
+++ b/src/databrowser.js
@@ -138,11 +138,14 @@ define([
         }
         function handleDrop(node, info, target) {
             var widget = widgets.util.getWidget(target, vellum),
-                path = widget.mug.form.richText ? node.hashtag : node.xpath;
+                path = widget.mug.form.richText ? node.hashtag : node.xpath,
+                id;
             while (info) {
-                var id = widget.addInstanceRef({id: info.id, src: info.uri});
-                if (id !== info.id) {
-                    path = widget.mug.form.updateInstanceQuery(path, id, info.id);
+                if (info.id && info.uri) {
+                    id = widget.addInstanceRef({id: info.id, src: info.uri});
+                    if (id !== info.id) {
+                        path = widget.mug.form.updateInstanceQuery(path, id, info.id);
+                    }
                 }
                 info = info._parent;
             }
--- a/src/datasources.js
+++ b/src/datasources.js
@@ -26,6 +26,11 @@
  *                      hashtag: string (optional hashtag prefix)
  *                      source: string (optional data source id, defaults to this data source)
  *                      subset: string (optional subset id)
+ *                      subset_key: string (optional subset key, if omitted then
+ *                                  the first subset with an `id` matching
+ *                                  this reference's `subset` will be used)
+ *                      subset_filter: boolean (optional, if true include
+ *                                     subset filter in xpath expression)
  *                      key: string (referenced property)
  *                  }
  *              },
@@ -281,14 +286,18 @@ define([
                             info.hashtag = '#case/' + ref.subset;
                         }
                     }
-                    path = "instance('" + source.id + "')" + source.path +
-                           "[" + ref.key + " = " + path + "]";
+                    var keyPath = path;
+                    path = "instance('" + source.id + "')" + source.path;
+                    if (ref.subset_filter && ref.subset_key && ref.subset) {
+                        path += "[" + ref.subset_key + " = '" + ref.subset + "']";
+                    }
+                    path += "[" + ref.key + " = " + keyPath + "]";
                     if (source.subsets && ref.subset) {
-                        // magic: match key: "@case_type"
-                        source = _.findWhere(
-                            source.subsets,
-                            {id: ref.subset, key: "@case_type"}
-                        ) || source;
+                        var where = {id: ref.subset};
+                        if (ref.subset_key) {
+                            where.key = ref.subset_key;
+                        }
+                        source = _.findWhere(source.subsets, where) || source;
                     }
                     var name = source.name || source.id;
                     if (name) {
--- a/src/datasources.js
+++ b/src/datasources.js
@@ -49,7 +49,11 @@
  *              name: string (human readable name)
  *              structure: { ... }
  *              related: {
- *                  string (relationship): string (related subset id),
+ *                  string (relationship): {
+ *                      // same keys as structure.ref-element.reference
+ *                      // plus one more:
+ *                      index: string (optional index path; default is '/index')
+ *                  } or string (related subset id, legacy)
  *                  ...
  *              }
  *          }]
@@ -278,8 +282,8 @@ define([
                         hashtag: ref.hashtag,
                     });
                     if (!ref.hashtag && source.id === "casedb") {
-                        // magic: case hashtags
-                        // TODO put these in reference.hashtag (on server)
+                        // legacy magic: case hashtags
+                        // TODO remove when HQ sends new schema format
                         if (ref.subset === "case") {
                             info.hashtag = '#case';
                         } else {
@@ -334,11 +338,18 @@ define([
                 .value();
             if (source && source.related) {
                 nodes = _.chain(source.related)
-                    .map(function (subset, relation) {
-                        // magic: reference key: @case_id
-                        var item = {reference: {subset: subset, key: "@case_id"}};
-                        // magic: append "/index" to path
-                        return node(source, path + "/index", info, true)(item, relation);
+                    .map(function (ref, relation) {
+                        var item, index;
+                        if (_.isObject(ref)) {
+                            item = {reference: ref};
+                            index = ref.index || "/index";
+                        } else {
+                            // legacy/magic
+                            // TODO remove when HQ sends new schema format
+                            item = {reference: {subset: ref, key: "@case_id"}};
+                            index = "/index";
+                        }
+                        return node(source, path + index, info, true)(item, relation);
                     })
                     .sortBy("text")
                     .value()
--- a/src/datasources.js
+++ b/src/datasources.js
@@ -19,7 +19,7 @@
  *                  structure: {
  *                      inner-element: { }
  *                  }
- *                  name: "Element" (the text used in dropdown for this element)
+ *                  name: "Element" (the text used to represent this element)
  *              },
  *              ref-element: {
  *                  reference: {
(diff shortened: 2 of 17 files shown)


src/util.js
```diff
@@ -363,8 +363,13 @@ define([
...
 
-    that.isCaseReference = function (value) {
-        return value.startsWith('#case/');
+    that.getReferenceName = function (value) {
+        var ref = /^#([^\/]+)\//.exec(value);
+        if (!ref) {
+            return "Form Reference";
+        }
+        ref = ref[1][0].toUpperCase() + ref[1].substring(1).toLowerCase();
+        return ref + " Reference";
```
millerdev: This is used for analytics. It returns "Case Reference" and "Form Reference" (the same categories that have been used for hashtag analytics up to this point), and will also send a new category for each new hashtag namespace, for exmaple "User Reference" for `#user/...` hashtags. FYI @dimagi/product

millerdev:
I'm leaning toward not doing QA on this since there are minimal user-facing changes, and I'm pretty confident that all changes to data sources schema are backward compatible. However, I'm probably biased in favor of my own work. @emord or @orangejenny do you have thoughts on that?

tests/form.js
```diff
@@ -527,7 +527,61 @@ define([
...
+                "/data": null,
+                "/data/": null,
+                "/data/prop": "xpath",
+
+                "#case": null,
+                "#case/": "is",
+                "#case/prop": "has xpath",
+                /* TODO should these pass? (they don't)
+                "#case/ ": null,
+                "#case/+": null,
```
emord: Agree `#case/+` should be `null`. I think `#case/: "is"` makes sense, why do you think it should be `null`?

tests/form.js
```diff
@@ -527,7 +527,61 @@ define([
...
+            });
+
+            _.each({
+                "": null,
+                "#": null,
+
+                "#form": null,
+                /* TODO should these pass? (they don't)
+                "#form/": "is",
+                "#form/prop": "has",
```
emord: From my memory. I think  this may be due to only checking valid hashtags against #case due to being interweaved with the data sources code, agreed with your  comment that they should pass

src/richText.js
```diff
@@ -424,12 +422,12 @@ define([
      */
     function makeBubble(form, xpath) {
         function _parseXPath(xpath, form) {
-            if (CASE_REF_REGEX.test(xpath)) {
+            if (!FORM_REF_REGEX.test(xpath)) {
```
emord: I think this means that #user and #case will be noted as the with the same classes and icons. Guessing that that was intentional, but curious what were the arguments against using a different UI for them?

src/form.js
```diff
@@ -174,12 +174,16 @@ define([
...
                 oldHashtags = form.hashtagMap;
+            form.hashtagNamespaces = {form: true};
             if (form.richText) {
                 // TODO always load hashtags, even when rich text is disabled
                 form.hashtagMap = _.clone(vellum.datasources.getHashtagMap({}));
                 form.invertedHashtagMap = _.invert(form.hashtagMap);
                 form.hashtagTransformations = vellum.datasources.getHashtagTransforms({});
                 form.hasCaseHashtags = vellum.datasources.isReady();
+                _.each(form.hashtagTransformations, function (x, prefix) {
+                    form.hashtagNamespaces[/^#([^\/]+)/.exec(prefix)[1]] = true;
```
emord: This regex could be extracted out to a constant with a name. Also I think something like `#user+case` would match here, but we probably only want to match letters right?

src/datasources.js
```diff
@@ -281,14 +286,18 @@ define([
                             info.hashtag = '#case/' + ref.subset;
                         }
                     }
-                    path = "instance('" + source.id + "')" + source.path +
-                           "[" + ref.key + " = " + path + "]";
+                    var keyPath = path;
+                    path = "instance('" + source.id + "')" + source.path;
+                    if (ref.subset_filter && ref.subset_key && ref.subset) {
```
emord: Do you think it would be useful to have subsets be a list, so that there could be multiple filters here?

tests/databrowser.js
```diff
@@ -481,5 +481,119 @@ define([
...
+                });
+            });
+
+            it("should drag/drop user property", function (done) {
+                var form = util.loadXML(""),
+                    mug = util.addQuestion("DataBindOnly", "mug"),
+                    input = $("[name=property-calculateAttr]"),
+                    editor = input.ckeditor().editor,
+                    widget = util.getWidget('property-calculateAttr');
+                widget.input.promise.then(function () {
```
emord: I think current tests usually initialize the widget in the `before` call. Any reason you didn't do that here? I don't have strong opinions other than would be nice to be consistent

src/databrowser.js
```diff
@@ -138,11 +138,14 @@ define([
...
         function handleDrop(node, info, target) {
             var widget = widgets.util.getWidget(target, vellum),
-                path = widget.mug.form.richText ? node.hashtag : node.xpath;
+                path = widget.mug.form.richText ? node.hashtag : node.xpath,
+                id;
             while (info) {
-                var id = widget.addInstanceRef({id: info.id, src: info.uri});
-                if (id !== info.id) {
-                    path = widget.mug.form.updateInstanceQuery(path, id, info.id);
+                if (info.id && info.uri) {
```
emord: Commit message says that this fixes a bug. What was the bug and should it be tested?

tests/datasources.js
```diff
@@ -38,21 +38,95 @@ define([
                 "@id": {},
                 name: {}
             }
+        }, {
+            id: "commcaresession",
```
millerdev: b0baecfedf71b358e8ef8a52883e83ffc606fa34

src/datasources.js
millerdev: I think it could make sense if we ever needed it, but I'm reluctant to do that now since we don't.

src/datasources.js
```diff
@@ -300,7 +307,14 @@ define([
         }
         function getNodes(source, path, info) {
             var nodes = _.chain(source && source.structure)
-                .map(node(source, path, info))
+                .map(function (item, id) {
+                    if (item.merge) {
```
millerdev: The latter. `merge: true` was added because case properties and user properties have different base paths in the casedb (`/session/data` and `/session/context` respectively), but we don't want `data` and `context` to show up in the tree.

src/datasources.js
```diff
@@ -357,7 +362,7 @@ define([
             _.each(nodes, function (node) {
                 if (node.hashtag && !node.index) {
                     hashtags.map[node.hashtag] = node.xpath;
-                    hashtags.transforms[node.hashtagPrefix] = function (prop) {
+                    hashtags.transforms[node.sourceInfo.hashtag + '/'] = function (prop) {
```
millerdev: Do you mean an assertion? `node.sourceInfo.hashtag` will only ever have a hashtag value that was sent in the data sources schema, which is by definition a hashtag prefix.

src/logic.js
```diff
@@ -23,7 +23,7 @@ define([
             "hintItext",
         ],
         NO_SELF_REFERENCES = _.without(XPATH_REFERENCES, 'constraintAttr'),
-        CASE_REF_ID = "#case/";
+        EXTERNAL_REF = "#";
```
millerdev: This constant (variable) is only ever used as a key in the `forward` and `reverse` mappings. It is not (nor is it intended to be) used to match the beginning of a hashtag or path. It's simply a key that should never collide with a mug ufid, and it's value could be anything that meets that criteria. I figured `#` was good enough for that, but happy to entertain other suggestions.

tests/form.js
millerdev: I added these tests to verify things that previously worked continued to work after refactoring and happened to uncover these surprises. I think changing/fixing that is out of scope for this PR.

src/richText.js
millerdev: That's correct. The spec shows the same icons for case and user properties. @dimagi/product might have more comments on this?

src/form.js
millerdev: b0baecfedf71b358e8ef8a52883e83ffc606fa34

The input keys here are hashtag hashtag transformations, which are always populated from the hashtags supplied in the data sources schema (or in the short term, `#case/` guesses). I don't think there is any way a value like `#user+case` could get here unless it was specified in the schema, in which case we'd likely want it here.

tests/databrowser.js
millerdev: I see some tests do that and [some](https://github.com/dimagi/Vellum/blob/b8714350b636b5a4a05fed9d7a5d8dd5c912f815/tests/databrowser.js#L119) do it this way. In this case I wanted to reset the editor since I'm dragging a different value into the same property widget in the previous test. Is there something I'm overlooking?

src/databrowser.js
millerdev: It's hard to explain, but because of the `merge: true` flag, there is now an object in the `info._parent` hierarchy that does not have these attributes.

I had been trying to write a test to make this fail before adding `merge: true` to the test schemas, but was unable to. Once I added it, existing tests failed, and this fixes them.

tests/form.js
millerdev: There's a space after `case/ `.

src/datasources.js
emord: Ah. I see. so merge is for the UI layer, not the data layer? 

slightly related: May be useful to have better documentation for this builder function? I find myself having to really dig into the details whenver i look at it

src/logic.js
emord: Yeah. I think `#` is good enough for now, but i can see someone getting confused in the future. I think there's a comment in here about not splitting the case references as it gets rid of important inofrmation. This could be fixed if that is ever fixed

src/datasources.js
millerdev: Correct, merge is for the UI. I'll think about how to improve the documentation. Do you want to see that in this PR? The schema transformation is complicated, and the docstring at the top of the file is supposed to explain how it's interpreted. I wonder if the schema itself could be simplified so the code was easier to understand?

emord: I wasn't expecting it here. Just thinking out loud about how this function is always a little difficult to get spun up on

PR #828, opened 2016-12-08
Kill link to highlight question in tree

orangejenny:
This was done as an experiment, and it didn't see enough usage to justify keeping it: ~88,000 hovers over easy references, but only ~300 clicks on the link.

### DIFF
--- a/src/richText.js
+++ b/src/richText.js
@@ -38,7 +38,6 @@ define([
     'underscore',
     'jquery',
     'tpl!vellum/templates/date_format_popover',
-    'tpl!vellum/templates/easy_reference_popover',
     'vellum/dateformats',
     'vellum/escapedHashtags',
     'vellum/logic',
@@ -52,7 +51,6 @@ define([
     _,
     $,
     date_format_popover,
-    easy_reference_popover,
     dateformats,
     escapedHashtags,
     logic,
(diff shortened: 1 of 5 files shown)


src/richText.js
```diff
@@ -741,36 +739,12 @@ define([
...
-                var type = isFormRef ? 'form' : 'case';
-                analytics.fbUsage("Hovered over easy " + type + " reference");
-                analytics.workflow("Hovered over easy reference");
-                if (isDate || $this.attr("data-date-format")) {
-                    var pos = $(this).offset(),
-                        x = pos.left,
-                        y = pos.top + $(this).height();
-                    $("#" + dateFormatID).click(function () {
-                        $imgs.popover('hide');
-                        dateformats.showMenu(x, y, function (format) {
```
millerdev: Was it intentional that you also killed the date format changer? I think this also depends on the [Delay popover closing](https://github.com/dimagi/Vellum/pull/828/files#diff-0c6adbb645e08e26dfcc96c660406ee4) code.

emord:
Curious to know what the threshold was for keeping the feature. Was it a defined % of usage? Would also be interesting to control for how long the form is (what % of the form is displayed in the scroll bar), but that sounds fairly hard to do.

Disclaimer: I'm in favor of keeping this feature around for power users and long forms

orangejenny:
> Disclaimer: I'm in favor of keeping this feature around for power users and long forms

So am I! Dug into analytics more with @amsagoff and found that using the raw counts isn't the best proxy for what we really wanted to measure, which was of the people who tried clicking the link, how many found it useful enough to use it repeatedly. Those numbers are decent (~25%), but the sample size is small, so we're going to let the test run for longer and I'm closing out this PR.

PR #831, opened 2016-12-12
delay on easy reference popover

orangejenny:
http://manage.dimagi.com/default.asp?243816

### DIFF
--- a/src/richText.js
+++ b/src/richText.js
@@ -749,8 +749,8 @@ define([
                     (isFormRef ? '<div class="popover-content"><p></p></div>' : '') +
                     '</div></div>',
                 delay: {
-                    show: 0,
-                    hide: 200,
+                    show: 350,  // be less annoying
+                    hide: 200,  // allow time for user to move cursor into popover
                 },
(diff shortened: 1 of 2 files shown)


src/richText.js
```diff
@@ -749,8 +749,8 @@ define([
                     (isFormRef ? '<div class="popover-content"><p></p></div>' : '') +
                     '</div></div>',
                 delay: {
-                    show: 0,
-                    hide: 200,
+                    show: 350,  // be less annoying
```
emord: Slightly related: Does this mean that the analytics event is fired after 350 ms instead of instantaneously on hover?

orangejenny:
fixed tests

orangejenny:
@millerdev Do you remember why there's a `try` in those tests in the first place? Wondering if I should assume that an exception means the test failed, and then assert false inside the `catch`.

millerdev:
I think the `try`/`finally` is important for the popover to be removed, otherwise other tests that test popovers will fail.

orangejenny:
Fixed tests for real.

PR #832, opened 2016-12-14
Do not slice hashtag off of case references

millerdev:
https://github.com/dimagi/commcare-hq/pull/14221#issuecomment-267066954

@emord

### DIFF
--- a/tests/logic.js
+++ b/tests/logic.js
@@ -171,7 +171,7 @@ define([
                 var form = util.loadXML(MOTHER_REF_XML),
                     manager = form._logicManager;
                 assert.deepEqual(manager.caseReferences(),
-                    {load: {"/data/mug": ["parent/edd"]}});
+                    {load: {"/data/mug": ["#case/parent/edd"]}});
             });
 
             it("should not send deleted references", function () {
@@ -198,7 +198,7 @@ define([
                     ["/select/choice", "Choice", '<output value="#case/dob" />'],
                 ]);
                 assert.deepEqual(manager.caseReferences(),
-                                 {load: {"/data/select": ["dob"]}});
+                                 {load: {"/data/select": ["#case/dob"]}});
             });
 
             it("should send all properties referenced by a question", function () {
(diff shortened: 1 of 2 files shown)


tests/logic.js
```diff
@@ -171,7 +171,7 @@ define([
                 var form = util.loadXML(MOTHER_REF_XML),
                     manager = form._logicManager;
                 assert.deepEqual(manager.caseReferences(),
-                    {load: {"/data/mug": ["parent/edd"]}});
+                    {load: {"/data/mug": ["#case/parent/edd"]}});
```
emord: I think that the result of this will be the case summary shows `#case/parent/edd` and it currently shows `parent/edd` which probably isn't intended. That becomes worse once we send user properties as under a child case it could say `#user/property` which doesn't make sense to display

millerdev: Sorry, I don't understand what the problem is or what you're suggesting should be done differently. I think there must be some context I'm missing. As I understand it, `caseReferences()` returns (in the `load` value) a mapping of question paths to lists of referenced case properties, and this change makes case references less ambiguous.

emord: Sorry I didn't say that very well.

This is an issue with the Case Summary in HQ . The reason we remove the #case was to provide parity with old case management and how it saves case properties. Currently the Case Summary expects that form which has always been `property` or `parent/property`.

`caseReferences` is a method to provide HQ with loaded case properties to display in the Case Summary (https://github.com/dimagi/Vellum/pull/832/files#diff-e8c10ecfceaf2d2a578b7321207927c9R463)

If we send back `#case/dob` instead of `dob` we'll end up with something like 
image

In the future it could also show `#user/property` if we don't filter that, or come up with a different way to store these things.

I'm pro sending `#case/property`  and `#user/property`, so that these can be differentiated, but it will also require changes to the case summary in HQ. That could possibly be just removing `#case` when displaying in the summary, but there is probably a better way. The summary still needs to support the old form of `property` or `parent/property` for displaying saved case properties

millerdev:
Should be fine to merge this once https://github.com/dimagi/commcare-hq/pull/14221 is in.

emord:
@millerdev we forgot to merge this in after https://github.com/dimagi/commcare-hq/pull/14221 was merged. I think this might cause conflicts with #850

millerdev:
@emord OK, I'll fix it up after #850 gets merged

millerdev:
@emord fixed

PR #833, opened 2016-12-15
Fix drag/drop in data source widgets

millerdev:
http://manage.dimagi.com/default.asp?243450

Not particularly proud of this one. There are no tests because the data source widgets don't have tests. Also, they haven't been updated to support rich text. Hashtags do work, but no bubbles (obviously) and no auto-complete.

@emord


proteusvacuum:
🎉 Thanks!

PR #836, opened 2016-12-20
Dropdown for toggling section visibility

orangejenny:
Haven't yet implemented logic to show a section if it contains any values, even if it user preferences say to hide it. Not certain we should do this, as it makes the show/hide behavior harder to predict.

Wait on UAT.

screen shot 2016-12-20 at 5 44 39 pm

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1781,12 +1781,16 @@ define([
         }
     };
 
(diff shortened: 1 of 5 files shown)


orangejenny:
Closing in favor of https://github.com/dimagi/Vellum/pull/840

PR #839, opened 2016-12-27
Analytics for tools menu items

orangejenny:
@biyeun


orangejenny:
nm, this exists, i just hadn't found the code yet

PR #840, opened 2016-12-28
Onboarding milestone 2: question-level changes

orangejenny:
- Drop "Basic" header
- Menu to show/hide non-Basic sections
- Move required checkbox into Basic
- Move question ID below label
- Auto-generate question ID based on label

Need to
- [ ] fix/update/write tests
- [x] respond to PR feedback below

Overall:
screen shot 2017-01-06 at 6 34 52 pm

Menu to toggle sections:
screen shot 2017-01-06 at 6 34 56 pm


src/core.js
```diff
@@ -1545,6 +1589,20 @@ define([
...
+                "class": "fa fa-trash-o action_remove",
+                "after": true,
+                "selector": "a",
+                "event": "click",
+                "callback": function (node_id, node, action_id, action_el) {
+                    _this.data.core.form.removeMugsFromForm([node.data.mug]);
+                    _this.refreshCurrentMug();
+                }
+            });
+        }
```
millerdev: Looks like the `return` was removed, and probably needs to be re-added here.

src/core.js
```diff
@@ -1579,14 +1635,28 @@ define([
...
+                })
                 .filter(_.identity);
 
             if (section.properties.length) {
                 this.getSectionDisplay(mug, section).appendTo($content);
             }
         }
 
+        this.$f.find(".fd-section-changer").html(section_changer({
+            sections: _.map(_.filter(_.rest(_this.getSections(mug)), function(s) {
```
millerdev: This might be a good place to use [`_.chain()`](http://underscorejs.org/#chain).

src/core.js
```diff
@@ -1819,20 +1894,13 @@ define([
...
-                isDeleteable: mugs && mugs.length && _.every(mugs, function (mug) {
-                    return _this.isMugRemoveable(mug, mug.hashtagPath);
-                }),
                 isCopyable: !multiselect && mug.options.isCopyable,
             }));
-        $baseToolbar.find('.fd-button-remove').click(function () {
-            var mugs = _this.getCurrentlySelectedMug(true, true);
-            form.removeMugsFromForm(mugs);
-            _this.refreshCurrentMug();
-        });
```
millerdev: Does this mean it's no longer possible to bulk-delete questions?

orangejenny:
> The delete button in the tree is pretty close to the expansion triangle for groups. I'm concerned that it could be clicked accidentally. Do you think that will be an issue?

That's a good point. I never use the expand/collapse, just tried it, and yeah, it's too close to the trash can. I'll play with better spacing and some indication of whether you're going to click on the trash can or the triangle.

millerdev:
I think I saw an earlier mockup where the delete button was on the right, and I think that would be my preference. Is there a reason why that was abandoned?

src/core.js
orangejenny: It does...which isn't good. I'll put the delete button back for the multiple-questions-selected window at least (where the copy-paste help is shown).

I'm debating whether or not to put it back in question properties, too. I've been wondering if it would be good to get rid of the top bar in question properties altogether. Suspect it depends on how often people use the question type changer.

millerdev: Oh, if it's in the multiple-questions-selected view then I think that's fine since it's still possible to delete multiple. We could also enable the *Cut* command...

orangejenny:
Icons on the right don't work well when labels are long and the question tree scrolls horizontally. But I could try cutting off the labels with ellipses and then moving the trash cans back to the right. I'm skeptical that people get a lot of value out of scrolling horizontally within the question tree.

orangejenny:
Made changes after talking with @amsagoff :
- Took delete action out of tree altogether (it doesn't look great on the right side either, decided to punt for now since it isn't obvious where to put it and it isn't crucial to onboarding)
- Moved the section changer out of the dark blue header and turned it back into a button

Updated screenshots, fyi @biyeun

orangejenny:
Also made the section headers no longer accordions.

millerdev:
> Also made the section headers no longer accordions.

What was the motivation for this? I rarely like accordions, but in this case I didn't hate them. I'll need to play with it a bit, but here's a scenario that seems worse with this change:

- Media section is expanded/enabled. Logic and Advanced are collapsed/hidden.
- Properties are scrolled to bottom of Media, and the section menu is scrolled off the top of the page.
- What do I need to do to get to Advanced?
  - Scroll to top, use section menu to enable/expand Advanced, scroll back down to Advanced properties.
- What do I need to do to get to Logic?
  - Scroll to top, use section menu to enable/expand Logic, (on smaller screen) scroll back down to Logic properties.

This feels like a step back. Maybe I just need to try/get used to the new way?

millerdev:
Responding to my own question ("What was the motivation for this?") above. I see from [your comment on #841](https://github.com/dimagi/Vellum/pull/841#discussion_r95044532) that it's about removing distractions, which is reasonable, although maybe regrettable for more seasoned form builders.

orangejenny:
My hope is that more advanced form builders will be using logic and/or advanced for most of their apps/forms/questions, so they'll open those sections once and won't close them (this PR doesn't interfere with the section stickiness added a few months back).

The people who would be negatively affected are those who like the open and close sections often. If we get feedback along those lines, another way to deal with it could be making the type changes / delete / section menu bar fixed at the top of the section.

biyeun:
screen shots look good. only thing that's a bit strange is that the more menu is right after the delete button. I feel like having the more menu to the right of the type button is a more natural grouping.

orangejenny:
Closing in favor of https://github.com/dimagi/Vellum/pull/847

PR #841, opened 2016-12-28
Onboarding milestone 3: top bar changes

orangejenny:
- Only one "add question" button
- Move "add choice" into tree
- Move tools into languages menu

Opening this largely for discussion with @amsagoff , probably not worth reviewing at this point (also, branched off of [onboarding-2](https://github.com/dimagi/Vellum/tree/onboarding-2), hence the many commits). The save command doesn't work yet, and tests need to be fixed/updated.

Overall look:
screen shot 2017-01-06 at 7 22 38 pm

"Add Question" control:
screen shot 2017-01-06 at 7 22 52 pm

"Add Choice" in tree:
olb1hm6ve4

Tools moved into menu:
screen shot 2017-01-06 at 7 23 02 pm


src/core.js
```diff
@@ -241,63 +204,85 @@ define([
...
-            if (groupData.related && groupData.related.length) {
-                groupData.related = _.map(groupData.related, getQuestionData);
-            }
+        });
 
-            groupData.group[2] = groupData.group[2] || 
-                _this.data.core.mugTypes[groupData.group[0]].icon;
-            $questionGroupContainer.append(
-                new QuestionTypeGroup(groupData, _this));
+        _this.$f.find(".fd-add-question").after($(add_question({
```
millerdev: minor: looks like there's a variable above called `$addQuestion` that already has this node.

src/core.js
```diff
@@ -403,11 +390,17 @@ define([
             });
         });
 
+        // Section toggling events: within section changer and when clicking on section header
```
millerdev: I think this means there are now two ways to expand/collapse a section group: clicking the section header and also new items in the question properties menu. Is that right? Out of curiosity, how was it determined that the items in the question properties menu were needed?

orangejenny: I misunderstood why we were doing this and didn't implement it right. Have now changed the section headers to no longer be clickable, and now when you hide a section, it removes both the section and the header. This makes the basic experience, where you're only using the basic section, more manageable (you don't see the headers for Logic, etc. and therefore don't   click on them and get distracted). We're planning to user test this next week to see if it's helpful.

orangejenny:
Made changes after discussion with @amsagoff :
- Reinstated the top bar, which now just has "Add Question" and "Save" (and will later have "Preview" as well)
- Split lookup tables out into their own question group
- Removed the "New Form" content that shows on blank forms

Updated screenshots, fyi @biyeun (and I'd also welcome feedback on the screenshots/direction from @millerdev )

biyeun:
here's a thought to throw out there... what if there was an additional companion button next to the add question button that had the icon for the most recently added question. Perhaps you want tot add a lot of text questions one after the other. Otherwise this is a good start. 

Mulling over some redesign possibilities here in terms of having the style of formbuilder fit more closely with app manager v2, but I think that's safe for another PR. Let's see if this functionality makes sense first.

orangejenny:
Closing in favor of https://github.com/dimagi/Vellum/pull/847

PR #843, opened 2017-01-04
Fix popover bug causing expression editor fail

millerdev:
http://manage.dimagi.com/default.asp?244242

@orangejenny cc @benrudolph


orangejenny:
thank you for fixing

PR #845, opened 2017-01-10
improve comments

emord:
@proteusvacuum @orangejenny http://manage.dimagi.com/default.asp?244064

not all of the improvements listed, but it's the easy wins

buddy @benrudolph

### DIFF
--- a/src/mugs.js
+++ b/src/mugs.js
@@ -982,10 +982,6 @@ define([
                 lstring: 'Comment',
                 visibility: 'visible',
                 widget: widgets.multilineText,
-                setter: function (mug, attr, value) {
-                    mug.p.set(attr, value);
-                    $('.fd-question-comment').text(value);
-                },
             }
         },
 
--- a/src/mugs.js
+++ b/src/mugs.js
@@ -982,6 +982,10 @@ define([
                 lstring: 'Comment',
                 visibility: 'visible',
                 widget: widgets.multilineText,
+                setter: function (mug, attr, value) {
+                    mug.p.set(attr, value);
+                    $('.fd-question-comment').text(value);
+                },
             }
         },
 
--- a/src/widgets.js
+++ b/src/widgets.js
@@ -267,12 +267,6 @@ define([
             return widget.input.val();
         };
 
-        mug.on("property-changed", function (e) {
-            if (e.property === "comment") {
-                $('.fd-question-comment').text(widget.mugValue(mug));
-            }
-        }, null, "teardown-mug-properties");
-
         return widget;
     };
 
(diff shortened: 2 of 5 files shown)


src/mugs.js
```diff
@@ -982,6 +982,10 @@ define([
                 lstring: 'Comment',
                 visibility: 'visible',
                 widget: widgets.multilineText,
+                setter: function (mug, attr, value) {
+                    mug.p.set(attr, value);
+                    $('.fd-question-comment').text(value);
```
millerdev: ooh, that's brutal, coupling the mug (model) directly with the UI. What do you think of making the comment UI listen for the form `mug-property-change` event intsead?

orangejenny:
agree with Daniel's comment, otherwise looks good

src/mugs.js
emord: D'oh not sure what I was thinking.

https://github.com/dimagi/Vellum/pull/845/commits/7f8341908a1a1dad88962285588df0075c95d2c1

Still not 100% pleased with where I put it, but decided it was alright as we do something similar in the same file and it's the file that handles the UI layer for the inputs anyways https://github.com/dimagi/Vellum/blob/edea3c6b121a1dc4114d825265c73fe88e793aaf/src/widgets.js#L351-L355

src/widgets.js
```diff
@@ -267,6 +267,12 @@ define([
             return widget.input.val();
         };
 
+        mug.on("property-changed", function (e) {
```
millerdev: Hmm, another dependency on `property-changed`, which is ["legacy"](https://github.com/dimagi/Vellum/blob/7f8341908a1a1dad88962285588df0075c95d2c1/src/form.js#L890-L892). I don't really have a better solution, so not saying it should be done differently, just pointing it out. I think I did a similar thing elsewhere in this module.

Musing here: wonder if we should remove that "legacy" comment and formally endorse usage of this event, especially now that we have a mechanism for removing event listeners on `teardown-mug-properties`? Also wondering if that was the reason why Mike wanted to avoid more dependencies on `property-changed` or if there was some other reason?

emord: Oh. I didn't realize it was "legacy". I don't think I know enough about the context of this one to suggest another option either. 

> I think I did a similar thing elsewhere in this module.

Yep. that's why I used this (link in my last comment)

orangejenny: > wonder if we should remove that "legacy" comment and formally endorse usage of this event

This sounds eminently reasonable.

proteusvacuum:
🙌

millerdev:
The test failure is regrettably real (lint):
```
tests/comments.js: line 61, col 48, Missing semicolon.
```

PR #847, opened 2017-01-27
Onboarding

orangejenny:
@millerdev / @emord 

This combines https://github.com/dimagi/Vellum/pull/840 and https://github.com/dimagi/Vellum/pull/841 into one PR with a reasonable commit history.

The current plan is to do a few more usability tests before green-lighting release, but I'm fairly confident that it's in good shape. So I think it's a good time to review (but would understand if y'all want to wait to read this until we're for real about to release).

@millerdev you've already reviewed a fair amount of this...but it's also had substantial changes since then, I don't know that there's a good way to get around re-reading some code. I did incorporate the comments that you made in https://github.com/dimagi/Vellum/pull/840

code buddy @mkangia 

QA ticket: http://manage.dimagi.com/default.asp?247101


tests/richText.js
```diff
@@ -553,9 +553,13 @@ define([
...
                     widget.setValue('testing cursor');
-                    editor.focus();
-                    var selection = editor.getSelection(true);
-                    assert.strictEqual(selection.getNative().focusOffset, 14);
+                    // Make sure focus is elsewhere, then focus on the rich text input
+                    editor.on('instanceReady', function() {
+                        $('[name=property-nodeID]').focus();
+                        editor.focus();
+                        var selection = editor.getSelection(true);
+                        assert.strictEqual(selection.getNative().focusOffset, 14);
```
millerdev: what does `14` mean? maybe add a comment?

src/form.js
```diff
@@ -802,7 +802,32 @@ define([
...
+            var _this = this;
+            _this.mugTypes.changeType(mug, questionType);
+            _this.setMugActions(mug);
+        },
+        setMugActions: function(mug) {
+            var _this = this,
+                tree = _this.vellum.data.core.$tree,
+                action_id = "add_choice";
+
+            if (mug.__className === 'Select' || mug.__className === 'MSelect') {
```
millerdev: Rather than hard-coding mug type names here, what do you think of adding attributes to the mug type options (like `mug.options.canAddChoices`)?

src/form.js
```diff
@@ -802,7 +802,32 @@ define([
...
+                tree.jstree(true).add_action(mug.ufid, {
```
millerdev: I think this UI code should not be in the form object. It should probably be moved into core.js near where the question tree UI is configured. A separate, but maybe related, issue is that core.js is getting pretty big and contains a lot of things. Maybe we should move the question tree into its own module (probably not in this PR)?

src/core.js
```diff
@@ -245,47 +197,63 @@ define([
...
         });
+    };
 
+    fn._init_toolbar = function () {
         var $saveButtonContainer = this.$f.find('.fd-save-button');
         this.data.core.saveButton.ui.appendTo($saveButtonContainer);
     };
 
-    fn._getQuestionGroups = function () {
+    fn.getQuestionGroups = function () {
```
millerdev: I'm not sure if it matters here, but I think this change has some effect in the plugin system (see [base.js:35](https://github.com/dimagi/Vellum/blob/5960d9d3ed90d11a327ec5488c45667a83599fee/src/base.js#L35))

src/less-style/structure.less
```diff
@@ -177,25 +183,3 @@ body.vellum-full-screen {
         }
     }
 }
-
-// TODO: clean up
```
millerdev: I like to see this kind of clean up (just delete it) 😄

bower.json
```diff
@@ -35,6 +35,7 @@
 
         "bootstrap": "3.3.7",
         "jstree": "~3.3.3",
+        "jstree-actions": "orangejenny/jstree-actions#make-module",
```
emord: This can be set back to the original repo. May be worth asking the guy to make a tag first though

src/core.js
```diff
@@ -1445,9 +1445,19 @@ define([
...
+            if ($firstGroup.length) {
+                var $input = $firstGroup.find("input, textarea");
+                if ($input.length) {
+                    // Rich text is off
+                    $input.focus();
+                } else {
+                    // Rich text is on
+                    $input = $firstGroup.find(".fd-textarea, .fd-input");
+                    richText.editor($input).focus();
+                }
```
emord: I think that the `.form-group` takes care of selecting the first widget. I don't think we have any widgets that have two inputs except for the javarosa itexts

tests/richText.js
emord: Looks like the end of the editor (This was one of my original tests). I think ckeditor has some magic constants that can be used for end of input instead of having this number

src/form.js
```diff
@@ -802,7 +802,32 @@ define([
             }
         },
         changeMugType: function (mug, questionType) {
-            this.mugTypes.changeType(mug, questionType);
+            var _this = this;
```
emord: nit: I don't think you need to use `_this`, because you aren't changing contexts anywhere in this function

tests/utils.js
```diff
@@ -492,6 +492,13 @@ define([
                 }
                 mug.p[name] = val;
             });
+
+            // Open all sections
```
emord: I'm not excited about this as it means that most tests are run from a non-default setup. With that said, I wouldn't volunteer to go through each test to fix this either

src/core.js
```diff
@@ -2080,7 +2080,7 @@ define([
                 properties: this.getMainProperties(),
                 help: {
                     title: "Basic",
-                    text: "<p>The <strong>Label</strong> is text that appears in the application. " +
+                    text: "<p>The <strong>Display Text</strong> appears in the application. " +
```
orangejenny: This was an Amelia suggestion; apparently it used to be `Display Text` before it was `Label`. I don't have a strong preference for either - I don't think there's a good word to use here that'll convey the concept to everyone - but I do like getting rid of the "Display Text" placeholder.

src/core.js
```diff
@@ -1188,9 +1188,26 @@ define([
...
+        return mug.form.generate_question_id(suggestedID, mug);
+    };
+
     // Attempt to guard against doing actions when there are unsaved or invalid
     // pending changes.
     fn.ensureCurrentMugIsSaved = function (callback) {
+        var currentMug = this.getCurrentlySelectedMug();
+        if (currentMug && !currentMug.p.nodeID) {
+            currentMug.p.nodeID = this.nodeIDFromLabel(currentMug);
+        }
```
orangejenny: thanks, that's a good point

src/core.js
orangejenny: Interesting, didn't know about that. I think this code is okay, though.

src/form.js
orangejenny: Yeah, not necessary, but I like using `_this` as a habit. Otherwise I get bitten when making code more complex later on.

tests/utils.js
orangejenny: Yeah...I see your point, but I am also not eager to change all of those tests. Looking towards the future, if we change the default sections again, tests that fail because some element wasn't visible can be pretty obnoxious to track down.

src/core.js
millerdev: > but I do like getting rid of the "Display Text" placeholder.

👍

millerdev:
Looks good, but can't merge.

src/form.js
orangejenny: Does https://github.com/dimagi/Vellum/pull/847/commits/2042113ca7e23288f5308bc86293c51baa364893 strike you as an improvement?  I'll take a stab at separating tree logic into its own file, but yeah, that might not be part of this PR.

millerdev: Yes, looks great!

orangejenny:
@emord I added the menu-aim plugin you suggested in https://github.com/dimagi/Vellum/pull/847/commits/e2f3917280420b36572c479a383f6e65816d25ec  Not convinced that it makes the menu feel much better, but having it there isn't hurting anything.

orangejenny:
@emord removed it

millerdev:
merged :tada: woohoo!

PR #848, opened 2017-01-31
add bower/npm update to readme

czue:
@orangejenny let me know if this isn't the right place/language for this.


orangejenny:
good call

orangejenny:
docs only, killed build

PR #850, opened 2017-02-01
add case references from save to case questions

czue:
@emord

### DIFF
--- a/src/logic.js
+++ b/src/logic.js
@@ -500,10 +500,7 @@ define([
                 );
                 save[mugPath] = {
                     case_type: mug.p.case_type || '',
-                    properties: _.filter(propertyNames, function (property) {
-                        // filter out empty properties
-                        return Boolean(property);
-                    }),
+                    properties: _.filter(propertyNames, _.identity), // filter out empty properties
                     create: mug.p.useCreate || false,
                     close: mug.p.useClose || false,
                 };
--- a/src/logic.js
+++ b/src/logic.js
@@ -490,20 +490,11 @@ define([
                 }
             });
             var saveToCaseQuestions = _.filter(this.form.getMugList(), function(mug) {
-                return mug.__className === "SaveToCase";
+                return mug.options.getCaseSaveData !== undefined;
             });
             _.each(saveToCaseQuestions, function (mug) {
                 var mugPath = _this.form.getAbsolutePath(mug);
-                var propertyNames = _.union(
-                    _.keys(mug.p.createProperty || {}),
-                    _.keys(mug.p.updateProperty || {})
-                );
-                save[mugPath] = {
-                    case_type: mug.p.case_type || '',
-                    properties: _.filter(propertyNames, _.identity), // filter out empty properties
-                    create: mug.p.useCreate || false,
-                    close: mug.p.useClose || false,
-                };
+                save[mugPath] = mug.options.getCaseSaveData(mug);
             });
             return {load: load, save: save};
         },
--- a/src/logic.js
+++ b/src/logic.js
@@ -463,8 +463,10 @@ define([
         // This is to tell HQ's case summary what is referenced
         // TODO maybe do not slice prop because it removes unrecoverable context
         caseReferences: function () {
+
             var _this = this,
-                load = {};
+                load = {},
+                save = {};
             _.each(_.flatten(_.values(this.reverse[EXTERNAL_REF] || {})), function(ref) {
                 var prop = ref.path.slice(ref.path.indexOf('/') + 1),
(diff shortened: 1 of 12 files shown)


src/logic.js
```diff
@@ -487,8 +489,26 @@ define([
                     load[path] = [prop];
                 }
             });
-
-            return {load: load};
+            var saveToCaseQuestions = _.filter(this.form.getMugList(), function(mug) {
+                return mug.__className === "SaveToCase";
```
millerdev: Update: minor edits for clarity of above comment.

czue: I'll PR the second half of this now, but main use case was to get this info in app summary, exports, and data dictionary which is a huge win for enterprise and GS efficiency. I had actually originally thought i'd do it in HQ and @emord convinced me to do it here. [some discussion here](https://docs.google.com/document/d/1VE0vtYNiR0xkyXjos9YfVgH6f6Dii4kxLnhnqT3ELuI/edit#)

czue:
sorry commented in the wrong place. @millerdev 

I'll PR the second half of this now, but main use case was to get this info in app summary, exports, and data dictionary which is a huge win for enterprise and GS efficiency. I had actually originally thought i'd do it in HQ and @emord convinced me to do it here. some discussion here

czue:
er - sorry didn't read your comment properly before responding.

> The SaveToCase mug type is not how we want to support saving to case in the long run, and building out more functionality around it rather than solving the problem in a way that conforms with our long-term vision seems like a not great idea

I believe save to case is not going away any time soon and this is a huge boon for GS and devs who maintain complex apps so I disagree with the premise that we shouldn't improve our support and tooling around until we have a plan that has a concrete path for not only doing case management saving in formbuilder, but also getting legacy save to case migrated.

> All of the discussions the form builder team has had on how to support save to case have been very preliminary and high-level and more recently it's been moved to the back burner in favor of other efforts to make case management in velllum better for the product in other ways. 

This makes perfect sense to me and I don't believe is at odds with this change at all. As in, we should be explicitly thinking about the two differently from a product perspective as they fulfill entirely different use cases.

> On the other hand, this might be a good start at laying the groundwork for building out save to case in a way that all mug types can use.

I think this is what @emord hoped would be true and is also well-aligned with other GS initiatives in flight (e.g. data dictionary)

millerdev:
@czue thanks for the context. I'm on board now, and think it's fine to move forward with this. I agree, this is high value for GS, so makes sense to do it now. My main concern is tighter coupling of the SaveToCase mug type with other parts of vellum. However, I think this can be minimized if we build out a standardized interface that can be used by all mugs as I was describing in https://github.com/dimagi/Vellum/pull/850#discussion_r98896018

millerdev:
I also agree that Vellum is the right place to do this.

src/logic.js
czue: > To simplify the interface, it might make sense to add a function (something like `mug.options.getCaseSaveData(mug)`) to the mug options object that returns `{case_type: ..., properties: [...], create: ..., close: ...}` so the implementation below, which looks specific to the SaveToCase mug type, can be moved into saveToCase.js. If you go that route, the presence of that function could be the flag used here.

I really like this idea - will give it a shot.

src/logic.js
```diff
@@ -487,8 +489,26 @@ define([
...
+                var propertyNames = _.union(
+                    _.keys(mug.p.createProperty || {}),
+                    _.keys(mug.p.updateProperty || {})
+                );
+                save[mugPath] = {
+                    case_type: mug.p.case_type || '',
+                    properties: _.filter(propertyNames, function (property) {
+                        // filter out empty properties
+                        return Boolean(property);
+                    }),
```
czue: https://github.com/dimagi/Vellum/pull/850/commits/2ef7c83373854a3c8ac37a75978616bd87f3ae7f

src/logic.js
czue: is this ~what you had in mind? https://github.com/dimagi/Vellum/pull/850/commits/3e29c61b3df6a9e3f82470421dc4fecc38d1a6ad

millerdev: yes, perfect!

src/logic.js
```diff
@@ -463,8 +463,10 @@ define([
         // This is to tell HQ's case summary what is referenced
         // TODO maybe do not slice prop because it removes unrecoverable context
         caseReferences: function () {
+
```
emord: nit: why the extra space?

czue: no reason. https://github.com/dimagi/Vellum/pull/850/commits/7b3e669fdd1de84e9020a02bd9cbf24dfb21b812

PR #851, opened 2017-02-02
Case property descriptions

millerdev:
The other half of https://github.com/dimagi/commcare-hq/pull/14718 - should be safe to merge in any order.

@calellowitz @emord

### DIFF
--- a/src/datasources.js
+++ b/src/datasources.js
@@ -52,6 +52,7 @@
  *              },
  *              @attribute: {
  *                  name: string (optional human readable name),
+ *                  description: string (optional description of the element),
  *              }
  *          },
  *          subsets: [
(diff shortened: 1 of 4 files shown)


millerdev:
@emord bump

src/datasources.js
```diff
@@ -306,7 +314,11 @@ define([
             };
         }
         function getTree(item, id, path, info) {
-            var tree = {name: item.name || id, recursive: false},
+            var tree = {
+                    name: item.name || id,
+                    description: item.description || '',
+                    recursive: false
+            },
```
emord: nit: weird indentation here

millerdev: indeed. Not sure how it got that way, but I'll fix it the next time I visit that code. Going to merge now on account of how long it takes to run tests these days.

PR #854, opened 2017-02-14
Refresh hqstyle

orangejenny:
Realized we haven't done this in a while, so standalone was missing the new purple button styling.

The vast majority of `hqstyle-core.css` changes are no-ops, either whitespace or unnecessary semicolons. Whenever we next refresh this file, provided we use the same unminifier, the diff should be more reasonable. For this one, it's probably easiest to ignore the first two commits and just review the later 4 commits, which are mostly just docs.


tests/lib/hqstyle-core.css
```diff
@@ -1,3 +1,16 @@
+/*
+ * HQ styling, for the sake of standalone vellum looking like HQ vellum.
+ *
+ * To create this file:
+ *  1) On HQ, go to form builder and View Source on the page.
+ *  2) Grab the HQ stylesheets linked from the page. At the time of writing, there were
+ *     two relevant files (one for app manager styling, one for the rest of HQ).
+ *  3) Run the stylesheets through a minifier like http://unminify.com/ and overwrite this file.
```
millerdev: Should say "through _an unminifier_"?

tests/lib/hqstyle-core.css
```diff
@@ -293,19 +293,13 @@ th {
...
-format('embedded-opentype'),
-url('https://dnwn0mt1jqwp0.cloudfront.net/static/bootstrap/fonts/glyphicons-halflings-regular.woff2?448c34a56d69')
-format('woff2'),
-url('https://dnwn0mt1jqwp0.cloudfront.net/static/bootstrap/fonts/glyphicons-halflings-regular.woff?fa2772327f55')
-format('woff'),
-url('https://dnwn0mt1jqwp0.cloudfront.net/static/bootstrap/fonts/glyphicons-halflings-regular.ttf?e18bbf611f2a')
-format('truetype'),
-url('https://dnwn0mt1jqwp0.cloudfront.net/static/bootstrap/fonts/glyphicons-halflings-regular.svg?89889688147b#glyphicons_halflingsregular')
-format('svg')
+    src: url('/static/style/lib/bootstrap-3.2.0/fonts/glyphicons-halflings-regular.eot?f4769f9bdb74');
```
millerdev: how does this work with stand-alone vellum? is it just not found, or does it need to be in a specific place relative to the HQ repo for it to work?

tests/lib/hqstyle-core.css
orangejenny: oops, fixed

tests/lib/hqstyle-core.css
orangejenny: It's not found in standalone. This is for bootstrap 3 icons, which aren't used in vellum - we use either fontawesome or custom HQ icons.

I could have left these links alone, but having cloudfront references seemed weird.

millerdev: When I load vellum stand-alone I now get two 404 errors that I did not get before:
```
GET http://localhost:8081/static/style/font/commcarehqfont-regular.woff?v=1.1&d68065ecb660 
GET http://localhost:8081/static/style/font/commcarehqfont-regular.ttf?v=1.1&b7dbbd6d243e 
```

Is there a way to make it so these errors do not occur? I don't like to get into a state where I'm ignoring errors in my dev environment.

orangejenny: fixed

PR #855, opened 2017-02-21
Fix pasted styles leaking into editor content

millerdev:
Happens when pasting text copied from Google Sheets.
http://manage.dimagi.com/default.asp?245729

@orangejenny @emord

### DIFF
--- a/src/richText.js
+++ b/src/richText.js
@@ -269,6 +269,12 @@ define([
             }
         });
 
+        // Adapted from http://www.keyvan.net/2012/11/clean-up-html-on-paste-in-ckeditor/
+        var STYLE = /<style type="text\/css">.*?<\/style>/g;
+        editor.on('paste', function(event) {
+            event.data.dataValue = event.data.dataValue.replace(STYLE, "");
+        }, null, null, 2 );
+
         if (_.isFunction(options.createPopover)) {
             editor.addCommand('createPopover', {
                 exec: options.createPopover,
--- a/tests/richText.js
+++ b/tests/richText.js
@@ -460,6 +460,12 @@ define([
                 });
             });
 
+            it("should not paste style content into editor", function () {
+                var html = '<style type="text/css"><!--td--></style><span>A</span>';
+                input.ckeditor().editor.execCommand('paste', html);
+                assert.equal(editor.getValue(), 'A');
+            });
+
             function applyArgs(func) {
                 return function (args) {
                     return func.apply(this, args);


emord:
Is this something that happens on all ckeditor instances or is it something special about hte way we use it?

orangejenny:
Curious why this doesn't need to strip out the `<span style="font-size:13px;...` that's in dataValue, does ckeditor take care of that?

millerdev:
@emord asked:
> Is this something that happens on all ckeditor instances or is it something special about hte way we use it?

Good question, and I'm not sure. My guess is it has something to do with us setting `CKEDITOR.config.allowedContent = true`, which turns off ckeditor's content filter.

@orangejenny asked:
> Curious why this doesn't need to strip out the `<span style="font-size:13px;...` that's in dataValue, does ckeditor take care of that?

Yes, it removes `<span>` tags without extra help, so I didn't touch them. CKEditor dev often feels a bit like voodoo to me: poke it here and there and magic happens.

orangejenny:
> CKEditor dev often feels a bit like voodoo to me: poke it here and there and magic happens.

Agree with that.

Regexing for HTML markup feels fragile, but I don't have a definitively better approach. Might consider doing something like `$("<div>").append(event.data.dataValue).text()` ...but that might break people pasting in legitimate HTML, if there is such a thing.

src/richText.js
```diff
@@ -269,6 +269,11 @@ define([
             }
         });
 
+        var STYLE = /<style type="text\/css">.*?<\/style>/g;
+        editor.on('paste', function(event) {
+            event.data.dataValue = event.data.dataValue.replace(STYLE, "");
+        }, null, null, 2 );
```
orangejenny: Didn't notice these arguments on first reading, what's the 2?

millerdev:
well we at least want to preserve `<output>` tags, so I think we can't do `$("<div>").append(event.data.dataValue).text()`.

I intended to include an attribution comment, but accidentally deleted it before creating the PR. This solution was adapted from http://www.keyvan.net/2012/11/clean-up-html-on-paste-in-ckeditor/

millerdev:
The way I'm looking at this is it fixes a very specific issue (see linked FB case) that happens when pasting from Google Sheets into Vellum, which is something our users do frequently. It's true that it is fragile, but I can't think of a better way.

src/richText.js
millerdev: See last comment on http://www.keyvan.net/2012/11/clean-up-html-on-paste-in-ckeditor/ (and my voodoo comment)

orangejenny:
Yeah, that all makes sense to me.

PR #857, opened 2017-02-22
Make undo alert red when deleted questions are still referenced

millerdev:
First part of http://manage.dimagi.com/default.asp?242658
Best reviewed by commit 🐠 

@orangejenny @emord

### DIFF
--- a/src/form.js
+++ b/src/form.js
@@ -984,19 +984,13 @@ define([
             this._logicManager.forEachBrokenReference(updateReferences);
         },
         isReferencedByOtherMugs: function (mug, except) {
-            var ref = false,
-                ufids = {},
-                exclude = _.object(_.map(except, function (mug) {
+            var exclude = _.object(_.map(except, function (mug) {
                     return [mug.ufid, null];
                 }));
             exclude[mug.ufid] = null;
-            ufids[mug.ufid] = null;
-            this._logicManager.forEachReferencingProperty(ufids, function (mug) {
-                if (!exclude.hasOwnProperty(mug.ufid)) {
-                    ref = true;
-                }
+            return this._logicManager.hasReferencingMug(mug, function (mug) {
+                return !exclude.hasOwnProperty(mug.ufid);
             });
-            return ref;
         },
         /**
          * Get the logical path of the mug's node in the data tree
--- a/src/form.js
+++ b/src/form.js
@@ -983,6 +983,21 @@ define([
             var _this = this;
             this._logicManager.forEachBrokenReference(updateReferences);
         },
+        isReferencedByOtherMugs: function (mug, except) {
+            var ref = false,
+                ufids = {},
+                exclude = _.object(_.map(except, function (mug) {
+                    return [mug.ufid, null];
+                }));
+            exclude[mug.ufid] = null;
+            ufids[mug.ufid] = null;
+            this._logicManager.forEachReferencingProperty(ufids, function (mug) {
+                if (!exclude.hasOwnProperty(mug.ufid)) {
+                    ref = true;
+                }
+            });
+            return ref;
+        },
         /**
          * Get the logical path of the mug's node in the data tree
          *
--- a/src/form.js
+++ b/src/form.js
@@ -1037,35 +1037,39 @@ define([
             }
             var _this = this,
                 seen = {},
-                ufids = {},
-                undoUfids = {};
+                ufids = {};
             this.undomanager.resetUndo();
-            _.each(mugs, function (mug) {
-                _this._addToUndoManager(mug, undoUfids);
-            });
+            _this._addToUndoManager(mugs);
             _.each(mugs, function (mug) {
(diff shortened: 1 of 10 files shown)


src/form.js
```diff
@@ -1022,35 +1037,39 @@ define([
...
+                    hasChildren = false,
+                    previousSibling = mug.previousSibling,
+                    position = previousSibling === parentMug ? 'first' : 'after';
+                stack.push([mug, previousSibling, position]);
+                if (node) {
+                    var children = node.getChildrenMugs();
+                    hasChildren = children.length > 0;
+                    for (var i = 0; i < children.length; i++) {
+                        pushUndoItem(children[i], ufids);
+                    }
```
millerdev: This didn't change much. Best [view with `w=1`](https://github.com/dimagi/Vellum/pull/857/commits/010cada40cc75fde9d093dd1087dc526f685a968?w=1)

src/form.js
```diff
@@ -983,6 +983,21 @@ define([
             var _this = this;
             this._logicManager.forEachBrokenReference(updateReferences);
         },
+        isReferencedByOtherMugs: function (mug, except) {
```
emord: nit: seems like a better fit for the logic manager

src/form.js
```diff
@@ -983,6 +983,21 @@ define([
...
+            var ref = false,
+                ufids = {},
+                exclude = _.object(_.map(except, function (mug) {
+                    return [mug.ufid, null];
+                }));
+            exclude[mug.ufid] = null;
+            ufids[mug.ufid] = null;
+            this._logicManager.forEachReferencingProperty(ufids, function (mug) {
+                if (!exclude.hasOwnProperty(mug.ufid)) {
+                    ref = true;
```
emord: Would be nice if you could break out of this loop. I imagine that for large forms there's going to be a lot of properties referencing each other

src/undomanager.js
```diff
@@ -7,7 +7,7 @@ define([
     $,
     _,
     util,
-    UNDO_ALERT
+    undo_alert
```
emord: aren't these usually uppercase by vellum convention? Or is that just for test xml?

src/mugs.js
```diff
@@ -473,6 +473,15 @@ define([
             }
             return this.parentMug && this.parentMug.isInRepeat();
         },
+        /**
+         * Check if mug is referenced by other mugs
+         *
+         * @param except - Array of mugs to exclude when looking for references.
+         * @returns Boolean
+         */
+        isReferencedByOtherMugs: function (except) {
```
emord: really nitpicky: in a perfect world should mugs know what a form is like this? I usually think of it as forms encapsulate mug and mugs shouldn't necessarily know about others. I'm sure that convention is broken all over though or I may not be thinking of other use cases

src/form.js
millerdev: Agreed, although it's notoriously hard to break out of `_.each`. I'll see if I can redesign it to allow that.

src/undomanager.js
millerdev: I think the convention is lowercase except for test XML. Possibly uppercase for static strings, which is what this was before, although I think this was the first place I saw that.

src/mugs.js
millerdev: Good question. I think we've bent this line all over the place because `mug.form` exists and has become the defacto way to get at the form when you have a mug. More general encapsulation best practices say rather than doing `mug.form.something(arg)` it's better to do `mug.something(arg)` and leave the fact that `mug` has a form property as an implementation detail of `mug` rather than making the calling code aware of the mug's internals.

It could be argued that this information should be passed from the form to the undo manager rather than having the undo manager reach through the mug to get it. I personally think since we reference `mug.form` all over the place, it's cleaner to ask "Is this mug referenced by other mugs?" than "Does this mug's form know if this mug is referenced by other mugs?"

src/form.js
millerdev: Are you proposing that the caller would reach through the form to get to the logic manager? Like `form._logicManager.isReferencedByOtherMugs`? That breaks encapsulation and makes the caller have to know things like the logic manager mainly deals with ufids rather than mugs in many cases.

millerdev: After submitting that, I think maybe I didn't understand your suggestion. Help me understand if I missed your point.

src/form.js
millerdev: Oh yeah, this was a great idea.  	4eb3157

src/mugs.js
millerdev: Update: wording in previous comment

src/form.js
emord: I meant I think it would make more sense if the workload for this was in the logic manager. This only uses the ufid of the mugs, and I usually think of the logic manager as doing everything with references

emord: Just looked at your last commit. Looks like you pushed a bulk of this to the logic manager which is what I was thinking about :+1:

PR #858, opened 2017-02-23
don't let people change the form name

emord:
@orangejenny @millerdev This came from field dev/stack exchange post. We shouldn't have this box because it isn't reflected in HQ, but it would change it on mobile leading to very confusing behavior. Changing on the form page in HQ should be the only way to change this

@benrudolph @gcapalbo


millerdev:
The form name is such a disaster...

PR #860, opened 2017-03-01
Preserve <output /> without value or ref attribute

millerdev:
https://manage.dimagi.com/default.asp?248521 - not really a fix, but at least preserve `<output />` tags without the correct attributes and do not error while loading the form.

@emord @orangejenny


src/javaRosa/plugin.js
```diff
@@ -290,6 +290,7 @@ define([
                             } else if (ref) {
                                 return tempOutput.attr('ref', _toHashtag(ref))[0].outerHTML;
                             }
+                            return this;
```
millerdev: Do not drop outputs that do not have a `value` or `ref` attribute.

src/richText.js
```diff
@@ -467,7 +467,8 @@ define([
         var info = extractXPathInfo($(output)),
             xpath = form.normalizeHashtag(info.value),
             attrs = _.omit(info, 'value'),
-            startsWithRef = FORM_REF_REGEX.test(xpath) || form.hasValidHashtagPrefix(xpath),
+            startsWithRef = xpath && (FORM_REF_REGEX.test(xpath) ||
+                                      form.hasValidHashtagPrefix(xpath)),
```
millerdev: This change prevents the error on load label containing `<output />`.

tests/javaRosa.js
```diff
@@ -783,7 +785,6 @@ define([
     describe("The javaRosaplugin with one language", function() {
         before(function(done) {
             util.init({
-                features: {rich_text: false},
```
millerdev: This felt safe since it only affected one other test, and that test only needed a tiny change to its XML and otherwise was not affected adversely.

millerdev:
> Have you been able to determine how the form got into this state?

nope

PR #862, opened 2017-03-01
adding popovers with descriptions to databrowser

calellowitz:
@emord @millerdev @orangejenny cc: @czue 
this adds popovers to the data browser, the are correctly rendered after searches or opening of the tree. tested with 300+ case properties for load issues and there didnt seem to be a problem. When there is no description, there is no popover not an empty one
screen shot 2017-03-01 at 5 03 59 pm


czue:
wow this is nice!

PR #864, opened 2017-03-09
Move javaRosa logic updates into logic manager

millerdev:
Refactor in preparation for another change coming soon. Have wanted to do this forever to eliminate that nasty regex find/replace in `javaRosa.handleMugRename`.

@emord @orangejenny

### DIFF
--- a/src/logic.js
+++ b/src/logic.js
@@ -326,8 +326,8 @@ define([
             var _this = this,
                 returned = {};
             this.clearReferences(mug, property);
-            if (!value && mug.p[property] && _.isFunction(mug.p[property].forEachLogicExpression)) {
-                return mug.p[property].forEachLogicExpression(function (expr) {
+            if (!value && mug.p[property] && _.isFunction(mug.p[property].mapLogicExpressions)) {
+                return mug.p[property].mapLogicExpressions(function (expr) {
                     var messages = _this._addReferences(mug, property, expr);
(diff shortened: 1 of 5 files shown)


src/logic.js
```diff
@@ -390,17 +390,28 @@ define([
...
+
                 var pkey = mug.ufid + " " + property + " " + paths[0];
                 if (seen.hasOwnProperty(pkey)) {
                     return;
                 }
                 seen[pkey] = null;
-                var expr = new LogicExpression(mug.p[property], form.xpath),
-                    orig = expr.getText();
-                expr.updatePath(paths[0], paths[1]);
-                if (orig !== expr.getText()) {
```
millerdev: This used to always evaluate to false for itext properties because `LogicExpression` cannot parse an `ItextItem` so it just returned the original value.

tests/javaRosa.js
```diff
@@ -486,6 +486,7 @@ define([
             var label = util.addQuestion("Trigger", "label"),
                 text2 = util.addQuestion("Text", "text2");
             label.p.labelItext.set('<output value="/data/load-one" />');
+            label.validate("labelItext");
```
millerdev: This is now necessary to register the output value reference because `handleMugRename` no longer does a comprehensive scan of all itext values.

emord:
Excellent refactor

PR #865, opened 2017-03-09
move calculate condition to basic

emord:
@orangejenny this is into your other PR

calculate attrs only come up for hidden values so i think it might be nice to have this up at the top for those


orangejenny:
good idea

PR #866, opened 2017-03-09
Do not throw away XPath expressions on save before data sources loaded

millerdev:
This is the real fix for https://manage.dimagi.com/default.asp?248521 It also does something I've been thinking about for a while: hashtag transformations are persisted in the form so all hashtags can be converted to xpath using information in the form. Possible alternative: save entire data schema in the form rather than just hashtag transformations—I haven't considered all the edge cases to figure out if that's a better idea or not.

Best reviewed by commit 🐠 The bulk of the work is in the last two commits.

@orangejenny @emord


src/core.js
```diff
@@ -1277,6 +1277,7 @@ define([
         }
         this.data.core.form = form = parser.parseXForm(
             formXML, options, this, _this.data.core.parseWarnings);
+        this.onXFormLoaded(form);
```
millerdev: Had to add this new extension point so javaRosa could load all data references in `<output>` tags before the tree is populated.

src/form.js
```diff
@@ -1172,6 +1184,80 @@ define([
...
+         * necessary to guess hashtag meansings using default transformations.
+         */
+        return function inferHashtags(hashtag, xpath_) {
+            var newpath;
+            changes = {};
+            usedTransforms = {};
+            info.hashtagMap = _.clone(form.hashtagMap);
+            info.hashtagNamespaces = form.hashtagNamespaces;
+            try {
+                newpath = parser.parse(hashtag).toXPath();
```
millerdev: This is where the magic happens. The hashtag is parsed and converted to xpath and then (below) compared to the original xpath expression. If they match, use the hashtag information gathered during the parse to update the form's hashtags and transformations.

src/javaRosa/plugin.js
```diff
@@ -271,32 +271,54 @@ define([
...
+                                if (value) {
+                                    form.inferHashtagMeanings(hashval, value);
+                                }
+                                value = hashval;
+                            } else {
+                                // hashtagify xpath expression if possible
+                                // NOTE the success of this could depend
+                                // on whether data sources are loaded,
+                                // which means outputs may load differently
+                                // based on network conditions.
```
millerdev: This warning could probably be removed now that hashtag transformations are written in the form. I wrote the comment before I implemented that feature.

tests/databrowser.js
```diff
@@ -401,7 +403,7 @@ define([
             it("overwrites the forms preloaded tags", function() {
                 var form = call('getData').core.form;
                 assert(form.isValidHashtag(dobProp));
-                assert.strictEqual(form.hashtagMap[dobProp], null);
+                form.hashtagMap[dobProp] = null;
```
emord: is the map not null at this point? Should we be asserting it to something else

millerdev: No, since xpaths are now inferred on loading the form the path is not null at this point. I had to set it to null here so I could verify that `loadCaseData` actually updated `hasthagMap` (because it loads the same path as the one that is inferred on loading the form).

emord: Would it be overkill to change the test to say:

```
assert.strictEqual(form.hashtagMap[dobProp], 'thecasepath')
form.hashtagMap[dobProp] = null;
loadCaseData
assert.strictEqual(form.hashtagMap[dobProp], 'thecasepath')
```

millerdev: de52f3d1cabc4fb024999fb647f9f78bb91c3ff1

PR #868, opened 2017-03-14
Bump add choice left

orangejenny:
@biyeun any objections to this? Suggestion from @benrudolph to make the "Add Choice" look less like a choice itself.

screen shot 2017-03-14 at 9 41 07 am

### DIFF
--- a/src/less-style/tree-content.less
+++ b/src/less-style/tree-content.less
@@ -34,7 +34,7 @@
 
             .add_choice {
                 display: block;
-                margin-left: 2px;
+                margin-left: 50px;
                 cursor: pointer;
                 padding: 0 5px;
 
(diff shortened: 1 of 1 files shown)


benrudolph:
ah sorry i actually meant the other way so it's on the same hierarchy plane as the choice

benrudolph:
my main complaint was that it looked too much like a _question_

orangejenny:
@benrudolph Hrm, I think that looks weird, though it might be that I'm not used to it. Will think about it.
screen shot 2017-03-14 at 9 53 49 am

benrudolph:
👍 trust your judgement. would recommend trying to do some real world app edits

orangejenny:
It's growing on me...I do like that it does't line up quite so much with choice/questions, so it looks like something different, and there's more space on the left (think I tend to click on the left side of tree elements). Will merge.

biyeun:
I definitely like @benrudolph's suggestion here more. Having it in line with the choice icon makes a lot more sense

PR #870, opened 2017-03-14
Keyboard shortcut for full screen

orangejenny:
@millerdev / @emord 

Used `Expand Editor` / `Shrink Editor` because leaving it as `Enter Full Screen` plus the shortcut code broke onto two lines.

cc @dannyroberts

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -413,7 +413,7 @@ define([
     fn.toggleFullScreen = function () {
         var _this = this,
             $fullScreenMenuItem = $(_.find(_this.$f.find('.fd-tools-menu').nextAll(), function(li) {
-                return $(li).find("a").text().match(/editor/i);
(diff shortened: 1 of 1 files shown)


src/core.js
```diff
@@ -400,27 +410,33 @@ define([
         });
     };
 
+    fn.toggleFullScreen = function () {
+        var _this = this,
+            $fullScreenMenuItem = $(_.find(_this.$f.find('.fd-tools-menu').nextAll(), function(li) {
+                return $(li).find("a").text().match(/editor/i);
```
millerdev: This seems pretty fragile since it will pick up any other menu item that happens to contain the word "editor" Suggestion: change to `/(shrink|expand) editor/i`

orangejenny: sure => https://github.com/dimagi/Vellum/pull/870/commits/e390b0f85e28d1c4a1c9905f4d84aeb809305a45

PR #879, opened 2017-03-27
Expand parent groups when showing in tree

millerdev:
@emord @orangejenny cc @benrudolph

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -215,7 +215,10 @@ define([
 
         $(document).on('click', '.jstree-hover', function(e) {
             e.preventDefault();
-            _this.jstree("hover_node", $(this).data("ufid"));
+            var ufid = $(this).data("ufid");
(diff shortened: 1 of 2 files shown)


orangejenny:
Thank you for fixing! I had mentally added this to my backlog but hadn't actually grabbed the ticket.

Test would be nice if it's feasible; there's a test for the basic behavior at https://github.com/dimagi/Vellum/blob/master/tests/richText.js#L621

millerdev:
heh, I thought about adding a test and then decided it wasn't worth it because I'm not sure we have a good way to detect if branches of the tree are collapsed or expanded... but I'll take a stab at it since you asked

millerdev:
@orangejenny added test: 60dcd80

orangejenny:
lovely, looks good to me

PR #884, opened 2017-04-10
Allow choices to start off without label or id

orangejenny:
Deal with some issues Derek raised about the workflow of adding choices to a select question.

Previously, when you added a choice, the label and id would default to `choiceN`, which the user would inevitably need to overwrite. Now I'm letting the id and label default to blank, which is a better workflow, but because this causes a validation error, I'm letting the validation run, assuming it's going to error on `nodeID` and the label, and then removing those errors, which is a bit sketchy.

@emord 
cc @esoergel

### DIFF
--- a/tests/questionTypes.js
+++ b/tests/questionTypes.js
@@ -605,7 +605,7 @@ define([
 
         it("adds and removes the add choice action when type changes", function() {
             util.loadXML("");
-            util.addQuestion("Text", "question1");
+            var mug = util.addQuestion("Text", "question1");
             var changerSelector = ".fd-question-changer";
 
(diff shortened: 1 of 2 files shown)


tests/questionTypes.js
```diff
@@ -605,7 +605,7 @@ define([
 
         it("adds and removes the add choice action when type changes", function() {
             util.loadXML("");
-            util.addQuestion("Text", "question1");
+            var mug = util.addQuestion("Text", "question1");
```
emord: can you add a test and/or assertion that asserts the choice's label is blank?

orangejenny: updated

PR #886, opened 2017-04-26
allow custom load options

biyeun:
@orangejenny cc @dannyroberts


orangejenny:
Looks good.

I hesitated for a moment wondering if the spinner shown during save should match the loading spinner, thought about it and agree that it's better to use the standard vellum spinner during save, since it lets you still see the form.

PR #889, opened 2017-05-01
Update add_choice tests

orangejenny:
https://github.com/dimagi/Vellum/pull/887#pullrequestreview-35520379

I haven't actually been able to reproduce the bug in a test, even when adding choices repeatedly. Did notice that tests don't need to call `ensureCurrentMugIsSaved` anymore, though.

@millerdev

### DIFF
--- a/tests/questionTypes.js
+++ b/tests/questionTypes.js
@@ -603,7 +603,7 @@ define([
             assert.ok($(changerSelector + " .change-question:not([data-qtype*='Select'])").length > 0);
         });
 
-        it("adds and removes the add choice action when type changes", function() {
+        it("jls adds and removes the add choice action when type changes", function() {
             util.loadXML("");
             util.addQuestion("Text", "question1");
             var changerSelector = ".fd-question-changer";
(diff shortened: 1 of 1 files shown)


tests/questionTypes.js
```diff
@@ -616,41 +616,47 @@ define([
...
-                "question2"
-            );
-
-            util.deleteQuestion("question1/choice1");
-
-            $(changerSelector + " > a").click();
-            $options = $(changerSelector + " .change-question");
-            $options.filter("[data-qtype='Text']").click();
-            assert.strictEqual($(".add_choice").length, 0);
+            _.defer(function() {
```
millerdev: Since this is a synchronous test, it looks like the assertions inside this `defer` block are likely called outside the scope of the test, which means that failure will either be silently ignored (hopefully not), or mocha may holler about failing assertions outside a test. The easy way to test this is to add a failing assertion inside the `defer` block and then run the test to see what happens. The solution is likely to add a `done` argument to the test function (line 606), making it an async test, and then to call `done();` as the last statement in this `defer`'d function.

orangejenny:
@millerdev I took another look at this and realized the changes I made in the first commit didn't actually address the bug fixed in https://github.com/dimagi/Vellum/pull/887, which made sure that when adding a choice, if the previously-selected question didn't have an id, it would get one, *not* that the new choice would immediately have an id. Updated PR to add multiple choices in one test and verify that all but the last one get ids, and verified that this test does fail without the changes made in https://github.com/dimagi/Vellum/pull/887.

PR #890, opened 2017-05-02
Validate broken references in itext

millerdev:
https://manage.dimagi.com/default.asp?252531 - I found a bug in the previous fix while re-testing on a real app.

@emord

### DIFF
--- a/src/logic.js
+++ b/src/logic.js
@@ -455,7 +455,7 @@ define([
                     .value();
             _.each(invalid, function(ref) {
                 var mug = _this.form.getMugByUFID(ref.mug);
-                _this.updateReferences(mug, ref.property, mug.p[ref.property]);
+                _this.updateReferences(mug, ref.property);
             });
         },
         /**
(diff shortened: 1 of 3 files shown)


src/logic.js
```diff
@@ -455,7 +455,7 @@ define([
                     .value();
             _.each(invalid, function(ref) {
                 var mug = _this.form.getMugByUFID(ref.mug);
-                _this.updateReferences(mug, ref.property, mug.p[ref.property]);
+                _this.updateReferences(mug, ref.property);
```
emord: how does removing this property add more validation?

millerdev: Line 329 starts with `if (!value && mug.p[property]`. If a value is passed in here then that `if` expression evaluates to false, and questions that only reference case properties through itext are not validated. I've tried to remove the `!value &&` part of that expression before, but it caused other issues. It may make sense to remove the `value` parameter of `addReferences` as well as other functions that call it, but I didn't dig that deeply this time because I think removing it will be more complicated than that.

PR #894, opened 2017-05-08
Use index XPath on drag/drop related index node

millerdev:
@emord @orangejenny

### DIFF
--- a/tests/databrowser.js
+++ b/tests/databrowser.js
@@ -583,6 +583,23 @@ define([
                 });
             });
 
+            it("should drag/drop case index", function (done) {
+                util.loadXML("");
+                var mug = util.addQuestion("DataBindOnly", "mug"),
+                    input = $("[name=property-calculateAttr]"),
+                    editor = input.ckeditor().editor,
+                    widget = util.getWidget('property-calculateAttr');
+                widget.input.promise.then(function () {
+                    editor.on('change', function() {
+                        assert.equal(mug.p.calculateAttr,
+                            "instance('commcaresession')/session/data/case_id");
+                        done();
+                    });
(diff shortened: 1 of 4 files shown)


tests/databrowser.js
```diff
@@ -583,6 +583,23 @@ define([
...
+            it("should drag/drop case index", function (done) {
+                util.loadXML("");
+                var mug = util.addQuestion("DataBindOnly", "mug"),
+                    input = $("[name=property-calculateAttr]"),
+                    editor = input.ckeditor().editor,
+                    widget = util.getWidget('property-calculateAttr');
+                widget.input.promise.then(function () {
+                    editor.on('change', function() {
+                        assert.equal(mug.p.calculateAttr,
+                            "instance('commcaresession')/session/data/case_id");
```
emord: Does this mean that you can't reference the case id by using `#case`?

tests/datasources.js
```diff
@@ -198,6 +199,20 @@ define([
                 var house = nodes.child.nodes.mother.nodes.household.nodes;
                 assert.equal(house.address.hashtag, "#case/grandparent/address");
             });
+
+            it("should use index path for related parent node", function() {
+                assert.strictEqual(nodes.child.nodes.mother.hashtag, null);
+                assert.equal(nodes.child.nodes.mother.xpath,
```
emord: Similar for this? `#case/parent` will always refer to a property of the case right? there's no way to nicely reference index ids

tests/databrowser.js
millerdev: That's correct. I did it this way because `#case/parent` and `#case/grandparent` are ambiguous if we allow that.

tests/datasources.js
millerdev: Correct.

PR #895, opened 2017-05-12
Copy/paste hashtags in rich text fields

millerdev:
cut copy paste logic expressions
*You have to imagine where I'm pressing CTRL+C and CTRL+V

This uncovered more edge cases than I had hoped. CKEditor has a lot of rough edges that make the editing experience annoying, and our integration is getting more tightly coupled with its internals as we try to address issues like this.

Tested in Chrome and Firefox.

@emord @orangejenny

### DIFF
--- a/src/richText.js
+++ b/src/richText.js
@@ -234,8 +234,8 @@ define([
                     editor.on('instanceReady', editor.focus);
                 }
             },
-            select: function (index) {
-                ckSelect.call(null, editor, index);
+            select: function (index, length) {
+                ckSelect.call(null, editor, index, length);
                 return wrapper;
             },
             on: function () {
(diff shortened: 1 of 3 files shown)


src/richText.js
```diff
@@ -277,11 +277,40 @@ define([
...
-        }, null, null, 2 );
+            var data = event.data;
+            if (data.dataTransfer && data.dataTransfer.getData("Text")) {
+                // Get plain text instead of HTML because HTML encoded
+                // content from applications like Word or your text
+                // editor often contains unwanted styling information.
+                //
+                // Insert HTML-ified plain text rather than HTML with
+                // bubbles because we cannot reliably find hashtags in
+                // text fragments that are not valid XPath expressions.
```
emord: It's been a while since I've looked at this code, but doesn't `toRichText` attempt to do something like this smartly? I think the right workflow might be something like

- attempt to process string as an xpath with hashtags (with or without the escaping `)
- if that fails, look for escaped hashtags (ones with `)
- if that fails assume it's a string and shouldn't be parsed

millerdev: It's more complicated than that. We have to evaluate the pasted value in the context of the expression it's being pasted into, and I don't know of a way to do that while preserving the insertion point after the pasted text. It seemed like it would take a deeper dive into CKEditor paste logic to accomplish that. Also, I found it surprisingly difficult to get bubble-ified HTML to paste into the editor properly. CKEditor's content filter was stripping parts of the bubbles out, even after I updated `CKEDITOR.config.allowedContent` to allow the relevant tags and attributes.

emord: Oh i see what you mean now. Thanks for the explanation.

I also had issues with allowedContent not behaving how I thought it should.

millerdev:
This should get deployed later today.

PR #900, opened 2017-06-05
Fix errors related to chai upgrade

millerdev:
`assert.sameMembers(["#form/text2","#form/text1","#form/text2"], ["#form/text1","#form/text2"])`
-  Before (chai 3.5.0) -> pass
- After (chai 4.0.1)  -> fail
- Replace with `assert.includeMembers`

`assert.include([{a: 1}], {a: 1})`
- Before (chai 3.5.0) -> pass
- After (chai 4.0.1)  -> fail
- Replace with `assert.deepInclude` (new in Chai 4)

@orangejenny @emord

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -46,7 +46,7 @@
     },
     "devDependencies": {
         "equivalent-xml-js": "mwhite/equivalent-xml-js#master",
-        "chai": ">4.0.0",
+        "chai": "^4.0.0",
         "xrayquire": "requirejs/xrayquire"
     }
 }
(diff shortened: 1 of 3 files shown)


bower.json
```diff
@@ -46,7 +46,7 @@
     },
     "devDependencies": {
         "equivalent-xml-js": "mwhite/equivalent-xml-js#master",
-        "chai": ">=1.5.0",
+        "chai": ">4.0.0",
```
emord: Perhaps this could be changed to ^ or ~ to allow for safer upgrades

millerdev: d6b98f6

orangejenny:
@millerdev thank you for taking this

PR #901, opened 2017-06-06
Reload fuse data on enable easy references

millerdev:
Earlier today while editing a form loaded with easy references disabled I noticed that I could not auto-complete case references after turning on easy references.

@emord @orangejenny

### DIFF
--- a/tests/main.js
+++ b/tests/main.js
@@ -172,6 +172,12 @@ requirejs(['jquery', 'jquery.vellum'], function ($) {
         if (mocha.env) {
             // ensure the first instance is fully loaded before running tests
             load("", function () { mocha.run(); });
+        } else if (/[?&]load=saved(&|#|$)/.test(window.location.href)) {
+            // Use Chrome dev tools to preset form XML
+            // (Application > Storage > Session Storage > http://localhost...
+            //  > vellum.tests.main.lastSavedForm value)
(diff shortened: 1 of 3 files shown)


tests/main.js
```diff
@@ -172,6 +172,12 @@ requirejs(['jquery', 'jquery.vellum'], function ($) {
         if (mocha.env) {
             // ensure the first instance is fully loaded before running tests
             load("", function () { mocha.run(); });
+        } else if (/[?&]load=saved(&|#|$)/.test(window.location.href)) {
+            // Use Chrome dev tools to preset form XML
+            // (Application > Storage > Session Storage > http://localhost...
+            //  > vellum.tests.main.lastSavedForm value)
+            // and then add ?load=saved to query string and reload.
+            load(session.getItem("vellum.tests.main.lastSavedForm") || "");
```
millerdev: Added this convenience because I'm often wanting a way to automate two things:
- Automatically load saved form on reload
- Debug initial load events/sequence with XML that may have been hand-edited or was otherwise broken in a way that cannot be easily simulated with Edit Source XML > Paste > Update Source. Hence the comment, which is there as a reminder for the next time I forget how to do that.

millerdev:
Looks like tests are going to fail. Fixing...

PR #902, opened 2017-06-07
Improve handling of choice with blank value

millerdev:
Trying to make this situation better: https://manage.dimagi.com/default.asp?255685

Review by commit 🐠 

729799c is not really part of the fix, but solves a long-standing annoyance I had with the view source XML feature.

@orangejenny @emord

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -115,6 +115,11 @@ define([
         }
     };
 
+    function validateMug(mug) {
+        mug.validate();
+        return !mug.getErrors().length;
+    }
+
     var fn = {};
 
     fn.init = function () {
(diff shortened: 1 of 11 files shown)


src/core.js
```diff
@@ -141,26 +146,34 @@ define([
...
+            var form = _this.data.core.form;
+            if (form.hasBrokenReferences()) {
+                return {
+                    title: "Errors in Form",
+                    content: "<div class='alert alert-danger'>Form has " +
+                        "reference errors.<br>Look for questions marked with " +
+                        "<i class='fd-valid-alert-icon fa fa-warning'></i> " +
+                        "and check they don't reference deleted questions.</div>",
+                };
+            } else if (!form.isFormValid(validateMug)) {
```
millerdev: Was thinking about this in bed last night, and I remembered why I didn't do this when I originally implemented the save button warning. This traverses all mugs in the form, so for very large forms could introduce a pause on mouse over the save button. I have some ideas of how to improve it:

- Validate current mug since that is likely the one that needs it most (easy)
- Check if there are other validation errors in the form in some way that is not O(n). I can't remember if that's possible right now or if we need to implement it? (maybe harder)

orangejenny: Could we allow the save to go through and then validate, without freezing the form?

emord: I think the issue is that this happens on mouseover. I think if that's the case then we can't merge this because it will make most enterprise form building unusable.

millerdev: Updated to only check the current mug. While not optimal yet, this solves the current issue of creating a new choice item and then immediately clicking save, which used to display no error or warning that anything is wrong, but now shows the save button validation warning.

src/core.js
```diff
@@ -174,7 +193,6 @@ define([
         });
 
         bindBeforeUnload(this.data.core.saveButton.beforeunload);
-        this.data.core.currentErrors = [];
```
millerdev: Dead code? not referenced anywhere.

PR #905, opened 2017-07-20
i18n

millerdev:
I opted to use global `gettext` functions since it seemed like the simplest way to get this effort off the ground. It shouldn't be too hard to switch to a different i18n framework later if we decide that's better for whatever reason. However, this is not very nice for anyone who wants to use translated vellum outside of HQ. I don't know of anyone doing that, so not worrying about it for now. I suppose it would be possible to extract the translation from the HQ source if we wanted to package translations with vellum in the future.

I ran `vellum-to-hq` and then `./manage.py makemessages -d djangojs --all` and verified that it's picking up text from compiled vellum source files. One thing I am assuming is that the global `gettext` functions are installed before vellum is loaded in `form_designer.html` in HQ. If not, we'll have to do some finagling there.

Add `?typoglycemia` to the vellum stand-alone URL to verify that `gettext()` is getting called as expected inside vellum.

Best reviewed by commit 🐠 

@orangejenny @emord

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -15,5 +15,9 @@
     "nonew": true,
     "undef": true,
     "unused": "vars",
-    "globals": {"define": false}
+    "globals": {
+        "define": false,
+        "gettext": false,
+        "ngettext": false
+    }
 }
--- a/buildmain.py
+++ b/buildmain.py
@@ -62,6 +62,13 @@ requirejs.config({
 define('css/css', {});
 define('less/less', {});
 
+if (!window.gettext) {
+    window.gettext = function (arg) { return arg; };
+    window.ngettext = function (singular, plural, count) {
+        return count === 1 ? singular : plural;
+    };
+}
+
 define([
     'jquery',
     %(main_components)s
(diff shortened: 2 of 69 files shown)


millerdev:
@emord I don't know for sure about performance, but I doubt it's going to be much impact. [Django's `gettext`](https://github.com/django/django/blob/master/django/views/i18n.py#L104-L111) is a pretty simple function. I'd be very interested if you have any evidence that this is slowing things down much.

millerdev:
I tested this on staging, and verified that `gettext` is present in the global context when vellum is loaded. It is, and so I'm going to say this is good to go. I think I'll wait until Monday to deploy.

emord:
@millerdev I don't have any evidence, I was just curious and didn't think it was blocking. 

It's been a while since I looked into vellum performance, but I recall that there is a lot of template rendering and we already had issues reported for performance in large forms. and it seemed like this would make template_rendering a small amount larger. would be nice if we had more insight into vellum's performance issues i think

millerdev:
This will be deployed to HQ today, which means we can ship new translation files to transifex and have them translate the form builder for us. @esoergel, are you the person who takes care of that?

esoergel:
@millerdev That's typically done during the Tuesday deploys by running `./scripts/update-translations.sh` (there's a prompt during the deploy script about it), but that can be run whenever.  I'd try running it now so you can verify that vellum strings appear on [transifex](https://www.transifex.com/dimagi/commcare-hq/translate/#es) in the djangojs.po file.

PR #912, opened 2017-07-28
Fix markdown toggle switch disappearing

millerdev:
https://manage.dimagi.com/default.asp?258455

@orangejenny @emord


orangejenny:
Do we have any tests for the ui aspect of markdown? It's such a finicky ui, would be nice to be tested.

millerdev:
Heh, I don't know. I usually don't spend a lot of time writing tests for UI because it's often so volatile and prone to be changed that automated tests often seem like a waste of time. However, it might be worth it if we're getting a lot of bugs reported about a particular/problematic UI component.

PR #915, opened 2017-08-18
Vellum Commander

millerdev:
screenshot 2017-08-18 17 02 41

@orangejenny @emord

### DIFF
--- a/src/commander.js
+++ b/src/commander.js
@@ -16,7 +16,8 @@ define([
 ) {
     var fn = {},
         handlers = {},
-        isMac = /Mac/.test(navigator.platform);
+        isMac = /Mac/.test(navigator.platform),
+        INSERT_AT = /^(.*) +(before|after|into|in|first|last)(?: +((?:#form|\/data)\/[^ ]+))? *$/;
 
     $.vellum.plugin('commander', {}, {
         init: function () {
@@ -24,7 +25,7 @@ define([
                 cmd = vellum.data.commander;
             cmd.vellum = vellum;
             cmd.container = $(commanderTemplate());
-            cmd.addQuestion = vellum.$f.find(".fd-add-question");
+            cmd.addQuestionButton = vellum.$f.find(".fd-add-question");
             cmd.input = cmd.container.find("input");
             cmd.input.on("keydown", function (e) {
                 var chord = getKeyChord(e);
@@ -59,13 +60,13 @@ define([
             setupAutocomplete(cmd.input, names);
             cmd.autocompleted = true;
         }
-        cmd.addQuestion.hide();
+        cmd.addQuestionButton.hide();
         cmd.container.show();
         cmd.input.focus().select();
     }
 
     function hideCommander(cmd) {
-        cmd.addQuestion.show();
+        cmd.addQuestionButton.show();
         cmd.container.hide();
         cmd.input.val("").removeClass("alert-danger");
     }
@@ -114,16 +115,34 @@ define([
     handlers.Escape = hideCommander;
 
     fn.doCommand = function (text, vellum) {
-        var types = fn.getQuestionMap(vellum);
         text = text.toLowerCase().trim();
+        var types = fn.getQuestionMap(vellum),
+            insertAt = INSERT_AT.exec(text),
+            position, refMug;
+        if (insertAt) {
+            text = insertAt[1];
+        }
         if (types.hasOwnProperty(text)) {
+            if (insertAt) {
+                position = insertAt[2];
+                if (position === "in") {
+                    position = "into";
+                }
+                if (insertAt[3]) {
+                    refMug = vellum.getMugByPath(insertAt[3]);
+                    if (!refMug) {
+                        return;
+                    }
+                } else {
+                    refMug = vellum.getCurrentlySelectedMug();
+                }
+            }
             try {
-                vellum.addQuestion(types[text].__className);
-                return true;
+                return vellum.addQuestion(types[text].__className, position, refMug);
             } catch (err) {
+                //window.console.log(err.message);
             }
         }
-        return false;
     };
 
     fn.getQuestionMap = function (vellum) {
--- a/src/commander.js
+++ b/src/commander.js
@@ -218,7 +218,8 @@ define([
             } catch (err) {
                 return;  // fail on bad argument
             }
-            return obj.config.run(args);
+            var value = obj.config.run(args);
+            return value && {name: obj.config.name, value: value};
         }
 
         var callbacks = $.fn.atwho["default"].callbacks,
@@ -249,6 +250,7 @@ define([
                 // first argument.
                 {
                     // add question: Type name [position [#question]]
+                    name: "add question",
                     args: [
                         literals(typeItems, getMugClassName),
                         literals(positions, getPosition),
@@ -285,6 +287,7 @@ define([
                 //},
                 {
                     // select question
+                    name: "select question",
                     args: [questionRef],
                     run: function (args) {
                         var mug = args[0];
@@ -366,13 +369,13 @@ define([
             _.defer(onCommand, cmd);
         } else {
             var text = cmd.input.val(),
-                ok = fn.doCommand(text, cmd.vellum);
-            if (ok) {
+                result = fn.doCommand(text, cmd.vellum);
+            if (result) {
                 hideCommander(cmd);
-                analytics.workflow("Commander success");
+                analytics.usage("Commander", result.name);
             } else {
                 cmd.input.addClass("alert-danger");
-                analytics.workflow("Commander fail");
+                analytics.usage("Commander", "bad command");
             }
         }
     }
--- a/src/commander.js
+++ b/src/commander.js
@@ -31,7 +31,18 @@ define([
                 if (handlers.hasOwnProperty(chord)) {
                     handlers[chord](cmd);
                     e.preventDefault();
+                } else {
+                    cmd.input.removeClass("alert-danger");
                 }
+            });
+            cmd.container.find(".fd-add-question-toggle").click(function (e) {
+                hideCommander(cmd);
+                e.preventDefault();
+                setTimeout(function () {
+                    // open add question menu after delay to allow atwo menu
+                    // to hide. https://stackoverflow.com/a/29572644/10840
+                    $(".fd-add-question-dropdown").addClass('open');
+                }, 0);
             });
             $(".fd-add-question-dropdown").append(cmd.container.hide());
             $(document).on("keydown", function (e) {
@@ -50,13 +61,13 @@ define([
         }
         cmd.addQuestion.hide();
         cmd.container.show();
-        cmd.input.focus().select().removeClass("error");
+        cmd.input.focus().select();
     }
 
     function hideCommander(cmd) {
         cmd.addQuestion.show();
         cmd.container.hide();
-        cmd.input.val("");
+        cmd.input.val("").removeClass("alert-danger");
     }
 
     function setupAutocomplete(input, choices) {
@@ -83,7 +94,7 @@ define([
             if (ok) {
                 hideCommander(cmd);
             } else {
-                cmd.input.addClass("error");
+                cmd.input.addClass("alert-danger");
             }
         }
     }
@@ -110,7 +121,6 @@ define([
                 vellum.addQuestion(types[text].__className);
                 return true;
             } catch (err) {
-                // TODO display error in UI
             }
         }
         return false;
--- a/src/commander.js
+++ b/src/commander.js
@@ -34,14 +34,8 @@ define([
                     cmd.input.removeClass("alert-danger");
                 }
             });
-            cmd.container.find(".fd-add-question-toggle").click(function (e) {
+            cmd.container.find(".fd-commander-close").click(function (e) {
                 hideCommander(cmd);
-                e.preventDefault();
-                _.defer(function () {
-                    // open add question menu after delay to allow atwo menu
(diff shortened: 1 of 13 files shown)


orangejenny:
snazzy

src/less-style/structure.less
```diff
@@ -37,6 +37,23 @@ body.vellum-full-screen {
...
+
+                    .fd-add-question-toggle {
+                        color: white;
+                        height: 17px;
+                    }
+
+                    input {
+                        border: 1px solid @cc-dark-cool-accent-low;
+                        flex: 2;
+                        margin: -2px 0;
```
millerdev: Would appreciate help with these styles. `flex` is only supported on modern browsers, which I think is fine, but happy to change if not recommended. Also the negative margin may not work well for other OS's text inputs (I only tested on Mac).

src/commander.js
```diff
@@ -0,0 +1,348 @@
...
+                    sorter: function(query, items) { return items; },
+                },
+            },
+        };
+    }
+
+    function onCommand(cmd) {
+        if (cmd.input.atwho('isSelecting')) {
+            // HACK delay until value has been inserted
+            setTimeout(function () { onCommand(cmd); }, 0);
```
orangejenny: Minor, but `_.defer` does the same thing and is slightly more readable.

src/less-style/structure.less
orangejenny: It looks like you're using flex and the negative margin so the input takes up all width except the plus icon, and so it lines up nicely with the plus, is that right? And the plus is there so that you can click on it to get out of the commander?

Another option could be to hide the plus icon when the commander is shown, let the input take up the full width, and have a different way to exit the commander. Maybe just an x, something like below. It would mean two clicks to add a question the traditional way (click the x, then click the new-visible add question button) but I'm not certain that clicking the plus is an intuitive way to close the commander in the first place.

screen shot 2017-08-22 at 10 25 28 am

screen shot 2017-08-22 at 10 28 48 am

orangejenny:
Thoughts on interaction: when atwho is open, could enter select just the option, so you'd have to press enter again to execute the command? In the rest of vellum you can use enter to select without losing focus (realize this is a different use case, but I think it's going to be jarring for people who will press enter to select a question type and expect to keep typing the rest of the command).

src/commander.js
```diff
@@ -0,0 +1,348 @@
...
+        } else {
+            var text = cmd.input.val(),
+                ok = fn.doCommand(text, cmd.vellum);
+            if (ok) {
+                hideCommander(cmd);
+            } else {
+                cmd.input.addClass("alert-danger");
```
orangejenny: Thoughts on displaying the error message, if there is one?

src/less-style/structure.less
orangejenny: After reading a little more, realizing that you can use this to do things besides add questions also makes me want the commander to not look like it's part of the add question button. I do think it makes sense to be part of the question tree, though. Did you experiment at all with showing the commander above/below add question? I could imagine it being its own bar like the search bar in the add properties panel. Vertical space in the question tree is precious, but this is something that's easy to show/hide so I don't think that's a big deal.

src/commander.js
```diff
@@ -81,6 +82,11 @@ define([
             suffix: " ",
             tabSelectsMatch: true,
         });
+        atwho.autocomplete(
+            input,
+            {form: cmd.vellum.data.core.form, on: _.identity},  // fake mug
+            {useHashtags: true, tabSelectsMatch: true}          // options
```
orangejenny: Similar to a comment I made on the overall PR, I think this atwho and the atwho in question properties should behave the same as much as possible, so I don't think this one should use `tabSelectsMatch`.

src/commander.js
```diff
@@ -72,30 +68,96 @@ define([
...
-                // match first argument only
-                matcher: function (flag, subtext) { return subtext; }
-            }
-        });
+    function setupAutocomplete(cmd) {
+        if (cmd.atwhoConfig) {
+            return;
+        }
+        var typeNames = _.pluck(fn.getQuestionMap(cmd.vellum), "typeName"),
+            positions = ["after", "before", "in", "first in"],
```
orangejenny: I love that there's an autocomplete for these; it makes the commander much easier to use.

src/commander.js
```diff
@@ -72,30 +68,96 @@ define([
...
             event.preventDefault();
         });
     }
 
+    function literals(names) {
+        var source = "(" + _.map(names, RegExp.escape).join("|") + ")";
+        return {regexp: source, names: names};
+    }
+
+    function getAtwhoConfig(commandForms) {
```
orangejenny: A summarizing comment here would be helpful, there's a lot going on here (programmatically generated regexes!).

src/commander.js
```diff
@@ -0,0 +1,348 @@
...
+
+        function literals(names, getArg) {
+            if (!getArg) {
+                throw new Error("getArg parameter is required");
+            }
+            var source = _.map(names, RegExp.escape).join("|");
+            return {regexp: source, names: names, getArg: getArg};
+        }
+
+        function xname(predicate) {
```
emord: I'm not exactly sure what this or itemizer do or why they do it. Could you add some comments?

Also xname is only used once. Are you planning on using it more later? I think it could be simplified by referencing itemizer directly. here

src/core.js
```diff
@@ -1457,37 +1472,54 @@ define([
...
+            } else {
+                var foo = _this.getInsertTargetAndPosition(
+                    _this.getCurrentlySelectedMug(), qType);
+                if (!foo) {
+                    throw new Error("cannot add " + qType + " at the current position");
                 }
+                position = foo.position;
+                refMug = foo.mug;
             }
+            analytics.workflow("Added question in form builder");
```
emord: @dimagi/product  should there be analytics differentiating between adding questions the normal way. and adding through commander?

tests/commander.js
```diff
@@ -0,0 +1,180 @@
...
+                assert.equal(mug.hashtagPath, path);
+            });
+        });
+
+        // TODO decide if : is a good prefix for commands other than add or select
+        // TODO fix selection after delete
+        //  (two nodes selected in tree after delete non-adjacent to current)
+        // TODO delete multiple
+        //_.each([
+        //    ":delete",
```
emord: I like where this is headed.  looks like you've got some big future plans for this

millerdev:
> What would you think of doing UAT on this?

Since this is mostly a hidden feature, I was imagining that we'd roll this out and then have a few select people try it out before making a release announcement. Do you think a UAT run would be better than that?

orangejenny:
@millerdev That sounds fine.

src/less-style/structure.less
millerdev: I like the idea of having an X button at the right end of the commander to dismiss it. I had originally wanted to keep the plus (+) icon to the left for that very purpose, especially for non-power users who accidentally get into the command bar and then don't know how to add a question. An X button is more intuitive for dismissing the commander. Great idea!

Don't really like the idea of showing the commander below the add question button, since that will either result in a content shift (jarring in its own way) or obscure part of the tree (quite possibly the part that the user wants to interact with), both of which seem undesirable. When you say "above add question" do you mean it would overlay the form name and menu button? That might be a reasonable place for it once you can invoke all of the menu items with the commander... (not yet)

I made it overlay/replace the add question button because I think that button will probably not be used very often by people who use the commander, and certainly not _while_ using the commander. Does it seem like a reasonable compromise if there is an X on the right to dismiss it and show the add question button?

millerdev:
> Thoughts on interaction: when atwho is open, could enter select just the option, so you'd have to press enter again to execute the command?

I had considered this, but that makes the interaction model more complex and less efficient: if there is an autocomplete popup present then you have to press enter twice (or three times if entering the question type since it automatically pops up the list of insert locations as soon as the question type is completed) to execute the command. The Enter key always executes the command with the currently selected item, no matter where you are in the commander. Use the tab key to complete and move on to the next field. I think this is pretty standard for command line interfaces.

> In the rest of vellum you can use enter to select without losing focus (realize this is a different use case...

Yes, different use case. Here it's about efficiency (the whole reason for the command bar). In other input fields the enter key has no special function like executing a command. I would consider this if a lot of people complain about the inconsistency after using the command bar for more than a day or two.

src/commander.js
millerdev: I'd like to do that in the future if I can think of a good place to show it. I put it off for now since it's a non-trivial feature involving more UI elements and I didn't want to get too distracted by that right now.

src/commander.js
millerdev: I disagree. See [other comment](https://github.com/dimagi/Vellum/pull/915#issuecomment-324151704).

src/commander.js
orangejenny: sounds good

orangejenny:
I agree they're conceptually different use cases, but I suspect muscle memory is going to win out - that your hand gets very used to typing `#question1 Enter` and it'll be difficult to train yourself to sometimes use `#question1 Tab` instead. But neither of us is a real user, so yeah, keeping it as-is and seeing if we get feedback on it sounds good.

tests/commander.js
millerdev: Indeed! I think we could eventually make most mouse interactions accessible through the commander. The other low hanging fruit is menu items, but don't want to get too carried away before I release something useful.

One thing I want to think about before doing an initial release (forgot to mention this when I made the PR) is whether we want to put commands like "Delete" and other menu items in the same namespace (autocomplete menu) as question types, or whether we should use a prefix character (like ":", but I think that's kinda ugly) or whether we should add a command specifically for adding questions? In the latter case, the add question command displayed in the screenshot at the top of this PR would change to "add Text after #group". That's nice for separating the command namespaces, but I'm not terribly fond of that idea because it means more typing to add a question, which is what I think people will use the commander for most of the time. Thoughts welcome.

src/less-style/structure.less
orangejenny: By "above add question" I meant something like the below image. But I'm just thinking about possibilities, especially for the future. Replacing the add question with the commander and adding an X to dismiss it does sound like a reasonable compromise.

screen shot 2017-08-22 at 5 33 45 pm

tests/commander.js
orangejenny: Haven't thought about this deeply, but one alternative would be to have different hotkeys for commands, instead of having a hotkey to open the commander and then you type a command. That would work with the current commands (add and jump), which both fit the pattern of opening the commander, typing a command, and then the commander closes and you need to re-open it. It wouldn't work so well if the commander becomes something more like a command-line interface that's used for multiple commands in a row.

Slack's command-K switcher is an elegant little in-between widget - it's definitely not a command-line interface, but it lets you do something more complex than a single hotkey would.

src/less-style/structure.less
millerdev: ca929f0903abc7274141d440ca0c4dc9aada7d1d

screenshot 2017-08-23 16 29 25

src/commander.js
millerdev: https://github.com/dimagi/Vellum/pull/915/commits/18f5dadd1bc6f7fa45c5411f1651c8d000369ee3

src/commander.js
millerdev: Added comments in https://github.com/dimagi/Vellum/pull/915/commits/ebbb027c946213edfa91aa23b31b2031826be41e and then removed these little functions in https://github.com/dimagi/Vellum/pull/915/commits/51698eb0712604744632591b3c9bba1eaacb8c47

src/core.js
millerdev: Added analytics for command bar in https://github.com/dimagi/Vellum/pull/915/commits/defb1daab749ef8fd919bd481449e7f09ca40833

tests/commander.js
millerdev: I'm going to release as is for now.

src/commander.js
```diff
@@ -362,8 +364,10 @@ define([
                 ok = fn.doCommand(text, cmd.vellum);
             if (ok) {
                 hideCommander(cmd);
+                analytics.workflow("Commander success");
```
orangejenny: Any chance it's trivial to distinguish between commands here?

src/less-style/inputs.less
```diff
@@ -76,3 +76,22 @@
         width: 100%;
     }
 }
+
+.fd-commander {
+    display: flex;
```
orangejenny: Is this still definitely necessary? Not blocking, but it'd be nice not to be using flex unless we confirm that our supported browsers all support flex (which I don't know off the top of my head).

src/core.js
krsdimagi: @emord @millerdev probably would be good to get some analytics here - although, if we're doing a soft rollout and know who is using it, probably isn't super high priority and shouldn't block the initial release.

src/less-style/inputs.less
millerdev: When I remove all the flex styles the it looks terrible. Feel free to suggest alternate styles if you know of a better way.

src/commander.js
millerdev: https://github.com/dimagi/Vellum/pull/915/commits/c4b585419c7c456d30f3388d68aaee9dd77424ba

src/less-style/inputs.less
orangejenny: I'll take a stab at it in another PR.

src/core.js
millerdev: @krsdimagi I added analytics on command execution. It records the type of command (currently "add question" or "select question") that was executed. If the command could not be executed it records "bad command".

PR #918, opened 2017-08-24
Tweak commander styles

orangejenny:
screen shot 2017-08-24 at 3 03 26 pm

@millerdev

### DIFF
--- a/src/less-style/inputs.less
+++ b/src/less-style/inputs.less
@@ -78,20 +78,22 @@
 }
 
 .fd-commander {
-    display: flex;
     background-color: @cc-bg;
-    padding: 4px 1px 3px 3px;
-    margin-right: 1px;
+    padding: 5px;
 
     input {
         border: 1px solid @cc-neutral-hi;
-        flex: 1;
+        height: 19px;
+        width: calc(~"100% - 15px");
     }
 
     .fd-commander-close {
         background: none;
         border: none;
         color: @cc-neutral-low;
-        flex: 0 0 20px;
+        position: absolute;
+        top: 8px;
+        right: 8px;
+        cursor: pointer;
     }
 }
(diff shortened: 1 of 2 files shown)


src/less-style/inputs.less
```diff
@@ -78,20 +78,22 @@
...
     background-color: @cc-bg;
-    padding: 4px 1px 3px 3px;
-    margin-right: 1px;
+    padding: 5px;
 
     input {
         border: 1px solid @cc-neutral-hi;
-        flex: 1;
+        height: 19px;
+        width: calc(~"100% - 15px");
```
millerdev: no way! I didn't even know this was possible. Is that a Less thing, or is that supported in raw CSS? I guess some form of this must be supported in raw CSS since it's a dynamic calculation that can't be done at Less compile time... Awesome!

src/templates/commander.html
```diff
@@ -1,4 +1,4 @@
 <div class="fd-commander">
-    <input type="text" placeholder="<%-gettext('Enter command')%>" />
-    <button class="fd-commander-close"><i class="fa fa-close"></i></button>
+    <i class="fa fa-close fd-commander-close"></i>
+    <input type="text" class="form-control" placeholder="<%-gettext('Enter command')%>" />
```
millerdev: I find this to be an annoying CSS/HTML interaction. The close button comes after the input in the final layout, and it should be the same in the HTML! (ranting at HTML/CSS, not @orangejenny)

orangejenny: agreed, absolute styling doesn't map to HTML well

src/less-style/inputs.less
orangejenny: Yeah, `calc` is amazing. I want to say it was added to CSS sometime in the past few years and was available in LESS before that, but not 100% sure.

PR #921, opened 2017-09-25
Expose messages publicly

orangejenny:
https://manage.dimagi.com/default.asp?260647

Originally did this as discussed on slack, with messages showing next to the save button, then remembered that we already have an area for messages.

screen shot 2017-09-25 at 5 52 31 pm

On additional messages, the message bar will quickly fade out and then fade back in, to stick to vellum's "one message at a time" stance while still catching the user's eye if an additional person has opened the form.

@emord / @millerdev 
cc @nickpell

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1856,7 +1856,6 @@ define([
 
     fn._resetMessages = function (errors) {
         var error, messages_div = this.$f.find('.fd-messages');
-        messages_div.empty();
 
         function asArray(value) {
             // TODO: I don't like this array business, should be refactored away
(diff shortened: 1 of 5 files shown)


millerdev:
Hmm, that big bar is pretty disruptive when you're working on a form. I had been trying to remove uses of that thing in hopes of finally removing it completely. Maybe this is still better than where that warning used to show though?

Test failure
```
1) The javaRosa plugin with multiple languages should show warning on load for with unknown language:
      expected 0 to equal 1

      + expected - actual
      -0
      +1
      
  AssertionError@bower_components/chai/chai.js:9320:24
  assert@bower_components/chai/chai.js:239:31
  assertEqual@bower_components/chai/chai.js:1383:18
  methodWrapper@bower_components/chai/chai.js:7709:30
  strictEqual@bower_components/chai/chai.js:4272:63
  tests/javaRosa.js:288:25
```

orangejenny:
Fixed scrollbars (preexisting, but indeed ugly) and tests.

@millerdev I didn't want to create yet another area for messages, but I'd be happy to make the bar itself less obtrusive (probably in an additional PR). It could go next to the save button, or above vellum (more of a "global" position) instead of on top of it. What do you think? I'm not fully in love with either of these options, but also really dislike the current bar.

screen shot 2017-09-26 at 10 14 39 am

screen shot 2017-09-25 at 4 23 04 pm

emord:
I thought the issue was that these are hidden during full screen mode? wouldn't your first option also be hidden in that mode?

PR #924, opened 2017-10-13
Add Remote Request question type

millerdev:
spec: https://docs.google.com/document/d/1x1xz2N7I_kTUjM4dFx31vt9EuzPrOWzSojnWD062XDs

The second commit fixes an unrelated test issue.

@orangejenny @emord cc @ctsims

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1670,7 +1670,7 @@ define([
 
         this.$f.find('.fd-props-toolbar').html(this.getMugToolbar(mug));
         for (var i = 0; i < sections.length; i++) {
-            var section = sections[i];
+            var section = _.clone(sections[i]);
 
             section.mug = mug;
             section.properties = _(section.properties)
--- a/src/remoteRequest.js
+++ b/src/remoteRequest.js
@@ -1,0 +1,127 @@
+/**
+ * Remote Request question type plugin
+ */
+define([
+    'jquery',
+    'underscore',
+    'vellum/mugs',
+    'vellum/tree',
+    'vellum/util',
+    'vellum/widgets',
+    'vellum/core'
+], function (
+    $,
+    _,
+    mugs,
+    Tree,
+    util,
+    widgets
+) {
+    var REMOTE_REQUEST_TYPE = "RemoteRequest",
+        REMOTE_REQUEST_SECTIONS = [
+            {
+                slug: "main",
+                displayName: gettext("Basic"),
+                properties: [
+                    "nodeID",
+                    "url",
+                    "parameters",
+                    "comment",
+                ],
+            },
+        ],
+        RemoteRequest = util.extend(mugs.defaultOptions, {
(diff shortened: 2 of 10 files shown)


src/core.js
```diff
@@ -1670,7 +1670,7 @@ define([
 
         this.$f.find('.fd-props-toolbar').html(this.getMugToolbar(mug));
         for (var i = 0; i < sections.length; i++) {
-            var section = sections[i];
+            var section = _.clone(sections[i]);
```
orangejenny: Why is this needed?

src/remoteRequest.js
```diff
@@ -0,0 +1,127 @@
...
+                },
+                parameters: {
+                    lstring: gettext('Parameters'),
+                    visibility: 'visible',
+                    presence: 'optional',
+                    widget: widgets.xPath,
+                    xpathType: "generic",
+                    serialize: mugs.serializeXPath,
+                    deserialize: mugs.deserializeXPath,
+                    help: gettext('A reference to a parameters node, "/data/request_params"'),
```
orangejenny: This says it's a reference to a single node, but the input accepts any xpath expression. Can/should we do anything to limit what people put in this box?

orangejenny: Also, should this say something like "Drag a question here" rather than talking about a node reference? I realize this is advanced functionality, but still.

src/remoteRequest.js
```diff
@@ -0,0 +1,127 @@
...
+            typeName: gettext('Remote Request'),
+            dataType: REMOTE_REQUEST_TYPE,
+            icon: 'fa fa-globe',
+            isDataOnly: true,
+            isTypeChangeable: false,
+            supportsDataNodeRole: true,
+            writeDataNodeXML: function (xmlWriter, mug) {
+                mug.form.vellum.data.remoteRequest.requestMugs.push(mug);
+            },
+            spec: {
```
orangejenny: Can/should there be validation for this mug? Like requiring a URL to be provided?

orangejenny:
Looks good overall, my comments are just questions.

orangejenny:
(didn't mean to close this)

src/core.js
millerdev: because `section` is mutated below on line [1676](https://github.com/dimagi/Vellum/pull/924/files#diff-7eb52b366866677666470e019283c8eaR1676)

src/remoteRequest.js
millerdev: Good idea.

src/remoteRequest.js
millerdev: Unfortunately we don't have a widget type that only accepts a single ref, so I opted to use `widgets.xPath` since it is a nicer experience than `widgets.text` for drag/drop, etc. I don't think this use case is important enough to add a new "ref" widget type.

Good suggestion on the wording change.

PR #926, opened 2017-10-23
Remote Request feedback

millerdev:
Fixes for feedback from #924 and also some bug fixes.

Review by commit 🐠 

@orangejenny

### DIFF
--- a/src/commander.js
+++ b/src/commander.js
@@ -77,6 +77,11 @@ define([
             {form: cmd.vellum.data.core.form, on: _.identity},  // fake mug
             {useHashtags: true, tabSelectsMatch: true}          // options
         );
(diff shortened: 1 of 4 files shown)


src/commander.js
```diff
@@ -77,6 +77,11 @@ define([
             {form: cmd.vellum.data.core.form, on: _.identity},  // fake mug
             {useHashtags: true, tabSelectsMatch: true}          // options
         );
+        cmd.vellum.data.core.form.on("change-display-language", function() {
+            // NOTE this is tightly coupled with atwho.autocomplete(...), which
+            // calls $input.atwho('destroy') on change-display-language
+            cmd.input.atwho(cmd.atwhoConfig);
```
orangejenny: This is unfortunate, but I don't have a better solution :-/

PR #929, opened 2017-11-20
Migrate analytics

biyeun:
@orangejenny


orangejenny:
@biyeun Added a commit that should fix the js test. The python test will be fixed once the prelogin PR is merged and the submodule update is committed to this PR.

orangejenny:
@biyeun I added a commit to fix tests.

@millerdev / @emord does https://github.com/dimagi/Vellum/pull/929/commits/836d8483ed45fbb271885770ed1741c042bf570e look reasonable? I don't love having a skeletal, no-op module, but I think it's slightly cleaner than passing analytics callbacks through VELLUM_OPTIONS.

emord:
Is the point of this so that you can add an hqAnalytics module in hq? is there a reason you can't use the existing analytics.js to accomplish this?

orangejenny:
@emord The existing analytics.js depended on `window.analytics`. The HQ refactor gets rid of that global, so we still need to pass that code through somehow.

emord:
That makes sense, but why do we need two analytics files to do that? Couldn't HQ drop in analytics.js or could we just rename that to hqanalytics.js?

orangejenny:
@emord Deleted src/analytics.js in https://github.com/dimagi/Vellum/pull/929/commits/48165c9b3374eb97e526beef89c8f9739960b6b2 and added its logic to HQ in https://github.com/dimagi/commcare-hq/pull/18615/commits/5d717558c2a28082cc9fbb94580729b425f5e6ba Went with the name `hqAnalytics`.

src/analytics.js
```diff
@@ -1,23 +0,0 @@
-define([], function () {
```
millerdev: I'm kinda annoyed that this module had to be removed/replaced with one called `hqAnalytics` (it messes up git history of our analytics code). The rename seems unnecessary. Am I missing something important about why this had to be done?

PR #930, opened 2017-11-29
Fix saved after overwrite warning

millerdev:
https://manage.dimagi.com/default.asp?266075

This fixes the issue where the save button is disabled after canceled overwrite warning. I am unable to reproduce (or even understand how it would happen) the second part where the person with a conflicted form overwrites the other users work without a warning.

🐠 Review by commit

@orangejenny @emord

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -2186,8 +2186,11 @@ define([
             success: function (data) {
                 if (saveType === 'patch') {
                     if (data.status === 'conflict') {
-                        if (_.isUndefined(data.xform)) {
+                        var force_full = _this.opts()
+                            .features.full_save_on_missing_conflict_xform;
+                        if (_.isUndefined(data.xform) && force_full) {
                             // unconditionally overwrite if no xform to compare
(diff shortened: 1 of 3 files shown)


src/core.js
```diff
@@ -2186,8 +2186,13 @@ define([
             success: function (data) {
                 if (saveType === 'patch') {
                     if (data.status === 'conflict') {
-                        if (_.isUndefined(data.xform)) {
+                        // reset save button to unsaved state
+                        _this.data.core.saveButton.fire("change");
+                        var force_full = _this.opts()
+                            .features.full_save_on_missing_conflict_xform;
```
emord: Does adding this mean that user's can no longer save on conflicting changes? I don't see an HQ PR to set this feature

millerdev: No, it's a feature meant to be used in vellum (standalone mode) only. It's set in `tests/options.js` and not intended to be used anywhere else. Basically it just ensures that the branch with the comment `unconditionally overwrite if no xform to compare` is never executed in production.

PR #931, opened 2018-01-09
CommTrack module: setvalue -> bind 

millerdev:
https://manage.dimagi.com/default.asp?267194

Ignore the first commit, which just adds [package-lock.json](https://docs.npmjs.com/files/package-lock.json)

Best to review the other two commits individually 🐠 

@emord @orangejenny

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,4171 @@
+{
+    "name": "Vellum",
+    "version": "1.6.0",
+    "lockfileVersion": 1,
+    "requires": true,
+    "dependencies": {
+        "async": {
+            "version": "0.9.2",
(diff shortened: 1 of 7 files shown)


millerdev:
@dimagi/product I'm not sure if this should be documented or not. It's a pretty low-level detail about how ledger question types are written by the form builder. Maybe this should be categorized as invisible?

sheelio:
@millerdev 

I'd mark as product/feature-flag since I think Supply is behind a feature flag.  The change is that instead of using a setvalue for certain ledger entries we'd use a bind?

millerdev:
> I'd mark as product/feature-flag since I think Supply is behind a feature flag.

Done. Thanks!

> The change is that instead of using a setvalue for certain ledger entries we'd use a bind?

Correct.

PR #933, opened 2018-02-12
Fix error on load form with accented question id

millerdev:
https://manage.dimagi.com/default.asp?269812

@emord @jmtroth0

### DIFF
--- a/src/parser.js
+++ b/src/parser.js
@@ -633,11 +633,15 @@ define([
      * @return absolute nodeset path.
      */
     function processPath (path, rootNodeName, form) {
-        var newPath;
-        var parsed = form.xpath.parse(path);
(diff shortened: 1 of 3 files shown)


src/parser.js
```diff
@@ -633,11 +633,15 @@ define([
...
+        var newPath, parsed;
+        try {
+            parsed = form.xpath.parse(path);
+        } catch (ex) {
+            return path;
+        }
         if (!(parsed instanceof form.xpath.models.XPathPathExpr ||
               parsed instanceof form.xpath.models.HashtagExpr)) {
-            return null;
+            return path;
```
millerdev: Changed this to `path` because returning null from this function causes errors in other places.

PR #934, opened 2018-02-12
Allow setting data parent of question in repeat group

millerdev:
A question's data parent must now be (in) the same repeat group that is the nearest repeat control ancestor of the question. Previously questions in repeat groups could not have a custom data parent.

@emord @jmtroth0 @ctsims

### DIFF
--- a/src/mugs.js
+++ b/src/mugs.js
@@ -1029,18 +1029,7 @@ define([
             },
             dataParent: {
                 lstring: gettext('Data Parent'),
-                visibility: function(mug) {
-                    function recFunc(mug) {
-                        if (!mug) {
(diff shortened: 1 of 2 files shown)


millerdev:
@dimagi/product setting **product/invisible** tag even though this is not quite invisible (the advanced *Data Parent* field now appears on questions inside repeat groups). It's a very advanced feature that likely will only be used by dimagi staff (i.e., Clayton), so probably does not deserve an announcement.

millerdev:
> is it something that should be hidden by a feature flag?

I don't think so. It's an "Advanced" question property, and those are generally _use at your own risk_.

PR #938, opened 2018-03-19
Warn on Question ID value of "registration"

millerdev:
https://manage.dimagi.com/default.asp?272602

@orangejenny @emord cc @gcapalbo

### DIFF
--- a/src/mugs.js
+++ b/src/mugs.js
@@ -790,6 +790,14 @@ define([
         mug.p.conflictedNodeId = null;
     }
 
+    var RESERVED_NAMES = {
+        "case": gettext("The ID 'case' may cause problems with case " +
+            "management. It is recommended to pick a different Question ID."),
+        "registration": gettext("The ID 'registration' may cause problems " +
+            "with form parsing. It is recommended ot pick a different " +
+            "Question ID."),
+    };
+
     var baseSpecs = {
         databind: {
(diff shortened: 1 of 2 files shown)


src/mugs.js
```diff
@@ -790,6 +790,14 @@ define([
         mug.p.conflictedNodeId = null;
     }
 
+    var RESERVED_NAMES = {
+        "case": gettext("The ID 'case' may cause problems with case " +
+            "management. It is recommended to pick a different Question ID."),
+        "registration": gettext("The ID 'registration' may cause problems " +
+            "with form parsing. It is recommended ot pick a different " +
```
orangejenny: `ot` => `to`

src/mugs.js
```diff
@@ -790,6 +790,14 @@ define([
         mug.p.conflictedNodeId = null;
     }
 
+    var RESERVED_NAMES = {
```
orangejenny: Does it matter what specific kinds of problems the ID can cause? I'd just do a flat list of reserved words and a slightly more generic error, but don't think it matters much either way.

millerdev: 80b95c6657763fb32e166a8ba279875ebafa6cff

Fair point about using a generic message. I was trying to preserve the original message (for translations), but this makes the code simpler.

PR #939, opened 2018-03-28
Fix ref tracking

millerdev:
https://manage.dimagi.com/default.asp?273148

A bit of a deep dive, but I think a substantial improvement in `LogicManager` reference tracking. Review commits individually 🐠 

@orangejenny @emord

### DIFF
--- a/src/javaRosa/plugin.js
+++ b/src/javaRosa/plugin.js
@@ -657,6 +657,11 @@ define([
                 return options;
             }
 
+            function trackLogicRefs(widget) {
+                widget.trackLogicReferences = true;
+                return widget;
+            }
+
             function validateConstraintMsgAttr(mug) {
                 var itext = mug.p.constraintMsgItext;
                 if (!mug.p.constraintAttr && itext && !itext.isEmpty()) {
@@ -673,7 +678,7 @@ define([
                     return mug.options.isSpecialGroup ? 'notallowed' : 'optional';
                 },
                 lstring: gettext('Validation Message'),
-                widget: function (mug, options) {
+                widget: trackLogicRefs(function (mug, options) {
                     return itextBlock.label(mug, $.extend(options, {
                         itextType: "constraintMsg",
                         messagesPath: "constraintMsgItext",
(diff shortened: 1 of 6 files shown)


src/javaRosa/plugin.js
```diff
@@ -673,7 +678,7 @@ define([
                     return mug.options.isSpecialGroup ? 'notallowed' : 'optional';
                 },
                 lstring: gettext('Validation Message'),
-                widget: function (mug, options) {
+                widget: trackLogicRefs(function (mug, options) {
```
millerdev: Unfortunately more indirection here, and it kinda fragile. We cannot set `widget.trackLogicReferences` on the base widget since widgets use a functional style where each derived widget wraps the function/constructor of the widget it's wrapping, effectively hiding all attributes on the inner function/constructor.

@orangejenny this is a disadvantage to the new functional class convention we just adopted (avoiding the `new` operator). However, I think it's worth working around it in a few cases like this rather than deal with the other shortcomings of `new`.

src/logic.js
```diff
@@ -162,12 +146,6 @@ define([
                     replacePathInfo(this._xpathParser.parse(to), path);
                 }
             }
-            for (i = 0; i < hashtags.length; i++) {
-                path = hashtags[i];
-                if (path.toHashtag() === from) {
-                    replacePathInfo(this._xpathParser.parse(to), path);
-                }
-            }
```
millerdev: This change was not strictly necessary. It's DRYing things up a bit, along with the change above on line 130.

millerdev:
Test failure is real. Working on it.

src/logic.js
```diff
@@ -162,12 +145,6 @@ define([
                     replacePathInfo(this._xpathParser.parse(to), path);
                 }
             }
-            for (i = 0; i < hashtags.length; i++) {
-                path = hashtags[i];
-                if (path.toHashtag() === from) {
```
emord: No code fills me with more regret than when I see `toHashtag` in vellum code. Thanks for hiding some of it :)

src/javaRosa/plugin.js
orangejenny: Yeah, I see your point. FWIW, I'm not so opposed to `new` in vellum or to generally having slightly different conventions in vellum versus HQ. I'm not categorically opposed to `new` - I don't like it but think it can be okay if used consistently. It worries me in HQ because we use a mix of inheritance styles and are sometimes sloppy about doing `var self = this` in a context where `this` isn't what you'd expect, so it's easy for someone to copy/paste bad code. In vellum the code is so much more consistent that I don't think that's as high of a risk.

PR #941, opened 2018-05-19
Mk icds app trans hq 2

mkangia:
ICDS is considering stretching to 15+ languages and have run some tests for a bottleneck with form builder not able to support that many lanugages. While we are working on other possibilities, i believe the blocker is heavy HTML and this would help tackle that.
I did some test and this change should do the following:
render display text in form view for the current language and the default language. The reason for keeping both of them and not just the current language/default language is to avoid some issue i saw when adding questions with just current language display text available. But if default langugage is the current lang, it will just show one language.

One can still do translations by switching to the other language but i don't think the app builders use that in ICDS. Translations are handled by bulk app translations upload.

Derek is going to run some test again early next week so would like to deploy this before that if possible.
If you guys think this is worth a short, i would just add this [commit](https://github.com/dimagi/commcare-hq/commit/8cbfd76a474bd3f18d29ea3d31294e4dcceb896f) to HQ.

@esoergel

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -228,6 +228,7 @@ define([
             this.opts().core.dataSourcesEndpoint,
             this.opts().core.invalidCaseProperties
         );
+        this.data.core.showOnlyCurrentLang = this.opts().core.showOnlyCurrentLang;
     };
 
     fn.postInit = function () {
--- a/src/core.js
+++ b/src/core.js
@@ -228,7 +228,6 @@ define([
             this.opts().core.dataSourcesEndpoint,
             this.opts().core.invalidCaseProperties
         );
-        this.data.core.showOnlyCurrentLang = this.opts().core.showOnlyCurrentLang;
     };
 
     fn.postInit = function () {
--- a/src/javaRosa/itextBlock.js
+++ b/src/javaRosa/itextBlock.js
@@ -77,7 +77,7 @@ define([
             _.each(block.getForms(), function (form) {
                 var $formGroup = block.getFormGroupContainer(form);
                 var langs_to_show = block.languages
-                if(options.vellum.data.core.showOnlyCurrentLang) {
+                if(options.vellum.data.javaRosa.showOnlyCurrentLang) {
                     langs_to_show = _.uniq([
                     options.vellum.data.javaRosa.Itext.defaultLanguage,
                     options.vellum.data.core.currentItextDisplayLanguage])
--- a/src/javaRosa/itextBlock.js
+++ b/src/javaRosa/itextBlock.js
@@ -78,7 +78,9 @@ define([
                 var $formGroup = block.getFormGroupContainer(form);
                 var langs_to_show = block.languages
                 if(options.vellum.data.core.showOnlyCurrentLang) {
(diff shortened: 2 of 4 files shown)


orangejenny:
I like this approach: it's good for performance and also reduces clutter. And thinking about it, it does seem unlikely that projects with more than 2 or 3 languages actually want/need to be looking at all of the translations all of the time.

That makes me want this to be more readily available than being behind a feature flag. What would you think of making `showOnlyCurrentLang` a form property (available in the same menu where you change language), and if your form has more than 3(?) languages, turning it on and showing a message that it's turned on and explaining how to turn it off (using the existing warning/message infrastructure - the ui for that isn't perfect, but that's a separate issue)?

src/core.js
```diff
@@ -228,6 +228,7 @@ define([
             this.opts().core.dataSourcesEndpoint,
             this.opts().core.invalidCaseProperties
         );
+        this.data.core.showOnlyCurrentLang = this.opts().core.showOnlyCurrentLang;
```
millerdev: This should be a javaRosa option

https://github.com/dimagi/Vellum/blob/34a8db231ae022b476126c494dcc786dc0af7324/tests/options.js#L176-L179

not a core option. Although I would not add a value in `tests/options.js` because that will change the default rendering for tests.

Also, this should be moved into

https://github.com/dimagi/Vellum/blob/34a8db231ae022b476126c494dcc786dc0af7324/src/javaRosa/plugin.js#L62-L66

The value should be stored in `this.data.javaRosa` instead of `this.data.core`.

mkangia:
@orangejenny 
> That makes me want this to be more readily available than being behind a feature flag. What would you think of making showOnlyCurrentLang a form property

Ya. I was thinking this would be useful for all. I think this should be an app property instead of a form property guessing one would want to to set it for all form at once instead of setting it per form again and again while form building. Also would want this to go through QA if this is going beyond ICDS for all users. So fine to release this behind FF as of now and then make it available for all with QA?

src/core.js
mkangia: Sure @millerdev. I was not sure where to put this. Will move it to `javaRosa` now.

mkangia:
thanks for the hints to expected changes @millerdev. Moved option to javaRosa and added tests.

mkangia:
> Maybe it's an app setting but you have access to it on the languages page? Anyway, no need to decide right now.

@orangejenny 
Ya.. i was thinking some settings or add-on. I guess you are better aware of app builder to suggest the right place for it though i can take a tour and see if i can find a good place. let me know if you have suggestions on where it should go and i will add it.

orangejenny:
> let me know if you have suggestions on where it should go and i will add it.

I had an idea: how about doing it as an app setting with three values: on, off, and smart? Default to smart, which would show all languages if there are 1 or 2, but more than that and it'll only show the current language plus default language.

There could be a message in the Language List section, where when you add a third language you get a message like "When working in a form, you will only see the currently-selected language. Click here to learn how to change this behavior." and link to docs. (There's already messaging in that section to display a "Please save" message so it's likely easy to add another one.)

PR #942, opened 2018-05-21
Update less

millerdev:
@emord

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,27 +4,68 @@
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {
-        "async": {
-            "version": "0.9.2",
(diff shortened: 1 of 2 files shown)


emord:
```
TypeError: Cannot read property 'contents' of undefined
In module tree:
    main
      vellum/core
        less/less
Error: TypeError: Cannot read property 'contents' of undefined
In module tree:
    main
      vellum/core
        less/less
    at Object.parse (/home/travis/build/dimagi/Vellum/node_modules/less/lib/less/parser/parser.js:168:21)
make: *** [_rjs] Error 1
```

millerdev:
Turns out that error is because [require-less does not support less 2+](https://github.com/guybedford/require-less/issues/77), and it's no longer maintained. Looks like we need a new solution to using less with requirejs. FYI @orangejenny

PR #944, opened 2018-06-06
set label itext

mkangia:
Saw a test failing on local system which is actually passing on travis.
To be honest, i am not sure how these two tests are passing on travis.

I would expect `util.addQuestion("Text", "question")` to just add a question of type Text with id question. But the test then checks for its label and expected it to be same as id. passes on travis but fails on local. 
Checked on HQ otherwise and looks like the label is not populated by default.

Made this change to set label itext first and now test passes locally.

### DIFF
--- a/tests/javaRosa.js
+++ b/tests/javaRosa.js
@@ -511,8 +511,9 @@ define([
 
         it("itext changes do not bleed back after copy", function () {
             util.loadXML("");
-            var mug = util.addQuestion("Text", "question"),
-                dup = mug.form.duplicateMug(mug);
+            var mug = util.addQuestion("Text", "question");
+            mug.p.labelItext.set('question');
+            var dup = mug.form.duplicateMug(mug);
             dup.p.labelItext.set("q2");
 
             util.saveAndReload(function () {
@@ -523,8 +524,9 @@ define([
 
         it("itext changes do not bleed back from copy of copy", function () {
             util.loadXML("");
-            var mug = util.addQuestion("Text", "question"),
-                dup = mug.form.duplicateMug(mug),
+            var mug = util.addQuestion("Text", "question");
+            mug.p.labelItext.set('question');
+            var dup = mug.form.duplicateMug(mug),
                 cpy = mug.form.duplicateMug(dup);
             cpy.p.labelItext.set("copy");
 


tests/javaRosa.js
```diff
@@ -511,8 +511,9 @@ define([
 
         it("itext changes do not bleed back after copy", function () {
             util.loadXML("");
-            var mug = util.addQuestion("Text", "question"),
-                dup = mug.form.duplicateMug(mug);
+            var mug = util.addQuestion("Text", "question");
+            mug.p.labelItext.set('question');
```
millerdev: The test was previously setup to verify a very specific test sequence, and modifying it this way doesn't feel right because you're effectively overriding the internal workings of javaRosa that are being tested. The test passes for me, but now that I'm looking closely at it and I have a bit of context I can see that the test is broken (has been since it was originally written).

The first problem is that this test is doing its final verification with `util.saveAndReload`, which is asynchronous. But the test itself is not an async test (it doesn't use a `done` callback), so that final verification can fail and not be reported, depending on the order of async callbacks happen.

The second problem is that sometime vellum changed from using the Question ID as the default label itext (this made more sense when the Question ID field was above the label input on the question properties screen) to having the default be an empty string. Maybe this was only ever true for Choice question type? But at some point it was removed 5887d9719cc2a7f01cde6d2b08e0491305a2cb34. Also possibly related is this change: afb3a1e6607c98f6ad1c6be5eeecccad44513771. Anyway, that is no longer true, so I think it's safe to change these tests. I made a new PR: https://github.com/dimagi/Vellum/pull/945

mkangia:
Thank You @millerdev

tests/javaRosa.js
mkangia: @millerdev 
> The test passes for me

Oh.. That one was definitely failing for me on my local system. 
Plus,
> sometime vellum changed from using the Question ID as the default label itext (this made more sense when the Question ID field was above the label input on the question properties screen) to having the default be an empty string. 

If this is true, how is it even passing? It should definitely fail.

millerdev: @mkangia 

> > sometime vellum changed from using the Question ID as the default label itext...
>
> If this is true, how is it even passing?

because of the first problem:

> the test itself is not an async test (it doesn't use a `done` callback), so that final verification **can fail and not be reported**, depending on the order of async callbacks

mkangia: Oh wow!.. thanks @millerdev 
I think i didn't absorb `can fail and not be reported` correctly.

PR #947, opened 2018-07-20
required condition

jmtroth0:
I'm mainly interested in this point on feedback on both UX and whether or not I found all the things I needed to change.

Right now it adds an input box in the logic section for the required condition. If you enter anything while it isn't marked "required" then it will warn you off. I was thinking of making it so the box only appeared if required was checked, but I wasn't sure about how to do that or if it was that much better.

https://trello.com/c/pb9LDmk8/6-make-it-possible-to-define-a-function-for-required-on-a-question-jen

Not sure who usually reviews vellum stuff --> @millerdev @orangejenny buddy: @snopoke 

I definitely recommend `w=1`, but probably can 🏛 . Only commits 2fac576, 685e7ea, and 05a8363 do anything

screen shot 2018-07-20 at 11 41 08 am

### DIFF
--- a/src/mugs.js
+++ b/src/mugs.js
@@ -985,7 +985,7 @@ define([
                     var warningKey = "mug-requiredCondition-warning",
                         warningAttrs = ["requiredAttr", "requiredCondition"];
                     if (mug.p.requiredCondition && !mug.p.requiredAttr) {
-                        message = gettext("The condition will be ignored unless you mark the question required.");
+                        var message = gettext("The condition will be ignored unless you mark the question required.");
                         _.each(warningAttrs, function (attr) {
                             mug.addMessage(attr, {
                                 key: warningKey,
--- a/src/mugs.js
+++ b/src/mugs.js
@@ -982,11 +982,19 @@ define([
                 serialize: serializeXPath,
                 deserialize: deserializeXPath,
                 validationFunc: function (mug) {
+                    var warningKey = "mug-requiredCondition-warning",
+                        warningAttr = "requiredCondition";
                     if (mug.p.requiredCondition && !mug.p.requiredAttr) {
-                        return gettext('You cannot have a Required Condition if the question isn\'t required!');
+                        message = gettext("Since this question isn't required, this condition will be ignored.");
+                        mug.addMessage(warningAttr, {
+                            key: warningKey,
+                            level: mug.WARNING,
+                            message: message
+                        });
                     } else {
-                        return 'pass';
+                        mug.dropMessage(warningAttr, warningKey);
                     }
+                    return 'pass';
                 }
             },
             nodeset: {
--- a/src/mugs.js
+++ b/src/mugs.js
@@ -1407,7 +1407,7 @@ define([
         }
     });
 
-    // Deprecated. Users may not add new longs to forms, 
+    // Deprecated. Users may not add new longs to forms,
     // but must be able to view forms already containing longs.
     var Long = util.extend(Int, {
         typeName: gettext('Long'),
@@ -1576,7 +1576,7 @@ define([
             defaultValue: { presence: 'optional', visibility: 'hidden' },
         }
     });
-    
+
     // This is just a group, but appearance = 'field-list' displays it as a list
     // of grouped questions.  It's a separate question type because it can't
     // nest other group types and it has a very different end-user functionality
@@ -1668,7 +1668,7 @@ define([
             }
         }
     });
-   
+
     function MugTypesManager(baseSpec, mugTypes, opts) {
         var _this = this,
             // Nestable Field List not supported in CommCare before v2.16
@@ -1680,7 +1680,7 @@ define([
         this.baseSpec = baseSpec;
 
         MugProperties.setBaseSpec(
-            util.extend.apply(null, 
+            util.extend.apply(null,
                 [baseSpec.databind, baseSpec.control].concat(_.filter(
                     _.pluck(
                         util.extend(this.normalTypes, this.auxiliaryTypes),
@@ -1690,7 +1690,7 @@ define([
         this.allTypes = $.extend({}, this.auxiliaryTypes, this.normalTypes);
 
         var allTypeNames = _.keys(this.allTypes),
-            innerChildTypeNames = _.without.apply(_, 
+            innerChildTypeNames = _.without.apply(_,
                   [allTypeNames].concat(_.keys(this.auxiliaryTypes)));
 
         if (!group_in_field_list) {
--- a/src/mugs.js
+++ b/src/mugs.js
@@ -970,6 +970,15 @@ define([
                 lstring: gettext("Required"),
                 widget: widgets.checkbox
             },
+            requiredCondition: {
+                visibility: 'requiredAttr',
+                presence: 'optional',
+                lstring: gettext("Required Condition"),
+                widget: widgets.xPath,
+                xpathType: "bool",
+                serialize: serializeXPath,
+                deserialize: deserializeXPath
+            },
             nodeset: {
                 visibility: 'hidden',
                 presence: 'optional' //if not present one will be generated... hopefully.
@@ -1183,7 +1192,7 @@ define([
                 constraint: mug.p.constraintAttr,
                 "jr:constraintMsg": constraintMsg,
                 relevant: mug.p.relevantAttr,
-                required: util.createXPathBoolFromJS(mug.p.requiredAttr),
+                required: mug.p.requiredCondition || util.createXPathBoolFromJS(mug.p.requiredAttr),
                 calculate: mug.p.calculateAttr,
             };
             _.each(mug.p.rawBindAttributes, function (value, key) {
--- a/src/mugs.js
+++ b/src/mugs.js
@@ -977,7 +977,14 @@ define([
                 widget: widgets.xPath,
                 xpathType: "bool",
                 serialize: serializeXPath,
-                deserialize: deserializeXPath
+                deserialize: deserializeXPath,
+                validationFunc: function (mug) {
+                    if (mug.p.requiredCondition && !mug.p.requiredAttr) {
+                        return gettext('You cannot have a Required Condition if the question isn\'t required!');
+                    } else {
+                        return 'pass';
+                    }
+                }
             },
             nodeset: {
                 visibility: 'hidden',
--- a/src/mugs.js
+++ b/src/mugs.js
@@ -1204,13 +1204,22 @@ define([
             } else {
                 constraintMsg = mug.p.constraintMsgAttr;
             }
+
+            var required;
+            if (mug.p.requiredAttr) {
+                required = mug.p.requiredCondition || util.createXPathBoolFromJS(mug.p.requiredAttr);
+            } else {
+                required = util.createXPathBoolFromJS(mug.p.requiredAttr);
+            }
+
             var attrs = {
                 nodeset: mug.hashtagPath,
                 type: mug.options.dataType,
                 constraint: mug.p.constraintAttr,
                 "jr:constraintMsg": constraintMsg,
                 relevant: mug.p.relevantAttr,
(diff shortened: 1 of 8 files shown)


jmtroth0:
I realized I could also ping @proteusvacuum and @nickpell as solutions people

orangejenny:
What happens when you have a validation condition and then uncheck the required box?

I agree it'd be nice to only display the input when the question is required, or to disable it when the question isn't required. Other people in @dimagi/ux might have further thoughts.

jmtroth0:
Agreed on only displaying it/disabling it when not applicable, but as I said, I'll need some help figuring out how to do that. I would guess there is a way to just make it happen, like it does for the validation messages, but I couldn't find it just by grepping around

jmtroth0:
>What happens when you have a validation condition and then uncheck the required box?

Validations run when you hover over the "save" box, so it catches it then. I'll have the same validation func run on the `requiredAttr` as well to improve that

orangejenny:
Good point, I forgot about the validation during save. It's unfortunate the the two inputs are far apart from each other, but I don't think it'd be good to move this up to where the checkbox is. Thinking about it...it might make sense to not make this an error/warning at all and just ignore the condition if the question isn't required. But then, I could imagine people writing a condition and forgetting to check the box. Thoughts?

jmtroth0:
I added a41a373, which also puts the same check on the required checkbox and gives the same warning. I think that's a pretty good balance. Thoughts?

src/parser.js
```diff
@@ -682,15 +682,16 @@ define([
             constraintAttr: parseVellumAttrs(form, el, 'constraint'),
             constraintMsgAttr: lookForNamespaced(el, "constraintMsg"),
             requiredAttr: parseBoolAttributeValue(el.popAttr('required')),
+            requiredCondition: el.popAttr('required'),
```
snopoke: shouldn't this be `el.popAttr('requiredCondition')`?

src/parser.js
```diff
@@ -515,15 +515,15 @@ define([
             return undefined;
         }
         var str = attrString.toLowerCase().replace(/\s/g, '');
-        if (str === 'true()') {
-            return true;
-        } else if (str === 'false()') {
+        if (str === 'false()') {
             return false;
+        } else if (str) {
+            return true;
```
snopoke: why was this change required?

millerdev:
> it might make sense to not make this an error/warning at all and just ignore the condition if the question isn't required. But then, I could imagine people writing a condition and forgetting to check the box.

I like this idea. What if we automatically checked the required box when a new required condition is added?

millerdev:
Thinking a bit more about that, it becomes problematic if we automatically check the required checkbox in the following scenario:

- Valery intends to write a validation condition, but accidentally begins entering in the Required Condition input.
- She realizes her mistake, removes the text from Required Condition and puts it in the Validation Condition input.
- Her question is now required, but she didn't intent for that to happen. It was a side-effect of accidentally typing in the Required Condition input.

src/parser.js
jmtroth0: the attribute value will no longer necessarily be 'true()' or 'false()'. there are now many other truthy values

millerdev:
> It's unfortunate the the two inputs are far apart from each other, but I don't think it'd be good to move this up to where the checkbox is.

What if we put _Required Condition_ up with the _Required_ checkbox, but only show it if the _Required_ checkbox is checked? That way we can hide it and ignore its value when the _Required_ checkbox is not checked.

What does it mean when the _Required_ checkbox is checked and the _Required Condition_ is empty? Should the value of _Required Condition_  be automatically set to `true()` when _Required_ is checked and _Required Condition_ is empty? Or should we automatically convert a value of `true()` to empty _Required Condition_ when parsing the form?

orangejenny:
> What if we put Required Condition up with the Required checkbox, but only show it if the Required checkbox is checked?

That's interesting, the condition box then wouldn't looks like clutter for non-required questions. On the other hand, I wonder how often this new input is going to be used, and if it's going to be distracting to check a box for a very basic and intuitive action (making a question required) and then have something suddenly appear that has a much less intuitive purpose. The app manager UI for this kind of action is much more new-user-friendly (in form settings, case management, when closing a case, the button that says `Only if the answer to...` and then you click it and get an input to enter a condition).

I'm leaning towards keeping the two inputs separate (checkbox in the top section, condition in Logic) and not worrying about appearing/disappearing since they're not near each other...which is to say, exactly what's in this PR. I like that https://github.com/dimagi/Vellum/commit/a41a373ee23d7604836b6e632a8596d89c05e11d added the warning to both the checkbox and the condition. I might change the warning text to something like "Since this question isn't required, this condition will be ignored." so it doesn't sound like the user did something wrong.

millerdev:
> I might change the warning text to something like "Since this question isn't required, this condition will be ignored." so it doesn't sound like the user did something wrong.

Will the _Required Condition_ content be lost if the user unchecks _Requried_ and then saves and reloads the form? I think it will. Is that a problem?

orangejenny:
@millerdev I can imagine someone having a complex condition, wanting to turn it off temporarily while testing, and then being unhappy about having to clear and re-enter it. But it's probably not worth a lot of effort if that's difficult behavior to change.

millerdev:
> I wonder how often this new input is going to be used, and if it's going to be distracting to check a box for a very basic and intuitive action... and then have something suddenly appear that has a much less intuitive purpose.

I agree. I'll also mention that if a question's display condition evaluates to false then the required status and validation conditions are ignored (just verified this with Clayton), so it's possible to implement this probably rarely used feature using two questions (one required, one not) with mutually exclusive display conditions. So maybe it's not even worth implementing this if it's rarely used and the UX is hard to do well?

ohmegasquared:
(👋, I'm the new product designer) 

For most this and most cases, I would recommend disabling over hiding, with helper text for the user to self resolve the issue. That way, if the user wanted to add a required condition and see that it was disabled, that would trigger her to turn it on (especially if there was helper text saying something like 'To require a condition, mark the question as required.' and/or possibly a link, 'Click to mark question as required.' ) I wouldn't assume that typing into a required condition field would make my question required, unless we had some research or data that supported that. 

I'm happy to chat more about it if wanted, and provide some UX articles if desired. Overall I think we want to show the potential of requiring conditions, so it makes sense to disable rather than hide. 

Also unsure why sometimes are warnings are red and sometimes yellow.

millerdev:
> I wouldn't assume that typing into a required condition field would make my question required

Agreed. My follow-up comment to that suggestion concluded that it was a bad idea. Sorry if that wasn't clear.

jmtroth0:
I've updated this pr to now show the message in both places, as well as saving the condition separately from the requiredAttr so that it will show up when parsing the doc even if the question is not required.

Any further thoughts?

tagging people who commented @millerdev @orangejenny @nickpell @snopoke @ohmegasquared

src/mugs.js
```diff
@@ -1177,13 +1208,22 @@ define([
             } else {
                 constraintMsg = mug.p.constraintMsgAttr;
             }
+
+            var required;
+            if (mug.p.requiredAttr) {
+                required = mug.p.requiredCondition || util.createXPathBoolFromJS(mug.p.requiredAttr);
```
millerdev: It's not clear why `required` and `requiredCondition` attributes need to be saved separately.

I think `required` should be one of
- `"true()"` – _Required_ is checked, _Required Condition_ is empty
- `null` or `"false()"` – _Required_ is not checked
- XPath expression in _Required Condition_ – _Required_ is checked, _Required Condition_ is not empty

The value of `requiredCondition` can be inferred from the value of the `required` attribute on parsing:

- `"true()"` – _Required_ is checked, _Required Condition_ is empty
- not set or `"false()"` – _Required_ is not checked, _Required Condition_ is empty
- other XPath expression – _Required_ is checked, _Required Condition_ has value stored in `required` attribute.

src/parser.js
```diff
@@ -676,21 +676,25 @@ define([
...
+
         var attrs = {
             relevantAttr: parseVellumAttrs(form, el, 'relevant'),
             calculateAttr: parseVellumAttrs(form, el, 'calculate'),
             constraintAttr: parseVellumAttrs(form, el, 'constraint'),
             constraintMsgAttr: lookForNamespaced(el, "constraintMsg"),
-            requiredAttr: parseBoolAttributeValue(el.popAttr('required')),
+            requiredAttr: parseBoolAttributeValue(required),
+            requiredCondition: (el.popAttr('requiredCondition') ||
+                                ['true()', 'false()'].indexOf(required) === -1 ? required : undefined),
```
millerdev: This is hard to follow. Would it be acceptable to use only the required attribute and the parsing algorithm I outlined above?

src/mugs.js
jmtroth0: I did this because of the above conversation about being able to save the state of the condition without having the condition be active. Sorry for not making that clear. It seems like a nice thing to be able to disable that to test things without losing the value of that field

Your description here is what it was before this set of commits. Shall I revert 070c641 to make it that again?

src/parser.js
jmtroth0: See the comment above. I can clean this up and not use a ternary nested in an if statement if this route is acceptable

src/mugs.js
millerdev: Ugh, sorry @jmtroth0, my preference is to use the algorithm outlined above. I think it's preferable to have a normalized and unambiguous source of truth in the serialized state rather than allowing the attributes to get out of sync and have ambiguous outcome. If we want to save the `requiredCondition` we could put its value in a `vellum:requiredCondition` attribute, which would be used as a hint for vellum only (`vellum:` attributes get stripped before XML is sent to the phone).

jmtroth0: I believe that is what this is doing, but I can check. Do you still have a preference against having this at all?

millerdev: Preserving it probably creates a better UX, so I guess it's a good idea to do that. And forget what I said about creating a `vellum:` attribute, that won't work in this context.

jmtroth0:
@millerdev I ended up putting the tests into this pr. Let me know what you think

tests/parser.js
```diff
@@ -144,6 +146,28 @@ define([
             assert(!$('[name=property-dataValue]').length);
         });
 
+        it("should set required correctly", function() {
+            util.loadXML(REQUIRED_XML);
+
+            var q1 = util.getMug("question1");
+            assert(q1.p.required = "true()");
+            assert(q1.p.requiredCondition = "true()");
```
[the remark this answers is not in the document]
jmtroth0: ooh good catch. moving between syntaxes always throws me off

tests/parser.js
```diff
@@ -148,24 +148,29 @@ define([
...
-
             var q1 = util.getMug("question1");
-            assert(q1.p.required = "true()");
-            assert(q1.p.requiredCondition = "true()");
-            assert(q1.messages.get("requiredAttr").length === 0);
-            assert(q1.messages.get("requiredCondition").length === 0);
+            assert(q1.p.requiredAttr, "required is false")
+            assert.equal(q1.p.requiredCondition, undefined);
+            assert(q1.messages.get("requiredAttr").length === 0,
+                   "requiredAttr has " + q1.messages.get("requiredAttr").length + " messages");
```
millerdev: Nit: curious why you didn't use `assert.equal`?
```js
assert.equal(q1.messages.get("requiredAttr").length, 0, "requiredAttr");
```
Note the third argument can be used for disambiguation when there are multiple asserts in the same test.

It would generate failure messages like

```
The parser should set required correctly
  requiredAttr: expected 1 to equal 0
```

Even better:
```js
assert.deepEqual(q1.messages.get("requiredAttr"), [], "requiredAttr");
```
would show the messages if there is a failure. This can be very useful for debugging test failures.

jmtroth0: yeah, as you note here it's mainly because of the message generated. it seemed clearer than just some arbitrary `expected 1 to equal 0`. I hadn't tried out a third arg for assert.equals though. I'll add something like that

PR #954, opened 2019-02-15
Select questions via URL

proteusvacuum:
This is a retry of https://github.com/dimagi/Vellum/pull/663

It was reverted here: https://github.com/dimagi/Vellum/pull/688 with the comment:
> Seems to be occasionally causing issues when loading a form for the first time (sometimes tries to load `form/source/form/question` instead of `form/source/#form/question`)

I noticed that with some question types (e.g. a `balance` or `transfer`) the hashtag doesn't actually exist (returns a `/data/path`) which seems like it could have caused the issues seen in the previous PR. The biggest change I made was to ensure that the `hashtagPath` actually starts with a `#` character.

This also addresses [this comment](https://github.com/dimagi/Vellum/pull/663#issuecomment-219426746) from the initial PR. 

This is my functional PR in Vellum, so looking forward to the review!

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -965,6 +965,7 @@ define([
             } else if (selected.length < 2) {
                 var mug = _this.data.core.form.getMugByUFID(selected[0]);
                 _this.displayMugProperties(mug);
+                window.history.replaceState(null, null, mug.hashtagPath);
             } else {
                 _this.displayMultipleSelectionView();
             }
@@ -1305,7 +1306,7 @@ define([
     };
 
     fn.loadXML = function (formXML, options) {
-        var form, _this = this;
+        var form, _this = this, qId = window.location.hash;
         _this.data.core.$tree.children().children().each(function (i, el) {
             _this.jstree("delete_node", el);
         });
@@ -1322,7 +1323,7 @@ define([
         this.onXFormLoaded(form);
         if (formXML) {
             _this._resetMessages(_this.data.core.form.errors);
-            _this._populateTree();
+            _this._populateTree(qId);
         }
 
         form.on('question-type-change', function (e) {
@@ -1437,7 +1438,7 @@ define([
         }
     };
 
-    fn._populateTree = function () {
+    fn._populateTree = function (qId) {
         // NOTE: this performs the final step in the mug parsing process.
         // It should only be called once after a new XForm is loaded.
         var _this = this,
@@ -1454,18 +1455,27 @@ define([
                 _this.setTreeActions(mug);
             }
         });
-        this.selectSomethingOrHideProperties(true);
+
+        var mug = qId && _this.getMugByPath(qId) || undefined;
+        _this.selectSomethingOrHideProperties(true, mug && mug.ufid);
     };
 
-    fn.selectSomethingOrHideProperties = function (forceDeselect) {
+    fn.selectSomethingOrHideProperties = function (forceDeselect, ufid) {
         if (forceDeselect) {
             this.jstree('deselect_all');
         }
         // ensure something is selected if possible
         if (!this.jstree('get_selected').length) {
             // if there's any nodes in the tree, just select the first
-            var all_nodes = this.data.core.$tree.find("li");
-            if (all_nodes.length > 0) {
+            var all_nodes = this.data.core.$tree.find("li"),
+                selected;
+            if (ufid) {
+                selected = all_nodes.filter('[id= ' + ufid + ']');
+            }
+            if (selected && selected.length > 0) {
+                this.jstree('select_node', selected[0]);
+                return true;
+            } else if (all_nodes.length > 0) {
                 this.jstree('select_node', all_nodes[0]);
                 return true;
             } else {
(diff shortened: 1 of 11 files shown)


proteusvacuum:
The tests pass, however there are a few places where there is this message which shows up in the Travis logs, which seems to be related to this change:

```
Error loading resource file:///#form/child/item (202). Details: Cannot open file:///#form/child/item: Path is a directory
```

tests/saveToCase.js
```diff
@@ -100,6 +100,9 @@ define([
             assert.equal(index.p.user_id, "/data/meta/userID");
             assert.equal(index.p.case_id, "/data/meta/caseID");
             util.assertXmlEqual(call("createXML"), INDEX_PROPERTY_XML);
+
+            // clean up from clickQuestion
+            window.history.replaceState(null, null, '/');
```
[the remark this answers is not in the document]
proteusvacuum: These were the only ones that were causing other tests to fail, but you are right that it is probably better to clean up from all `clickQuestion`s in the other places.

I tried looking for something like `addCleanup`, so I could clean up automatically but it seems like it doesn't exist in mocha (https://github.com/mochajs/mocha/issues/2741). 

The other option I could see is to run this in a global `afterEach` which would run after every single test, but that seemed like a bit of overkill to me.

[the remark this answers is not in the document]
millerdev: Could you add this at the end of `clickQuestion()`? It's not clear to me why this is even necessary.

src/core.js
```diff
@@ -1709,6 +1719,19 @@ define([
         this.toggleConstraintItext(mug);
     };
 
+    fn._setURLHash = function (mug) {
+        if (mug && mug.hashtagPath &&
+            mug.hashtagPath.startsWith("#") &&
+            mug.hashtagPath.indexOf('undefined') === -1 &&
```
millerdev: could this possibly cause a false positive if the name of the question actually contained "undefined"?

proteusvacuum: Yes. Is there a canonical way of figuring out whether the question has a defined path or not? When you make a new question in a group, for example, it has a hashtagPath of `#form/group_name/undefined`.

tests/saveToCase.js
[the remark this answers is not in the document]
proteusvacuum: If you call `clickQuestion` it adds question path to the URL, then in subsequent tests, when the form loads, it immediately selects the question that has that path. For things like save-to-case this actually changes the behaviour, because when you click the question it adds new empty values for things like the update properties. 

I suppose I could add it to the end of `clickQuestion`, although that would lead to different behaviour than what happens "in the wild" (which seems like what some of these test utils do anyway, so it might not be surprising).

src/core.js
```diff
@@ -1719,6 +1719,18 @@ define([
...
 
+    fn._setURLHash = function (mug) {
+        if (mug && mug.absolutePathNoRoot
+            && mug.absolutePathNoRoot.indexOf('undefined') === -1
+            && !mug.absolutePathNoRoot.endsWith('/')) {
+            window.history.replaceState(null, null, "#form" + mug.absolutePathNoRoot);
+        }
+        else {
+            // If the mug doesn't have a question id yet, remove the hash from the url
+            window.history.replaceState(null, null, '/');
```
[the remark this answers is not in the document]
proteusvacuum: Err... you are correct. I will fix this.

src/core.js
millerdev: I think `mug.getNodeID() !== undefined` is what you want here. However, I forget what that will return for Choice questions. Also, it may not do what you want for transaction question types.

tests/saveToCase.js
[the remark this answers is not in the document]
millerdev: Another option would be to clear the hashtag path in `util.loadXML()` before a new form is loaded. It may be necessary to manually clean up in some tests that do not use `util.loadXML()`. That would be a little dirty because the test is cleaning up state created by the test that came before it. However, I consider having to remember to add this cleanup to every test that calls `clickQuestion()` a little dirty as well because it's not obvious and easy to forget.

src/core.js
```diff
@@ -225,6 +225,10 @@ define([
             analytics.workflow("Clicked on easy reference popover's link to show in tree");
         });
 
+        $(window).on('hashchange', function () {
+            _this.selectSomethingOrHideProperties(true, window.location.hash);
```
proteusvacuum: I wasn't able to write a test for this because the `onhashchange` event seems to trigger some arbitrary amount of time after the hash actually changes. Short of putting some kind of sleep statement in the test, I just made sure that this worked manually in the UI.

proteusvacuum:
@millerdev this is currently on staging and I've submitted a QA request for it.

proteusvacuum:
@millerdev this passed QA, but @Ben-Talbot asked me to scroll the selected node into view (which is totally reasonable). Could you give a final look at the latest commit? https://github.com/dimagi/Vellum/pull/954/commits/575520d1edc3dc178663c72624edb0cc5987ee90

src/core.js
```diff
@@ -1481,9 +1481,11 @@ define([
             }
             if (selected && selected.length > 0) {
                 this.jstree('select_node', selected[0]);
+                selected[0].scrollIntoView();
```
millerdev: Did you consider using our scroll function: `this.scrollTreeTo(ufid)`

It looks like we're not using `scrollIntoView` anywhere else.

proteusvacuum: Updated: https://github.com/dimagi/Vellum/pull/954/commits/e0f0dc35a8e8a399ff3ab03aa201f207da46a981

PR #955, opened 2019-04-25
Fix jumping to first question when links are clicked

proteusvacuum:
https://dimagi-dev.atlassian.net/browse/HI-557

This fixes 2 issues:
- Help bubbles for section headers were not showing up in firefox (so I never noticed this issue)
- Clicking any "empty link" with `href="#"` was resetting the url anchor, and therefore jumping the user to the first question. (https://stackoverflow.com/a/19167175/2957657)


src/less-style/main.less
```diff
@@ -12,3 +12,7 @@
 @import "question-props";
 @import "structure";
 @import "tree-content";
+
+a {
```
proteusvacuum: I wasn't sure where to put this rule, since there doesn't seem to be any place that contains global styles.

orangejenny: Generic styles are in [/corehq/apps/hqwebapp/static/hqwebapp/less/_hq/](https://github.com/dimagi/commcare-hq/tree/master/corehq/apps/hqwebapp/static/hqwebapp/less/_hq/); I'd put this in typography.less.

proteusvacuum: https://github.com/dimagi/commcare-hq/pull/24141

PR #956, opened 2019-05-03
Gracefully load choice with missing value node

millerdev:
Not sure how [a form gets into this state](https://dimagi-dev.atlassian.net/browse/HI-596), but we can at least attempt to handle it gracefully when it happens, allowing the user to fix the error and move on.

@orangejenny @sravfeyn


tests/static/form/missing-value.xml
```diff
@@ -0,0 +1,32 @@
...
+					</text>
+				</translation>
+			</itext>
+		</model>
+	</h:head>
+	<h:body>
+		<select1 vellum:ref="#form/choice" ref="/data/choice">
+			<label ref="jr:itext('choice-label')" />
+			<item>
+				<label ref="jr:itext('choice-empty-label')" />
```
orangejenny: Just so I'm clear, the issue with this form is that this label doesn't have a sibling value node, right?

millerdev: correct

PR #958, opened 2019-08-06
Maintain xpath referencess on copy/paste

millerdev:
This fixes an unintuitive behavior when copying/pasting multiple questions that contain references to each other. I've wanted to fix this ever since [Jessica Long reported it almost two years ago](http://manage.dimagi.com/default.asp?265411), and finally decided to do it today.

🐡 Review by commit.

### Create form with question referencing another question
image

### Select and copy/paste `num` and `plus1` into group
image

@dimagi/product this makes a subtle change in behavior, and I think it's more intuitive and therefore better than the previous behavior. We should announce it. Do we need to do anything else?

### DIFF
--- a/src/copy-paste.js
+++ b/src/copy-paste.js
@@ -11,7 +11,7 @@ define([
     $,
     _,
     copy_paste_help,
-    mugs,
+    mugsModule,
     tsv,
     util,
     analytics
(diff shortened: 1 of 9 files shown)


jessica-muremure:
I would like to state, for the Github record, that I am incredibly excited about this feature! Thank you so much, Daniel!!!

PR #960, opened 2019-08-07
Updated languages menu to scroll vertically

orangejenny:
https://dimagi-dev.atlassian.net/browse/HI-813

Most of the changes are `height` => `max-height` so that scrolling content isn't forced to be as tall as the window.

Tested out on a form with a lot of languages and a form with a few languages. Tested out full screen mode.


orangejenny:
before
Screen Shot 2019-08-07 at 3 28 37 PM


after
Screen Shot 2019-08-07 at 10 53 12 AM

PR #964, opened 2019-10-03
Updated invalid XML error to be friendlier

orangejenny:
Noticed while working on https://github.com/dimagi/commcare-hq/pull/25531

The code is checking for an "Invalid XML" prefix but the message starts with "Error: Invalid XML" instead.

before
Screen Shot 2019-10-02 at 10 05 44 PM

after
Screen Shot 2019-10-02 at 10 04 45 PM

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1284,7 +1284,7 @@ define([
                 // hack: don't display the whole invalid XML block if it
                 // was a parse error
                 var msg = e.toString();
-                if (msg.indexOf("Invalid XML") === 0) {
+                if (msg.indexOf("Invalid XML") !== -1) {
                     msg = gettext("Parsing Error. Please check that your form is valid XML.");
                 }
 


orangejenny:
It'd be better if modal didn't close and/or it directed you to form settings, but the former is non-trivial and the latter breaks the boundaries of form builder as a standalone tool.

millerdev:
I've used that nasty XML source print-out in the past. Would be nice if the error dialog provided a link or some way to easily review and possibly edit the bad XML.

orangejenny:
@millerdev If it were thrown as a js error so you could get it from the console, would that work?

millerdev:
Heh, I suppose that would be better than nothing.

orangejenny:
@millerdev Cool, that already happens. (I didn't realize that when I commented earlier, thought my change might have affected what went to the console.) Let's leave it as is for now and consider adding a link or some other solution if the current state ends up being annoying.

PR #968, opened 2020-06-04
JS Library Upgrades

biyeun:


millerdev:
All tests passed locally, both on command line and in browser.

millerdev:
Note for future reference: as of this PR, 4 Travis tests are broken with no known fix. We should look into that sometime since it's not great to have failing tests on our shared test runner.

PR #970, opened 2020-06-24
Fix JS Tests

biyeun:
going to approach this by first getting travis to successfully run a known succeeding test. then I need to figure out how to handle the known errors and timeout issues with our tests...


start-server.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+http-server -p 8088 &
+sleep 10
```
millerdev, line 3: Why sleep after it stops?

package.json
```diff
@@ -22,18 +22,17 @@
...
-        "phantomjs-prebuilt": "^2.1.16",
+        "mocha-headless-chrome": "^3.1.0",
         "requirejs": "^2.3.6"
     },
     "main": "src/main.js",
     "scripts": {
         "install": "bower install",
-        "test": "phantomjs node_modules/mocha-phantomjs-core/mocha-phantomjs-core.js 'built.html' spec '{\"bail\":false}' && jshint src/*.js && jshint -c tests/.jshintrc tests/*.js"
+        "testserver": "`npm bin`/http-server -p 8088",
+        "test": "node_modules/mocha-headless-chrome/bin/start -f http://localhost:8088 -t 500000"
```
millerdev, line 36: I'm curious about `-t 500000`. I assume that's a timeout. Is that for each individual test, or the entire test run? Do any tests take anywhere near that long? Previously the timeout was something like 2s/test (with specific exceptions for known outliers), which was nice because it highlighted long-running tests quickly. I'd prefer to keep the per-test timeout low if possible.

biyeun: this is for the entire test run because it treats all of index.html as a single test.

src/xml.js
```diff
@@ -49,8 +49,7 @@ define([
                 .replace(/^<xml\b[^>]*>/, "")   // remove <xml ...>
                 .replace(/<\/xml>$/, "")        // remove </xml>
                 .replace(emptytag, "$1 />")     // <tag></tag> to <tag />
-                .replace(/"\/>/g, '" />')       // space before />
-                .replace(/&nbsp;|\xa0/g, " ");  // &nbsp; is not a valid XML entity
+                .replace(/"\/>/g, '" />');      // space before />
```
millerdev, line 52: I think this removed the wrong line. Should have removed line 99
https://github.com/dimagi/Vellum/blob/de78396615c7d32d2e2892f2329e27750c72ae27/src/xml.js#L99

tests/form.js
```diff
@@ -538,7 +538,7 @@ define([
                 var form = util.loadXML(""),
                     xml = util.parseXML(call('createXML'));
                 assert.equal(form.formName, "New Form");
-                assert.equal(xml.find("title").text(), "New Form", "title");
+                assert.equal($(xml[0].getElementsByTagName('h:title')).text(), "New Form", "title");
```
millerdev, line 541: would `xml.find("h:title")` work?

package.json
millerdev: Interesting. Are there still time limits on individual tests?

tests/form.js
biyeun: nope that didn't work

src/xml.js
biyeun: thanks! whoops 😳

package.json
millerdev: Previously `built.html` caused this to run against the built version of vellum

https://github.com/dimagi/Vellum/blob/31900bc23dd859e8eae784116971066808646b0a/built.html#L15

and `npm test` is was (and still is) run by Travis via `make test` after the build:

https://github.com/dimagi/Vellum/blob/31900bc23dd859e8eae784116971066808646b0a/Makefile#L37-L42

Does `useBuilt` still get set somewhere during the travis test run?

PR #973, opened 2020-07-06
Removed expanded audio from formbuilder

michelleliang1:
I removed all instances of expanded-audio in formbuilder, including tests.
When adding multimedia, users will no longer be able to select expanded-audio in the formbuilder.


michelleliang1:
@millerdev Thanks! Fixed!

PR #974, opened 2020-08-06
Upgrade JS dependancies to deal with dependabot alerts

biyeun:
addresses dependabot alerts for critical, high, and most moderate severity packages and sub dependencies


biyeun:
this technically shouldn't need QA as this affects dev dependencies and their sub dependencies. I verified that JS tests passed and that the build succeeds. but this shouldn't touch any of the production-level dependencies. @millerdev thoughts? should I put this thru QA just to be safe?

package.json
```diff
@@ -21,40 +21,49 @@
         "XMLWriter": "dimagi/XMLWriter#master",
         "bootstrap": "3.4.1",
         "codemirror": "5.30.0",
-        "fuse.js": "~2.7.4",
-        "jquery": "~3.5.1",
+        "fuse.js": "2.7.4",
+        "jquery": "3.5.1",
```
millerdev, line 25: Why switch to pinning to specific versions rather than using approximate version specifiers?

One disadvantage of this is that every point release must be explicitly updated rather than happening automatically on the next build. Is there a new workflow to work around that?

biyeun: hmm, i guess we can keep the approximate ones `~`. at least with `jquery` i wanted it to be exactly the same as HQ's unless both are explicitly updated. for `fuse.js` I don't have a great answer

millerdev: Pinning `jquery` to the same version used by HQ seems reasonable.

PR #975, opened 2020-08-18
fix truncated text issue with output refs

biyeun:
https://dimagi-dev.atlassian.net/browse/SAAS-11032

fixes an issue where if an ampersand and an output ref are present in an itext block, the text following the output ref gets truncated due to a strange behavior that this triggers in jquery's `parseXML`

### DIFF
--- a/tests/xml.js
+++ b/tests/xml.js
@@ -270,6 +270,14 @@ define([
                 value = parsedXml.find("value");
             eq(value, REGEXP_CRASHING_DEBUG_ITEXT_PARSED, false);
         });
+
+        it("should not wrap trailing text in output refs", function () {
+            var value = "<value>&amp; <output value='/path' /> " +
+                "text-after *should* remain. ~&amp; not #disrupt the\nflow" +
+                "</value>";
+            eq($(value),"& <output value=\"/path\" /> text-after " +
+                "*should* remain. ~& not #disrupt the\nflow");
+        });
     });
 
     describe("The XML query", function () {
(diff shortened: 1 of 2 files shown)


biyeun:
all tests pass locally

tests/xml.js
```diff
@@ -270,6 +270,14 @@ define([
                 value = parsedXml.find("value");
             eq(value, REGEXP_CRASHING_DEBUG_ITEXT_PARSED, false);
         });
+
+        it("should not wrap trailing text in output refs", function () {
+            var value = "<value>&amp; <output value='/path' /> " +
+                "text-after *should* remain. ~&amp; not #disrupt the\nflow" +
+                "</value>";
+            eq($(value),"& <output value=\"/path\" /> text-after " +
```
millerdev, line 278: formatting nit
```suggestion
            eq($(value), "& <output value=\"/path\" /> text-after " +
```

tests/xml.js
```diff
@@ -298,5 +306,17 @@ define([
         it("should round-trip newline", function () {
             eq(xml.query("\n").toString(), "\n");
         });
+
+        it("should not insert text inside output refs", function () {
+            eq(xml.query(
+                "& <output value=\"/data/question1\" " +
+                "vellum:value=\"#form/question1\"> " +
+                "text-after *should* remain. ~& not #disrupt the\nflow</output>"
```
millerdev, line 314: How important is this test? I'd prefer not to have this behavior enshrined in our test suite since it is not a valid XML transformation even though it is a quirk of jQuery 3.5.0. I would say the behavior is undefined if someone actually wrote a label expression like this in their form.

biyeun: so this test is there to guarantee we don't run into the same behavior as described in the ticket. the thing is jquery consistently parsed `& <output /> text` as `& <output> text</output>` so regardless of whether a user is doing this or not jquery definitely is always doing this :/

tests/xml.js
```diff
@@ -298,5 +306,17 @@ define([
         it("should round-trip newline", function () {
             eq(xml.query("\n").toString(), "\n");
         });
+
+        it("should not insert text inside output refs", function () {
```
millerdev, line 310: Can we at least add a comment that says something to the effect of
```suggestion
        it("should not insert text inside output refs", function () {
            // this test is verifying a quirk of jQuery 3.5.0 that results in
            // invalid XML transformations. It is not necessary to preserve
            // the behavior if/when jQuery no longer mangles XML this way.
```

biyeun:
thanks @millerdev sorry it's been a slow day

PR #977, opened 2020-09-10
Added support for Markdown Tables

AmitPhulera:
https://dimagi-dev.atlassian.net/browse/ICDS-1544

We now support Markdown Tables on Commcare Mobile https://github.com/dimagi/commcare-android/pull/2266
This change will add support for the same on HQ.

Image of Markdown Tables supported on Local HQ.

Screenshot 2020-09-10 at 1 31 28 PM


mkangia:
Super quick! :100:

orangejenny:
Usually app building features that require a minimum mobile version aren't available in HQ if your app isn't at that version, which is controlled by [feature_support.py](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/app_manager/feature_support.py). What would you think of adding an option like `markdownTables` to [_get_vellum_features](https://github.com/dimagi/commcare-hq/blob/346fba741ef756943873707fa7860595733ebdc0/corehq/apps/app_manager/views/formdesigner.py#L312) so that people on old CC versions don't get confused?

src/javaRosa/util.js
```diff
@@ -55,8 +55,9 @@ define([
          * headings
          * italics/bold/bold italics
          * links
+         * tables
          */
-        return /^\d+[\.\)] |^\* |~~.+~~|# |\*{1,3}\S.*\*{1,3}|\[.+\]\(\S+\)/m.test(val);
+        return /^\d+[\.\)] |^\* |~~.+~~|# |\*{1,3}\S.*\*{1,3}|\[.+\]\(\S+\)|^(\|[^\n]+\|\r?\n)((?:\|:?[-]+:?)+\|)(\n(?:\|[^\n]+\|\r?\n?)*)?$/m.test(val);
```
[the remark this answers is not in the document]
AmitPhulera: Thanks for the suggestion @orangejenny. It will definitely be more readable.

AmitPhulera:
> Is Web Apps support part of this ticket?

I think it is now and I am guessing that it would have to do something with Form player? Can you please point me to resources that will help me in achieving it?

> Usually app building features that require a minimum mobile version aren't available in HQ if your app isn't at that version, which is controlled by feature_support.py. What would you think of adding an option like markdownTables to _get_vellum_features so that people on old CC versions don't get confused?

I did not understand the entire workflow for it. How are we going to markdown tables from rendering for App versions smaller than a particular version? Are we going to add check in Vellum also for it?

orangejenny:
Web apps markdown is handled in HQ javascript, so there shouldn't be a need to touch the formplayer repository. The markdown renderer is set up [here](https://github.com/dimagi/commcare-hq/blob/d5e0ccaf9684e03ab3eeeedbe754340228ecca0d/corehq/apps/cloudcare/static/cloudcare/js/form_entry/fullform-ui.js#L8-L37). It doesn't use the same `zero` initialization as vellum, so maybe tables will just work, and it's just a matter of verifying that.

If you do need to make changes to web apps js, be aware that there's a [big open PR](https://github.com/dimagi/commcare-hq/pull/28443) that is likely to conflict, so it'd likely be a good idea to branch off of that PR. Feel free to offline with me about any of this.

orangejenny:
As far as the app versioning, I was thinking it could work to add a setting that HQ would set based on whether or not the app's CC version supports tables in markdown. The existing `markdown_in_groups` setting does this. Then in Vellum, the code you added to `markdown.js` and to `javaRosa/util.js` would both check that setting. If it *isn't* set, vellum would ignore tables.

The ideal thing would be to detect that someone is trying to use markdown and show them a message like, "Hey, upgrade to 2.50 to use tables!" (or whatever the right CC version is). We have a similar interaction for app settings, where you can see all app settings but any that you can't use are disabled and have a message telling you to upgrade if you want to use them.

orangejenny:
^ @millerdev I don't think vellum currently has any of that "upgrade to CommCare X.X to use this feature" messaging, does it? It doesn't look like `markdown_in_groups` does that kind of thing.

AmitPhulera:
> Web apps markdown is handled in HQ javascript, so there shouldn't be a need to touch the formplayer repository. The markdown renderer is set up here. It doesn't use the same zero initialization as vellum, so maybe tables will just work, and it's just a matter of verifying that.

You are right, the webapps already support the table markdown syntax. Only some CSS was needed to make it tidy.

> The ideal thing would be to detect that someone is trying to use markdown and show them a message like, "Hey, upgrade to 2.50 to use tables!" (or whatever the right CC version is). We have a similar interaction for app settings, where you can see all app settings but any that you can't use are disabled and have a message telling you to upgrade if you want to use them.

I looked around in vellum to check how we check after which the Commcare version a feature supported but did not get any success. Can you please point me to some examples?

millerdev:
> ^ @millerdev I don't think vellum currently has any of that "upgrade to CommCare X.X to use this feature" messaging, does it?

It does not that I know of.

> I looked around in vellum to check how we check after which the Commcare version a feature supported but did not get any success. Can you please point me to some examples?

Vellum does not check the CommCare version directly. It uses [feature flags](https://github.com/dimagi/Vellum/blob/51c5e800d79a2e7a356abd4f7addecd957259817/tests/options.js#L237-L251), which are [passed into vellum by HQ](https://github.com/dimagi/commcare-hq/blob/0d51eb6625a09c229ca5258bbec0ea0248fa327f/corehq/apps/app_manager/views/formdesigner.py#L146). Examples of how they are used:

https://github.com/dimagi/Vellum/blob/51c5e800d79a2e7a356abd4f7addecd957259817/src/javaRosa/itextBlock.js#L111

https://github.com/dimagi/Vellum/blob/51c5e800d79a2e7a356abd4f7addecd957259817/src/core.js#L711

https://github.com/dimagi/Vellum/blob/51c5e800d79a2e7a356abd4f7addecd957259817/src/mugs.js#L1035

orangejenny:
@millerdev @AmitPhulera  What would you think of extending the feature flag support so that instead of a boolean it could accept a message string that would indicate why the feature isn't available? The message would likely either be version-related or plan-related, so either "Upgrade to CommCare 2.50 to use this." or "Upgrade to an Advanced plan to use this."  For this feature, we could potentially use the existing warning messages infrastructure, displaying a warning under **Display Text** if someone attempts to use table syntax.

@dimagi/product has talked about adding this kind of messaging before, since one of the things HQ does not do well is show you when there's a feature that's relevant to what you're doing, but your plan doesn't include it.

 It's also fair for to just say that adding a custom message is out of scope for this PR, and the only required change is to not allow table markdown on older versions of CC.

millerdev:
>  What would you think of extending the feature flag support so that instead of a boolean it could accept a message string that would indicate why the feature isn't available?

One issue here is that a non-empty string has a truthy value when evaluated as a boolean, so if you checked `if (options.features.markdown_in_groups)` that would evaluate to true if `markdown_in_groups === "Upgrade to CommCare 2.50 to use this."` even though that would presumably mean the feature is not enabled. I think we will need to extend or change the feature flag system in vellum to store messages like this in another place. For example, a parallel object to `options.features` like `options.disabled_feature_reasons = {"markdown_in_groups": "Upgrade to CommCare 2.50 to use this.", ...}`. This would allow for confusing contradictions between the flag in `features` and values in `disabled_feature_reasons`, but we could do some validation in vellum (and in HQ tests) to ensure they are always consistent.

Secondly, feature flags are checked in many places, and sometimes those places are likely not conducive to displaying a UI message. Finding an appropriate place to display the message seems like something that will need to be done on a case by case basis for each flag.

orangejenny:
Hmm, yes, I think we'd want to support both boolean values and strings, and would need to check the type as well as the value, which isn't pretty. And the UI would indeed be case by case. I'd like to make it easier to add plan-based messages in the future...but I don't know when or if product will prioritize that, so there's no point in making the code more complicated now. Let's forget about this tangent. This PR can just limit tables to recent versions of CC.

AmitPhulera:
Thankyou @millerdev and @orangejenny for your feedback and suggestions on the task. I was able to add restrictions based on the app version on markdown tables. 
Please take a look when you have time.

src/javaRosa/util.js
```diff
@@ -47,16 +47,21 @@ define([
...
          * italics/bold/bold italics
          * links
+         * tables (regex written after ||)
          */
-        return /^\d+[\.\)] |^\* |~~.+~~|# |\*{1,3}\S.*\*{1,3}|\[.+\]\(\S+\)/m.test(val);
+        if (supportTables) {
+            return /^\d+[\.\)] |^\* |~~.+~~|# |\*{1,3}\S.*\*{1,3}|\[.+\]\(\S+\)/m.test(val) || /^(\|[^\n]+\|\r?\n)((?:\|:?[-]+:?)+\|)(\n(?:\|[^\n]+\|\r?\n?)*)?$/m.test(val);
+        } else {
+            return /^\d+[\.\)] |^\* |~~.+~~|# |\*{1,3}\S.*\*{1,3}|\[.+\]\(\S+\)/m.test(val);
+        }
```
millerdev, line 64: The duplication could be removed here by using variables.
```suggestion
        var hasMarkdown = /^\d+[\.\)] |^\* |~~.+~~|# |\*{1,3}\S.*\*{1,3}|\[.+\]\(\S+\)/m.test(val),
            hasTable = false;
        if (supportTables) {
            hasTable = /^(\|[^\n]+\|\r?\n)((?:\|:?[-]+:?)+\|)(\n(?:\|[^\n]+\|\r?\n?)*)?$/m.test(val);
        }
        return hasMarkdown || hasTable;
```

millerdev (commented):
Looking good.

src/javaRosa/util.js
AmitPhulera: Sweet 😄

millerdev (approved):

src/javaRosa/util.js
```diff
@@ -47,16 +47,22 @@ define([
...
          * headings
          * italics/bold/bold italics
          * links
+         * tables (regex written after ||)
          */
-        return /^\d+[\.\)] |^\* |~~.+~~|# |\*{1,3}\S.*\*{1,3}|\[.+\]\(\S+\)/m.test(val);
+        var hasMarkdown = /^\d+[\.\)] |^\* |~~.+~~|# |\*{1,3}\S.*\*{1,3}|\[.+\]\(\S+\)/m.test(val),
+            hasTable = false;
+        if (supportTables) {
+            hasTable = /^(\|[^\n]+\|\r?\n)((?:\|:?[-]+:?)+\|)(\n(?:\|[^\n]+\|\r?\n?)*)?$/m.test(val);
```
orangejenny, line 63: @AmitPhulera This doesn't pick up tables that aren't on the first line, like this one:
```
Look at my table
| col1      | col2 |
| ----------- | ----------- |
| data1      | data2       |
```
Could you update this?

AmitPhulera: Thanks, @orangejenny for the feedback.

```
| col1      | col2 |
| ----------- | ----------- |
| data1      | data2       |
```

This syntax is not supported yet, I just tried to support the format that was there in the Mobile PR, but it is fairly easy to do. And I have not thought about the text in the beginning though 🤔 I will make the required changes.

orangejenny: That syntax seems to work: https://www.commcarehq.org/a/bosco/apps/view/b7d394e5700f41d994335e585829891d/form/cb777bf09e38efb22975e46f4c9bc0d0511f4b37/source/#form/question1

AmitPhulera: This one was a tricky one to figure out. So what's actually happening is that if you will paste 
```
| Column1      | Column2 |
| ----------- | ----------- |
| 100      | 200     |
| 110   | 215        |
```
to an input field which is already configured to be a markdown field then it will render a table because it's a valid table format and markdown-it render it as a table which is probably what happened in your case.

However, if you paste it directly to a new question or to an input field which is not a markdown field [then it will not render a table](https://www.commcarehq.org/a/bosco/apps/view/b7d394e5700f41d994335e585829891d/form/cb777bf09e38efb22975e46f4c9bc0d0511f4b37/source/#form/question_2) because the regex that I have used is not written to catch the spaces after `|` in the lines below the headers. 


And the formats that will work presently are 
```
| col1    | col2    | col3 |
|----|----|----|
| r1c1 | r1c2 | r1c3 |
| r2c1 | r2c2 | r2c3 |
```
```
| col1    | col2    | col3 |
|:----:|:----:|:----:|
| r1c1 | r1c2 | r1c3 |
| r2c1 | r2c2 | r2c3 |
```

I am updating the regex to support the third one too. And it actually [supports text with table,](https://www.commcarehq.org/a/bosco/apps/view/b7d394e5700f41d994335e585829891d/form/cb777bf09e38efb22975e46f4c9bc0d0511f4b37/source/#form/question_3) as the regex matches the table and text, is ignored.

@orangejenny  I have added two questions in the form, I am not sure if we should remove them afterward. Let me know if we have to remove them.

AmitPhulera: Raised the PR https://github.com/dimagi/Vellum/pull/978

orangejenny: Ah, I see. Thank you for the thorough explanation 🙂 

And no worries about changing the form, it's fine to play around with that one.

PR #980, opened 2020-11-03
Implement ability to make all questions: 1. load case properties as default values 2. required

biyeun:
https://dimagi-dev.atlassian.net/browse/SAAS-11387

#### User-facing changes

Adds one option in registration/survey forms to mark all questions required:
Screen Shot 2020-11-03 at 2 27 00 PM

adds an additional option in followup forms (or forms where the datasource panel is visible) to set default values of all matching properties to the existing matching case value:
Screen Shot 2020-11-03 at 2 28 35 PM

confirmation dialog for marking all questions as required:
Screen Shot 2020-11-03 at 2 29 18 PM

confirmation dialog for setting default values:
Screen Shot 2020-11-03 at 2 29 46 PM

#### Risks
Definitely pushing this through QA.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -462,9 +462,12 @@ define([
             caseProperties = _this.datasources.getHashtagMap({});
         _this.data.core.form.walkMugs(function (mug) {
             if (mug.isVisible('defaultValue')) {
-                var caseProp = '#case/' + mug.p.nodeID;
+                var caseProp = '#case/' + mug.p.nodeID,
+                    userCaseProp = '#user/' + mug.p.nodeID;
                 if (caseProperties.hasOwnProperty(caseProp)) {
                     mug.p.defaultValue = caseProp;
+                } else if (caseProperties.hasOwnProperty(userCaseProp)) {
+                    mug.p.defaultValue = userCaseProp;
                 }
             }
         });
--- a/src/core.js
+++ b/src/core.js
@@ -247,7 +247,9 @@ define([
             // Allow onReady to access vellum instance (mostly for tests)
             _this.opts().core.onReady.apply(_this);
         }
-        this._init_bulk_update_questions();
+        if (_this.opts().features.allow_bulk_form_actions) {
+            this._init_bulk_update_questions();
+        }
         this._init_extra_tools();
         parser.init(this);
         this.loadXFormOrError(this.opts().core.form, function () {
--- a/src/core.js
+++ b/src/core.js
@@ -457,11 +457,11 @@ define([
 
     fn.defaultMatchingQuestionsToCaseProperties = function () {
         var _this = this,
-            caseProperties = _.keys(_this.datasources.getHashtagMap());
+            caseProperties = _this.datasources.getHashtagMap({});
         _this.data.core.form.walkMugs(function (mug) {
             if (mug.isVisible('defaultValue')) {
                 var caseProp = '#case/' + mug.p.nodeID;
-                if (caseProperties.indexOf(caseProp) !== -1) {
+                if (caseProperties.hasOwnProperty(caseProp)) {
                     mug.p.defaultValue = caseProp;
                 }
             }
--- a/src/core.js
+++ b/src/core.js
@@ -415,19 +415,22 @@ define([
     };
 
     fn.getBulkUpdateMenuItems = function () {
-        var _this = this;
-        return [
-            {
-                name: gettext('Make Required'),
-                confirmMessage: gettext("You are about to make all existing " +
-                    "questions on this form required. This action will overwrite " +
-                    "the current configuration of this form and can only be " +
-                    "undone by editing individual questions."),
-                confirmAction: function () {
-                    // todo
-                }
-            },
-            {
+        var _this = this,
+            menuItems =  [
+                {
+                    name: gettext('Make Required'),
+                    confirmMessage: gettext("You are about to make all existing " +
+                        "questions on this form required. This action will overwrite " +
+                        "the current configuration of this form and can only be " +
+                        "undone by editing individual questions."),
+                    confirmAction: function () {
+                        _this.makeAllQuestionsRequired();
+                    },
+                },
+            ];
+
+        if (_this.data.core.databrowser) {
+            menuItems.push({
                 name: gettext('Load Default Values'),
                 confirmMessage: gettext("You are about to set matching case " +
                     "properties as Default Values for all existing questions. " +
@@ -435,10 +438,35 @@ define([
                     "all questions with matching case properties and can only " +
                     "be undone by editing individual questions."),
                 confirmAction: function () {
-                    // todo
+                    _this.defaultMatchingQuestionsToCaseProperties();
                 },
-            },
-        ];
+            });
+        }
+        return menuItems;
+    };
+
+    fn.makeAllQuestionsRequired = function () {
+        var _this = this;
+        _this.data.core.form.walkMugs(function (mug) {
+            if (mug.spec.requiredAttr && mug.isVisible('requiredAttr')) {
+                mug.p.requiredAttr = true;
+            }
+        });
+        _this.refreshCurrentMug();
+    };
+
+    fn.defaultMatchingQuestionsToCaseProperties = function () {
+        var _this = this,
+            caseProperties = _.keys(_this.datasources.getHashtagMap());
+        _this.data.core.form.walkMugs(function (mug) {
+            if (mug.isVisible('defaultValue')) {
+                var caseProp = '#case/' + mug.p.nodeID;
+                if (caseProperties.indexOf(caseProp) !== -1) {
+                    mug.p.defaultValue = caseProp;
+                }
+            }
+        });
+        _this.refreshCurrentMug();
     };
 
     fn.showConfirmBulkActionModal = function (confirmMessage, confirmActionFn) {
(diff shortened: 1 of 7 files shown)


src/core.js
```diff
@@ -392,6 +393,99 @@ define([
         ];
     };
 
+    fn._init_bulk_update_questions = function () {
```
millerdev, line 396: I know we discussed this earlier, but now that I'm seeing it in code, it does seem it may be nicer to implement this as a new plugin rather than adding another ~100 lines to core.js (which is already way too big).

src/core.js
```diff
@@ -392,6 +393,99 @@ define([
...
+
+    fn.getBulkUpdateMenuItems = function () {
+        var _this = this,
+            menuItems =  [
+                {
+                    name: gettext('Make Required'),
+                    confirmMessage: gettext("You are about to make all existing " +
+                        "questions on this form required. This action will overwrite " +
+                        "the current configuration of this form and can only be " +
+                        "undone by editing individual questions."),
```
millerdev, line 425: Would it be hard to integrate these actions with the undo manager? It would make a far better user experience if they could be undone like other modifications (and could omit the modal prompt).

src/core.js
```diff
@@ -392,6 +393,99 @@ define([
...
+                },
+            });
+        }
+        return menuItems;
+    };
+
+    fn.makeAllQuestionsRequired = function () {
+        var _this = this;
+        _this.data.core.form.walkMugs(function (mug) {
+            if (mug.spec.requiredAttr && mug.isVisible('requiredAttr')) {
```
millerdev, line 451: It probably makes sense to move the `mug.spec.requiredAttr` (more generally `mug.spec[property]`) check into `mug.isVisible()`. I'm surprised it doesn't already check for that, but it should be fine to make it return false immediately if there is no spec for the given property.

src/core.js
```diff
@@ -392,6 +393,99 @@ define([
...
+        _this.refreshCurrentMug();
+    };
+
+    fn.defaultMatchingQuestionsToCaseProperties = function () {
+        var _this = this,
+            caseProperties = _.keys(_this.datasources.getHashtagMap());
+        _this.data.core.form.walkMugs(function (mug) {
+            if (mug.isVisible('defaultValue')) {
+                var caseProp = '#case/' + mug.p.nodeID;
+                if (caseProperties.indexOf(caseProp) !== -1) {
```
millerdev, line 464: `indexOf()` is O(n), while [`hasOwnProperty()` is many times faster if not O(1)](https://stackoverflow.com/a/55057332/10840)
```suggestion
            caseProperties = _this.datasources.getHashtagMap({});
        _this.data.core.form.walkMugs(function (mug) {
            if (mug.isVisible('defaultValue')) {
                var caseProp = '#case/' + mug.p.nodeID;
                if (caseProperties.hasOwnProperty(caseProp)) {
```

millerdev (commented):
I realize this is design feedback, and code reviews are not the best place to provide that. Forgive me, I was not included in any design discussions (although I did volunteer to work with Dev on this feature), so I did not have a chance to provide it earlier.

What do you think of putting these new actions in the hamburger menu on the right beside the delete button? :arrow_down:

image

They would only be available when multiple questions are selected. That would enable more powerful/flexible workflows because it would allow executing the actions on a (selected) subset of questions rather than all questions in the form.

orangejenny:
> What do you think of putting these new actions in the hamburger menu on the right beside the delete button? ... They would only be available when multiple questions are selected. That would enable more powerful/flexible workflows because it would allow executing the actions on a (selected) subset of questions rather than all questions in the form.

This is interesting, I like this idea. It also sidesteps the problem of users potentially being confused about whether selecting the new item will affect future questions or not.

millerdev:
Forgot to mention in my first review, this PR needs tests.

biyeun:
hey, so this:

> What do you think of putting these new actions in the hamburger menu on the right beside the delete button?

feels like it will be a heavier lift. but I'm also not as familiar with Vellum as you two are. what are your thoughts on the challenges here @orangejenny and @millerdev ?

can we continue according to the product spec for now in an effort to get this done? Feels like we spent a fair amount of time muddling over the spec, but now that the work has been done making a spec change request this late in the game when there was a lot of time to comment on the ticket before seems a little backward to me.

yes, I was already planning to add tests after an initial code review and this goes into QA (I mentioned this in the ticket). I guess I'll work on a plugin structure now too which will further delay progress on this and then add some tests. But before I work on that I want to understand the thoughts on the spec change here.

Also, working with the undo feature is not familiar either. can we make that happen as a follow-up to this ticket? I assume Dev has a good reason for pushing on this. If he doesn't then sure I can spend even more time on this and change my estimates.

src/core.js
biyeun: yeah...this is why i had asked right before starting this work...but i guess it's not much more of an extra lift. It will certainly delay things going into QA though vs having done it as a plugin from the start. FYI Dev

src/core.js
biyeun: I'm not super familiar with the undo manager. so from my perspective that seems like a huge ask given my uncertainty level and should be considered as a separate ask from product unless Dev is ok with an additional delay.

orangejenny:
@biyeun @millerdev Can we discuss timelines in the ticket, with Dev? I agree this isn't a good time to be getting UX feedback, but there's also risk to deploying a feature and potentially making non-trivial UX changes shortly afterwards.

biyeun:
@millerdev besides the argument for making `core.js` 100 lines shorter, what is the advantage of having this tiny bit of code existing in a separate plugin? Separating 100 lines of code (where the bulk of those lines is due to nicely formatted human-readable text) feels like it adds more complexity and difficulty understanding when the rest of the menu items all follow completely different paradigms. it also feels a bit frustrating to have asked you this specific question before and now I'm getting a different answer.  :/  Also, I feel this is a part of core.js as it's always going to be required by any vellum page (whether it's a survey or followup form)

djmore9:
> I realize this is design feedback, and code reviews are not the best place to provide that. Forgive me, I was not included in any design discussions (although I did volunteer to work with Dev on this feature), so I did not have a chance to provide it earlier.
> 
> What do you think of putting these new actions in the hamburger menu on the right beside the delete button? ⬇️
> 
> image
> 
> They would only be available when multiple questions are selected. That would enable more powerful/flexible workflows because it would allow executing the actions on a (selected) subset of questions rather than all questions in the form.





> I realize this is design feedback, and code reviews are not the best place to provide that. Forgive me, I was not included in any design discussions (although I did volunteer to work with Dev on this feature), so I did not have a chance to provide it earlier.
> 
> What do you think of putting these new actions in the hamburger menu on the right beside the delete button? ⬇️
> 
> image
> 
> They would only be available when multiple questions are selected. That would enable more powerful/flexible workflows because it would allow executing the actions on a (selected) subset of questions rather than all questions in the form.


@millerdev 

Thank you for this feedback. Just so you know, I ran the spec by a few power users, to get their thoughts/feedback. I shared that feedback with @biyeun and @orangejenny, and I am happy to share that with you as well. 

We are squarely trying to solve the the problem of applying these 2 actions to **ALL** questions in a quick and convenient way, and I think we want to optimize our design for that. Although, I do think what you are suggesting could have it's own separate utility. 

That being said, we will definitely get feedback on application of these actions for selected questions in the UAT.

src/core.js
djmore9: @millerdev What is the fundamental concern associated with the current extra 100 lines of code to core.js? For example, does it negatively affect code maintainability? or are there other concerns like performance issues.

src/core.js
djmore9: I do like this idea, but note that currently we present the Undo option **only** on question deletions, so we will introducing a new UX pattern if we support Undo for form updates that not question deletions. Worth considering if this could have any drawbacks. Would love to hear your thoughts on this @millerdev, and also @orangejenny and @biyeun.

src/core.js
millerdev: > does it negatively affect code maintainability

Yes, this. The file is already huge, and adding more stuff in here makes it harder to navigate and find things within the file.

src/core.js
millerdev: Fair enough. This could be spun off as a separate effort too if we wanted to ship the feature and then improve it later.

src/core.js
djmore9: Got it. thanks.

src/core.js
djmore9: We will 100% address this during UAT.

djmore9:
@Ben-Talbot 

@millerdev brought up 2 important items ( Hamburger/multi-select idea, and Undo action) that we should de-risk during the UAT.

src/core.js
millerdev: Also, I think it would be fine to do this as a follow-up PR or while QA is in progress.

src/core.js
biyeun: hey, so I looked into this, and I think I don't want to do that because `isVisible` first checks `getPresence` which explicitly throws an error when the property is not present:
https://github.com/dimagi/Vellum/blob/4557be6d99e1bb5c081a83a0f57bf3bd077cbebb/src/mugs.js#L329-L331

I want this change to be as isolated as possible so modifying existing code when there is already a check in place for the opposite seems like a bad idea.

biyeun:
an update here:

1. put the current state of this feature behind a feature flag
2. added tests

I still need to respond to QA / UAT tickets.

millerdev (approved):
Thanks for writing the tests!

Approving phase one, with understanding that we will rework this before it goes to GA.

biyeun:
QA passed! 🎉

PR #983, opened 2020-12-17
Hashtags should now use the provided alias name for a property rather…

mjriley:
… than its id

Created because some properties we support are attributes -- '@property' rather than 'property'. However, attributes displayed with an @ sign break hashtags. The correct support is for them to be displayed with an alternate name property, while retaining the '@attribute' notation for its xpath.

Given that these changes pass all tests, I do not believe QA is required.


millerdev (approved):
Thanks!

PR #985, opened 2020-12-22
Fix - "Security Error" Dialogue Box in Firefox

AmitPhulera:
Ticket Link - https://dimagi-dev.atlassian.net/browse/SAAS-11597

### Description

Whenever the page was getting loaded for the first time `window.replaceState` was getting called for every question in the form. This was causing a Security Error dialogue box to appear in Firefox.

The fix that we made calls the browser API only when it is required i.e whenever question_id is modified or changes in the forms are saved.

### DIFF
--- a/src/widgets.js
+++ b/src/widgets.js
@@ -350,8 +350,13 @@ define([
             super_updateValue();
         };
 
+        widget._setURLHash = _.debounce(function(){
+                mug.form.vellum._setURLHash(mug)
+            }, 500);
+        
         widget.handleChange = function () {
             super_handleChange();
+            widget._setURLHash(mug);            
             if (!mug.p.nodeID) {
                 // HACK drop errors; nodeID will be set automatically on save
                 mug.dropMessage("nodeID", "mug-nodeID-error");
--- a/src/widgets.js
+++ b/src/widgets.js
@@ -350,7 +350,7 @@ define([
             super_updateValue();
         };
 
-        widget._setURLHash = _.debounce(function(){
+        widget._setURLHash = _.debounce(function(mug){
                 mug.form.vellum._setURLHash(mug)
             }, 500);
         
--- a/tests/urlHash.js
+++ b/tests/urlHash.js
@@ -29,13 +29,27 @@ requirejs([
             assert.equal(mug.absolutePath, "/data/group");
         });
 
-        it("should change the hash when you add a question", function(done) {
+        it("should change the hash when you add a question", function() {
             util.loadXML("");
-            util.addQuestion("Text", "text");
-            util.saveAndReload(function(){
-                assert.equal(window.location.hash, '#form/text');
-                done();
-            })
+            util.addQuestion("Text", "initial");
+            util.clickQuestion("initial");
+            assert.equal(window.location.hash, "#form/initial");
+        });
+
+        it("should change the hash when you delete a question", function() {
+            util.loadXML("");
+            
+            util.addQuestion("Text", "first");
+            util.clickQuestion("first");
+
+            util.addQuestion("Text", "second");
+            util.clickQuestion("second");
+            
+            assert.equal(window.location.hash, "#form/second");
+            
+            util.deleteQuestion("/data/second");
+            
+            assert.equal(window.location.hash, "#form/first");
         });
 
         it("should change the hash when you delete a question", function(done) {
--- a/tests/urlHash.js
+++ b/tests/urlHash.js
@@ -52,18 +52,26 @@ requirejs([
             assert.equal(window.location.hash, "#form/first");
         });
 
-        it("should change the hash when you delete a question", function(done) {
+        it("should call setURLHash only once when multiple questions are loaded", function(done){
+            
+            var count = 0;
             util.loadXML("");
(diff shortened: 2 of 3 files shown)


proteusvacuum (approved):
Do you think this needs to go through QA?

AmitPhulera:
@proteusvacuum thanks for your review.  
Regarding QA, I guess there are pretty good tests to test for the changes that I have made. So according to my understanding, we should be good here. But if you have any recommend any specific thing that can be QAed I would be happy to send them out to QA.

src/widgets.js
```diff
@@ -350,8 +350,13 @@ define([
             super_updateValue();
         };
 
+        widget._setURLHash = _.debounce(function(){
```
millerdev, line 353: The mug should be passed in here, right?
```suggestion
        widget._setURLHash = _.debounce(function (mug) {
```

tests/urlHash.js
```diff
@@ -29,19 +29,28 @@ requirejs([
...
 
-        it("should change the hash when you add a question", function() {
+        it("should change the hash when you add a question", function(done) {
             util.loadXML("");
             util.addQuestion("Text", "text");
-            assert.equal(window.location.hash, '#form/text');
+            util.saveAndReload(function(){
+                assert.equal(window.location.hash, '#form/text');
+                done();
+            })
```
millerdev, line 38: Why did these tests need to change? Seems like there might be a change in functionality. Is it now necessary to save and reload the form before the URL is updated with new hash?

millerdev (commented):
Would be nice if a test could be added to verify that `vellum._setURLHash()` is only called once when a form with multiple questions is loaded.

src/widgets.js
AmitPhulera: Yep. Yep. Thanks for pointing out, it was an oversight. I am wondering how it is was working then 🤔

tests/urlHash.js
AmitPhulera: The place where we have added handlers to `setURLHash` is when a `question id` is modified or when a question is saved. But in the [tests utils, addQuestion method](https://github.com/dimagi/Vellum/blob/master/tests/utils.js#L479) does not modify the question id of the question so the trigger isn't pulled. But on FormBuilder the question id is modified, so the trigger is pulled.
So for the tests to work fine I had to save the questions.

src/widgets.js
millerdev: We should either make this change or not pass `mug` below in `widget._setURLHash(mug)`

tests/urlHash.js
millerdev: Are you saying that `util.addQuestion("Text", "text")` does not trigger the `question id` modified event? It seems better to trigger that event than using `util.saveAndReload` in these tests. What do you think?

It should be possible to trigger a question id change by changing the question id explicitly after creating the question. Something like
```js
            var text = util.addQuestion("Text", "initial");
            text.p.nodeID = "text";
            assert.equal(window.location.hash, '#form/text');
```

src/widgets.js
AmitPhulera: Agreed. We should pass `mug` as you have suggested.

tests/urlHash.js
AmitPhulera: Manually triggering the event seems much better. I will try if that works.

millerdev: Oh, right. Since this is testing functionality that is implemented in the UI layer (widgets.js), it should trigger the widget change event. Here's an example:
```js
var text = util.addQuestion("Text", "initial");
util.clickQuestion("initial");
$("[name=property-nodeID]").val("text").change();
assert.equal(window.location.hash, '#form/text');
```

AmitPhulera: I realized that we can just update the hash by clicking on the question, so just `util.clickQuestion` worked out fine.

tests/urlHash.js
```diff
@@ -31,17 +31,40 @@ requirejs([
...
+            util.clickQuestion("second");
+            
+            assert.equal(window.location.hash, "#form/second");
+            
+            util.deleteQuestion("/data/second");
+            
+            assert.equal(window.location.hash, "#form/first");
+        });
+
+        it("should change the hash when you delete a question", function(done) {
```
millerdev, line 55: Did you intend to add a new test with the same name/description as the one above it?

src/widgets.js
```diff
@@ -350,8 +350,13 @@ define([
             super_updateValue();
         };
 
+        widget._setURLHash = _.debounce(function(mug){
+                mug.form.vellum._setURLHash(mug)
+            }, 500);
```
millerdev, line 355: Nit: non-standard indentation
```suggestion
            mug.form.vellum._setURLHash(mug)
        }, 500);
```

millerdev (commented):
I don't see the test to verify that `vellum._setURLHash()` is only called once when a form with multiple questions is loaded. Is that still in the works?

millerdev (approved):

AmitPhulera:
Merged the changes into HQ. Should be live with next changes.
FYI @millerdev

PR #994, opened 2021-01-07
Ap/web worker test

AmitPhulera:
## Safety Assurance

- [ ] Risk label is set correctly
- [ ] The set of people pinged as reviewers is appropriate for the level of risk of the change
- [ ] If QA is part of the safety story, the "Awaiting QA" label is used
- [ ] I am certain that this PR will not introduce a regression for the reasons below

### Rollback instructions

- [ ] This PR can be reverted after deploy with no further considerations

### DIFF
--- a/createXform.js
+++ b/createXform.js
@@ -1,0 +1,7 @@
+onmessage = function(e) {
+    console.log('Worker: Message received from main script');
+    console.log(e.data)
+    postMessage("It works")
+    
+    
+  }
--- a/createXform.js
+++ b/createXform.js
@@ -24,7 +24,7 @@ onmessage = function(e) {
     for (var i = 1; i < meta.length; i++) {
       _writeInstance(xmlWriter, meta[i]);
     }
-    createBindList(dataTree, xmlWriter);
+    //createBindList(dataTree, xmlWriter);
         
     postMessage("work");
     
--- a/createXform.js
+++ b/createXform.js
@@ -19,12 +19,12 @@ onmessage = function(e) {
     
     xmlWriter.writeStartElement('model');
     xmlWriter.writeStartElement('instance');
-    _writeInstanceAttributes(xmlWriter, meta);
+    _writeInstanceAttributes(xmlWriter, meta[0]);
 
-    for (var i = 1; i < form.instanceMetadata.length; i++) {
-      _writeInstance(xmlWriter, meta);
+    for (var i = 1; i < meta.length; i++) {
+      _writeInstance(xmlWriter, meta[i]);
     }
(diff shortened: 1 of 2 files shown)


createXform.js
```diff
@@ -0,0 +1,231 @@
+importScripts('node_modules/XMLWriter/XMLWriter.js')
```
AmitPhulera, line 1: This is the script that should have the functionality to create XML

src/writer.js
```diff
@@ -10,64 +10,86 @@ define([
...
-        xmlWriter.writeEndElement(); //CLOSE MAIN INSTANCE
-        
-        // other instances
-        for (var i = 1; i < form.instanceMetadata.length; i++) {
-            _writeInstance(xmlWriter, form.instanceMetadata[i]);
+        if(window.Worker){
+            try {
+                var xformWorker=new Worker('createXForm.js')
+                var dataTree = form.dataTree();
+                var dt = JSON.parse(JSON.stringify(dataTree))
```
AmitPhulera, line 17: `dataTree` is circular JSON. We are not able to serialize it.

src/writer.js
```diff
@@ -10,64 +10,86 @@ define([
     util
 ) {
     var createXForm = function (form) {
```
AmitPhulera, line 12: I have commented out the implementation of `createXForm` and tried to move the functionality to `createXform.js`

AmitPhulera:
Hey @millerdev,
I have created this draft PR and have added the comments. 
Some of the requirements that we need to keep in mind for worker threads to work are- 
- The code should be present in a separate file which you need to run in the thread.
- You are not allowed to use any DOM manipulation API in the worker thread.
- The data that you pass is serialized and the passed which mean we can't send complex objects with functions.

With these limitations, my approach was to move the functions that are used in `createXForm` to `createXForm.js` and pass only the required parameters to `onMessage` handler(`onMessage` is a predefined handler name that is executed in a worker file). 
https://github.com/dimagi/Vellum/pull/994/files#diff-296289903bc1453dbcb97c62c40d4dd232be9f28e6dff21f4fb65fa50565b2cfR26

I commented out original implementation of `createXForm` and moved it step by step to `createXForm.js` file. I was stuck when `dataTree` had to be passed to createXForms.js as it was a circular JSON and I did not know what attributes should we send to the worker file.

I need suggestions on how can we move/serialized `dataTree` to our worker file `createXForm.js`

AmitPhulera:
Closing this out because it was too much of an effort to bring it to a state where we can run comparisons with the current version.

PR #997, opened 2021-01-22
Upgrade less

millerdev:
Upgrade less to 3.10.3, the same as HQ

Ran tests locally, both headless and in browser.

:tropical_fish: Review by commit. Some of the earlier commits are cleanup for the fact that jshint was removed from travis runs a while ago. I also run the test server on a different port, so parameterized that as well.

## Safety Assurance

- [x] I am certain that this PR will not introduce a regression for the reasons below

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,14 +2,10 @@ language: node_js
 sudo: false
 
 node_js:
- - stable
-
-before_install:
-    - chmod +x start-server.sh
+- stable
 
 before_script:
-    - ./start-server.sh
+- npm run testserver &
 
 script:
- - make test
-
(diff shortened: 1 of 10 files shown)


biyeun (approved):
🎉 yess! looks great

millerdev:
Note: the description on this PR is wrong:

> Upgrade less to 3.10.3, the same as HQ

I did a bunch more iterations after I wrote that, and ended up with [less 2](https://github.com/dimagi/Vellum/pull/997/commits/036f853a5e45e9e5627ff3c8bf7c640fc946540b), the same as used by the new version of require-less.

PR #1001, opened 2021-03-16
Upgrade js-xpath to 0.0.4

biyeun:
## Summary
We recently upgraded most of the dev dependencies for `dimagi/js-xpath` due to security vulnerabilities. This bumps the version referenced in vellum to that new release/tag in our repo

## Safety Assurance

### Automated test coverage
All tests pass locally!

### QA Plan
I think our tests are reliable enough that we would have caught any major issues, both in the xpath tests and in vellum specific tests.

### Safety story
Tests should cover any issues.

### Rollback instructions

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
         "requirejs-text": "2.0.15",
         "requirejs-undertemplate": "0.0.4",
         "underscore": "1.8.3",
-        "xpath": "dimagi/js-xpath#v0.0.3",
+        "xpath": "dimagi/js-xpath#v0.0.4",
         "xrayquire": "npm:xrayquire-for-npm#^0.1.1"
     },
     "devDependencies": {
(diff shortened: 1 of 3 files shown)


millerdev (approved):

millerdev:
Nice!

millerdev:
@biyeun are you planning to merge this?

biyeun:
@millerdev thanks for the reminder. I didn't see many complaints on staging except that Ben had some issues where builds wouldn't succeed at times (but eventually did after x number of attempts). I didn't have a moment yet to dig in. Do you think `js-xpath` would affect builds in such a sporadic way, or if there was a failure it would be pretty obvious in the tests?

millerdev:
I would expect any issues with js-xpath to be obvious, but might be worth taking a look at the details to see if they point at this.

biyeun:
@millerdev yeah there's definitely no issue with builds for this. since this has been on staging for a while now with no other weird formbuilding complaints and that tests have all passed...i think it's g2g

PR #1010, opened 2021-05-19
Fix equivalent-xml version

millerdev:
Also fix benign typo in index.html


millerdev:
Self-merging since this is a minor tweak and I'd like to test it on staging. Will get a proper peer review before this gets merged into HQ master.

skodde:
:+1:

PR #1019, opened 2021-07-09
Fix ampersand question ID bug

minhaminha:
## Summary

Fixes: Ampersand in question ID bug

Jira: [SAAS-10981](https://dimagi-dev.atlassian.net/browse/SAAS-10981)

## Product Description

Correctly represents question IDs with '&' characters

## Safety Assurance

### Automated test coverage

New tests added that confirm the existing bug and confirms the fix. 

### QA Plan

QA Passed:
https://dimagi-dev.atlassian.net/browse/QA-3281

### Safety story

Added new tests and confident the fix is safe (existing tests pass). 

### Rollback instructions

### DIFF
--- a/src/xml.js
+++ b/src/xml.js
@@ -126,25 +126,11 @@ define([
             return value; // value contains no character entity references or empty tags.
         }
         var xml = normalize(value, inner),
-            //character entity reference matching regex
-            ltref = /(?:&lt;([^;&\s]+(&gt;)|.|$))/g,
-            gtref = /(?:(&lt;[^;&\s]+|\s|)&gt;)/g,
-            ampref = /(?:&amp;([^&;\s]+;|.|$))/g;
+            refs = /(?:&lt;(=?\s)|(\s)&gt;|&amp;([^&;\s]+;|.|$))/g;
 
-        xml = xml.replace(ltref, function(match, lt){
-            return (lt.slice(lt.length - 3) == "gt;" ? match : ("<" + lt));
+        return xml.replace(refs, function (match, lt, gt, amp) {
+            return lt ? ("<" + lt) : (gt ? (gt + ">") : amp.slice(amp.length-1) === ";" ? match : ("&" + amp));
         });
-
-        xml = xml.replace(gtref, function(match, gt){
-            return (gt.slice(0, 3) == "&lt" ? match : (gt + ">"));;
-        });
-
-        xml = xml.replace(ampref, function(match, amp) {
-            return (amp.slice(amp.length-1) == ";" ? match : ("&" + amp));
-        });
-
-        return xml
-        //<<< and &&& can't appear unescaped since one captures the next's '&'
     }
 
     /**
--- a/src/xml.js
+++ b/src/xml.js
@@ -126,10 +126,25 @@ define([
             return value; // value contains no character entity references or empty tags.
         }
         var xml = normalize(value, inner),
-            refs = /(?:&lt;(=?\s)|(\s)&gt;|&amp;)/g;
-        return xml.replace(refs, function (match, lt, gt, amp) {
-            return lt ? ("<" + lt) : (gt ? (gt + ">") : ("&"));
+            //character entity reference matching regex
+            ltref = /(?:&lt;([^;&\s]+(&gt;)|.|$))/g,
+            gtref = /(?:(&lt;[^;&\s]+|\s|)&gt;)/g,
+            ampref = /(?:&amp;([^&;\s]+;|.|$))/g;
+
+        xml = xml.replace(ltref, function(match, lt){
+            return (lt.slice(lt.length - 3) == "gt;" ? match : ("<" + lt));
         });
+
+        xml = xml.replace(gtref, function(match, gt){
+            return (gt.slice(0, 3) == "&lt" ? match : (gt + ">"));;
+        });
+
+        xml = xml.replace(ampref, function(match, amp) {
+            return (amp.slice(amp.length-1) == ";" ? match : ("&" + amp));
+        });
+
+        return xml
+        //<<< and &&& can't appear unescaped since one captures the next's '&'
     }
 
     /**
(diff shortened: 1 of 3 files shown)


biyeun (approved):
this is great! and thank you for the tests 🎉

src/xml.js
```diff
@@ -126,9 +126,9 @@ define([
             return value; // value contains no character entity references or empty tags.
         }
         var xml = normalize(value, inner),
-            refs = /(?:&lt;(=?\s)|(\s)&gt;|&amp;(\s))/g;
+            refs = /(?:&lt;(=?\s)|(\s)&gt;|&amp;)/g;
```
mjriley, line 129: Just want to question if this change is indeed safe. The whitespace captures looked random, but it seems likely that the reason that we're matching whitespace after '<', whitespace before '>', and whitespace after '&' is specifically because that means those symbols aren't being used maliciously. '&amp;'. I'd imagine that the above code means that 'A>B' comes out mangled, but intentionally so. Do we risk anything by allowing '&lt;' type tokens into the strings?

minhaminha: I can confirm that the the '<' and '>' characters without whitespaces do show up escaped in the question tree (A&gt;B for the example you described), though I'm not entirely sure if displaying them in that form pose any risks.

mjriley: I'm not familiar enough with this code to know why Daniel felt that we shouldn't escape '&' if used as a tag -- it might make sense to wait for his feedback. If we commit to this change as is, however, I think the function comment also needs to change. It claims:
> Un-escapes &, < and > when not used as tag delimiters...

which is no longer the case for ampersand with this change.

millerdev: I'm leaning toward @mjriley's concern that this may not be safe. I think I required a space so as to not double-escape character entity references like `&lt;`. However, it does not appear that I included any tests to verify that those references pass through unescaped (I would have expected such a test to fail with the change you have made here). I'm having a hard time evaluating if this will mangle any pre-existing forms if they are loaded and then saved by the form builder? It may be worthwhile to try to come up with and test a scenario where a character entity reference is expected to pass through.

> If we commit to this change as is, however, I think the function comment also needs to change. It claims:
>
>>    Un-escapes &, < and > when not used as tag delimiters...
>
> which is no longer the case for ampersand with this change.

+1

millerdev: Discussed in meeting with Minha (abbreviated):

- This regex/replacement should not touch or change `&amp;amp;`. A test should be added to assert this invariant.
- The fix for that is something like
```suggestion
            refs = /(?:&lt;(=?\s)|(\s)&gt;|&amp;([^\s];|\s)/g;
```
And then below the `("&" + amp)` expression would change to something like
`(amp.endsWith(";") ? match : ("&" + amp))`

millerdev (changes requested):
Unfortunately it is not safe to unescape `<` and `>` characters this way. For example, here's a new "XML humanizer" test that now fails but it should not (and did not prior to this PR):
```js
        it("should not convert escaped tag with attribute", function () {
            eq('&lt;div class="injection"&gt;', '&lt;div class="injection"&gt;', true);
        });
```
This test should be added to the suite so we know to handle the scenario in the future.

We cannot fix this issue with regular expressions alone because [HTML is not a context-free grammar](https://stackoverflow.com/a/1758162/10840). Instead, I think we would need to implement a custom HTML parser, which is clearly out of scope. This is the reason I had previously structured the matching expression to require a space after `<` (it only matches text that is [not part of a valid start- or end-tag](https://www.w3.org/TR/REC-xml/#sec-starttags)). In retrospect the requirement of a space before `>` may be less important, although it highlights a weakness in the existing unescaping logic since a `>` preceded by a space is a valid end tag delimiter that would be unescaped (although I'm not sure it's a problem in any practical scenario).

On reading the [XML start-tag spec](https://www.w3.org/TR/REC-xml/#sec-starttags), I noticed it is illegal to have a `<` in an XML attribute value, so the unescaping of that character in attribute values is producing malformed XML. Not sure if that ever matters for us in practice.

I think my suggestion at this point is to revert to something more like the previous implementation with expanded form as outlined in https://github.com/dimagi/Vellum/pull/1019#discussion_r678342670 to preserve character entity references.

package.json
```diff
@@ -21,7 +21,7 @@
         "MediaUploader": "dimagi/MediaUploader#master",
         "XMLWriter": "dimagi/XMLWriter#master",
         "bootstrap": "3.4.1",
-        "codemirror": "5.30.0",
+        "codemirror": "5.61.1",
```
millerdev, line 24: This file should not change in this PR

src/expressionEditor.js
```diff
@@ -450,6 +450,8 @@ define([
                 advancedInput.on('change keyup', handleChange);
             }
 
+            getTopLevelJoinSelect().change(handleChange);
```
millerdev, line 453: This file should not change in this PR

tests/expressionEditor.js
```diff
@@ -13,7 +13,7 @@ define([
 ) {
     var assert = chai.assert;
 
-    describe("The expression editor xpath validator", function () {
+    describe("The expression editor", function () {
```
millerdev, line 16: This file should not change in this PR

yarn.lock
```diff
@@ -173,6 +173,13 @@ assert-plus@1.0.0, assert-plus@^1.0.0:
   resolved "https://registry.yarnpkg.com/assert-plus/-/assert-plus-1.0.0.tgz#f12e0f3c5d77b0b1cdd9146942e4e96c1e4dd525"
   integrity sha1-8S4PPF13sLHN2RRpQuTpbB5N1SU=
 
+assert@1.4.1:
```
millerdev, line 176: This file should not change in this PR

millerdev (commented):
Looks like some commits got included in this PR that do not belong here. I recommend an interactive rebase on master to remove those commits.

millerdev (approved):
Looks good! :tada:

joxl (approved):

PR #1025, opened 2021-09-02
Update js-xpath to v0.0.6

gherceg:
## Summary

A followup to [this PR](https://github.com/dimagi/Vellum/pull/1024). I didn't update `package.json` in js-xpath to v0.0.5, so it was unclear which version was being referenced by vellum. Rather than messing with updating the v0.0.5 tag on js-xpath, we decided to make a new version that properly updates package.json. All this PR is responsible for is referencing that new tag.
## Safety Assurance

### Rollback instructions


millerdev (approved):

gherceg:
Did a quick test on staging.

PR #1068, opened 2022-08-29
Install lock file version of yarn for GitHub actions tests

joxl:
## Summary

Followup to #1063. Make GitHub actions install a version of yarn that complies with the `yarn.lock` requirement.

## Safety Assurance

### Automated test coverage

GitHub actions (tests) only.

### Safety story

Tests only.

### Rollback instructions

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -12,9 +12,12 @@ jobs:
     - uses: actions/setup-node@v3
       with:
         node-version: 14
+    - name: Install dependencies
     - run: |
         yarn_version=$(grep -Eo '^yarn@[^:]+' ./yarn.lock)
         npm install "$yarn_version"
         yarn install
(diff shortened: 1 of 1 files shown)


millerdev (approved):
Cool! Great to avoid the hard-coded version.

~However, seems like it would be possible to use an old, and possibly vulnerable to some type of attack, version of yarn by making a branch with a specially crafted yarn.lock file. I'm not sure if that could be abused for some nefarious end. Should we be concerned about that?~

Edit: Of course, this is why actions are not automatically run on PRs made by untrusted users. Ignore what I said.

joxl:
> making a branch with a specially crafted yarn.lock file. [...] Should we be concerned about that?

Not that I'm aware of. Without this change, the same can be accomplished by making a branch with a specially crafted `workflows/tests.yml` file.

PR #1075, opened 2022-12-23
Track metrics for vellum dependencies

biyeun:
## Summary
Using the structure we have for tracking HQ dependencies, I've copied the relevant utilities to vellum (not my favorite solution) to push dependency metrics to Datadog. Metrics are pushed to Datadog with its own github action

## Safety Assurance

### Automated test coverage
N/A

### QA Plan
Just metrics

### Safety story
Just Metrics

### DIFF
--- a/.github/workflows/metrics.yml
+++ b/.github/workflows/metrics.yml
@@ -1,0 +1,30 @@
+name: vellum metrics
+on:
+  pull_request:
+    branches:
+    - master
+jobs:
+  tests:
+    runs-on: ubuntu-22.04
+    timeout-minutes: 60
+    env:
+      DATADOG_API_KEY: ${{ secrets.DATADOG_API_KEY }}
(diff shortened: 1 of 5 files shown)


joxl (approved):
> (not my favorite solution)

I think the only way around this would be to put the tools in a separate repo, then define how other repos fetch and use them.  The only options I see are either duplication (this solution), or a lot of extra overhead.  Personally, I'm fine with this.

Alternatively, we could add a header in these scripts, something like:

```shell
# This script duplicates <other repo/script>, if you need to change it, consider
# also changing its other location(s).
```

Completely up to you, I'm fine merging this as-is.

millerdev (approved):
Which files in this PR other than `.github/workflows/metrics.yml` have custom Vellum code in them? It would be nice to have the dependency metrics logic be available as a Python library that can be installed with something like `pip install yarn_metrics`. The bash scripts may need to be re-implemented in Python.

PR #1081, opened 2023-03-31
Update CKEditor to latest version (4.21.0)

ajeety4:
## Summary
- [JIRA Ticket](https://dimagi-dev.atlassian.net/browse/INDIV-88)
- Updates CK Editor 4 to the at present latest version (4.21.0) and addresses below points. We are using a custom build (with a patch) and hence the update process is manual. 
- As part of this update, the plugin `toolbar` was included by default as clipboard is dependent on it. Plugin `floatingspace` was removed as it is not required and also because it created an empty toolbar on the rich text editor.
- Html to text handler was modified to remove additional span elements created by widgets plugin while copying widgets.
- Test cases for copying widgets were modified to wait for the copy process to complete as it is an async operation.
- **About the build**: This build was created from source. [Patch 1](https://github.com/ckeditor/ckeditor4/pull/308) was addressed in upstream repo  however [Patch 2](https://github.com/ckeditor/ckeditor4/pull/304) was not. Patch 2 is part of this build. For changes that were made on the build, see [commit1](https://github.com/dimagi/ckeditor-dev/commit/8ba91a8769c91d1e2b8a6c26b6f7681d3930e706) that reverts the patch1 for simplicity and  [commit2](https://github.com/dimagi/ckeditor-dev/commit/7c2c6abc0fa5df4f6cbcba0cfaf367384a461184) that merges the 4.21.0 release on this [branch](https://github.com/dimagi/ckeditor-dev/commits/vellum-build).

## Safety Assurance

- [ ] I have confidence that this PR will not introduce a regression for the reasons below

### QA Plan

https://dimagi-dev.atlassian.net/browse/QA-4973

### Safety story
- Local testing was done to ensure patches: [Patch 1](https://github.com/ckeditor/ckeditor4/pull/308)  and [Patch 2](https://github.com/ckeditor/ckeditor4/pull/304) are working correctly.
- Automated test cases were run and issues were addressed.

### Rollback instructions

### Note for Reviewers

- Since this is a manual update to the new version, this [commit](https://github.com/dimagi/Vellum/pull/1081/commits/b57a16afd3819f2772ac2946e6f2911ed514dc4c) has bulk of file changes, the CK Editor 4 build itself.


src/richText.js
```diff
@@ -731,6 +735,9 @@ define([
                    .replace(/<\/p>/ig, "\n")
                    .replace(/<br \/>/ig, "\n")
                    .replace(/(&nbsp;|\xa0|\u2005)/ig, " ")
+                   // While copying widgets with text, CKEditor adds these html elements
+                   .replace(/<span\b[^>]*?id="?cke_bm_\d+(\w)"?\b[^>]*?>.*?<\/span>/ig, "")
```
millerdev, line 739: Few questions
- Why group on `(\w)`? And is that intended to match a single word character or multiple?
- Is a quotation mark (") considered a word boundary? If not, would it be better to change the second `\b` to `\s`?
- Should the contents of the span be captured in a group and included in the replacement text?

lib/ckeditor/dimagi-README.md
```diff
@@ -1,36 +1,49 @@
...
 
 ## Building from source
 
-Use this procedure when building with patches that are not yet included in the
-main line. Feel free to use more recent version numbers if that makes sense.
-
-- clone https://github.com/millerdev/ckeditor-dev and `cd` into ckeditor-dev
-- checkout *vellum-build* branch
-- download the minimalist skin (v1.0) and unzip/place it in the ./skins
+The repo [dimagi/ckeditor-dev](https://github.com/dimagi/) which is fork of upstream repo is used for creating a build from source using the branch **vellum-build**.
```
millerdev, line 28:
```suggestion
The repo [dimagi/ckeditor-dev](https://github.com/dimagi/ckeditor-dev) which is fork of upstream repo is used for creating a build from source using the branch **vellum-build**.
```

ajeety4:
Also pasting here the sample of html generated by CK Editor while copying widgets for reference.
The additional elements created are highlighted.

Screenshot from 2023-04-03 19-38-01

src/richText.js
ajeety4: > Few questions
FYI - I adapted both the regex's from the CK code base itself so as to keep the match as close as possible.
Also see sample html for reference [here](https://github.com/dimagi/Vellum/pull/1081#issuecomment-1494450408)

> * Why group on `(\w)`? And is that intended to match a single word character or multiple?
My bad and good catch. Grouping is not needed at all and will be removing it. Intended to match single character as per the id format for bookmark.

> * Is a quotation mark (") considered a word boundary? If not, would it be better to change the second `\b` to `\s`?
Yes quotation mark is part of word boundary.

> * Should the contents of the span be captured in a group and included in the replacement text?
The bookmark span element has only fixed content: `&nbsp;`to support IE, which causes an additonal whitespace character in our plain text. This is not needed in replacement text.

src/richText.js
```diff
@@ -731,6 +735,9 @@ define([
...
+                   .replace(/<span\b[^>]*?data-cke-copybin-(start|end)[^<]*?<\/span>/ig, "")
```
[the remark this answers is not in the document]
ajeety4: - As per my understanding, these additional elements(including the bookmarks one) are created for copying widgets on top of the actual html content. 
- This pollutes our plain text with new line character at the end as it wraps the outermost p tag.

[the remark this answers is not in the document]
ajeety4: Sounds good! Will be adding them.

ajeety4:
> I don't have context on most of this so just took a quick pass.
> 
> I believe we are going to send this through QA?
 Yes, I will be creating a QA ticket for this.

lib/ckeditor/dimagi-README.md
```diff
@@ -6,7 +6,6 @@ Go to the [CK build editor](http://ckeditor.com/builder).
 Use the following plugins:
 * clipboard (for drag and drop)
 * entities
-* floatingspace
```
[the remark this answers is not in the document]
ajeety4: Good point. 
Not necessary as it is an internal dependency for clipboard which gets automatically included while using the online builder.
However, it would be useful to have a list of sub dependencies (there are others too) when building from source. 
**Will be including them in a sub section.**

ajeety4:
> @millerdev @ajeety4 Are you able to share what would be the impact if this is released with a bug despite our best efforts? The form updates on App builder will fail to save or we would end up with bad forms that might also get released to the mobile users?

For more context (ignore if already known), the rich text editor is built on top of CK Editor and it is used in text fields, display conditiions, expression editor etc. 
I am not sure on the mobile side. On the form builder side, depending on the bug it can impact things like copy , paste or even the [easy references](http://www.dimagi.com/blog/simplify-app-building-in-commcare-with-easy-references/) itself.

ajeety4:
@millerdev @mkangia (FYI)
While doing some additional testing for widgets, I had one observation related to the keyboard navigation.
This is applicable for live HQ version and the updated CK Editor version (this PR) for the below specific scenario.
- Add a widget(easy reference) at the start of the editor and add some text after that.
- Select the widget(fake):  The keys 'End , Shift+End(Right/Down) does not work i.e widget plus right side selection of text.
image

Note: Rest of keys work fine, things are also good if there is text before the widget, select text first and then widget and it works fine.
From the patch, I understand that this selection thing is handled by the default CK editor handler.
Not a showstopper, however still something wanted to point out! 
I think it is good to have a separate ticket for this in the backlog and prioritize accordingly. Any thoughts?

src/richText.js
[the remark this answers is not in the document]
ajeety4: These were added

tests/richText.js
```diff
@@ -499,6 +499,10 @@ define([
                 }
                 assert.equal(data.Text, value, 'text/plain');
                 assert.strictEqual(data["text/html"], undefined, 'text/html');
+                // Wait for CK Editor Async handler (copybin) to complete
+                setTimeout(function(){
+                        callback();
+                },100);
```
millerdev, line 546: nit
```suggestion
                setTimeout(callback,100);
```

tests/richText.js
```diff
@@ -273,6 +273,32 @@ define([
...
+                    '///wAAACH5BAEKAAAALAAAAAABAAEAAAICRAEAOw==" width="15" title="Click and drag to move" height="15" role="presentation" ' +
+                    'draggable="true"></span></span>',
+                suffix_html = '</p><span data-cke-copybin-end="1">​</span>';
+
+            var text_widgets_outputs = [
+                ['This dob: <output value="#case/dob" /> is of child', prefix_html + 'This dob:&nbsp;' + widget_html + ' is of child' + suffix_html],
+                ['This dob: <output value="#case/dob" />', prefix_html + 'This dob:&nbsp;' + widget_html +  suffix_html],
+                ['<output value="#case/dob" /> is of child', prefix_html + widget_html + ' is of child' + suffix_html],
+                ['<output value="#case/dob" />', prefix_html_1 + widget_html + suffix_html],
+                ['This dob: <output value="#case/dob" /> is of child', 'This dob: &lt;output value="#case/dob" /&gt; is of child']
```
millerdev, line 301: Could these should be added to the `text` array above? Seems like this would allow removing the `_.each(text_widgets_outputs, ...` constructs below, eliminating duplication.

src/richText.js
millerdev: Looks like the extra group around `\w` has not been removed. Is that group in the original regex you copied from CK code? If yes, then fine to keep it so it matches as closely as possible.

tests/richText.js
ajeety4: Sure. My initial thought was to keep the widgets html separate, however it is not needed as such and this is better as avoids duplication.

src/richText.js
ajeety4: Sorry got missed and thanks for catching :). Yes the one I copied has it and is used for some processing. Since we don't need it, will get rid of the extra group.

millerdev (approved):
nit: rebase or force-push in general after reviews have been submitted is disorienting if it modifies any commits that have been reviewed because comments become disassociated with the commits they were reviewing. Better to merge master into the branch, assuming that was the intent of the rebase.

ajeety4:
> nit: rebase or force-push in general after reviews have been submitted is disorienting if it modifies any commits that have been reviewed because comments become disassociated with the commits they were reviewing. Better to merge master into the branch, assuming that was the intent of the rebase.

Noted! Thanks Daniel.

ajeety4:
Update: QA: https://dimagi-dev.atlassian.net/browse/QA-4973 has passed for this.
Will be proceeding with local sanity testing on HQ master follwed by these [deploy](https://confluence.dimagi.com/display/GTD/Deploy+Process#DeployProcess-vellum:~:text=control%20staging%20deploy-,Deploying%20Vellum,-Vellum%20is%20deployed) steps.

mkangia:
@ajeety4 
Can you share the process?
Do you need to rebuild vellum and put that in a PR for commcare-hq repo?

millerdev:
I think you will need to omit the `--push` option when running `vellum-to-hq` and make a PR with the changes it creates. I updated the docs.

PR #1088, opened 2023-05-17
Replace metrics scripts with package

gherceg:
## Summary

https://dimagi-dev.atlassian.net/browse/SAAS-14266

Replaces the group of scripts used to track outdated dependency metrics with the new `dependency-metrics` pypi package.

Differences
The "exotic" metric has been replaced with "unknown" as this better reflects what this metric means, and is not specific to any language/package manager.

The "host" metric specifies which project the metrics are sent from, which means our datadog dashboard will need to be updated to filter to a specific host for HQ dependencies.

## Safety Assurance

### QA Plan

No QA.

### Safety story

Internal tool.

### Rollback instructions

Will result in more metrics disruptions.


millerdev (approved):
:star:

PR #1092, opened 2023-09-15
commcare connect deliver unit mug

snopoke:
Jira: https://dimagi-dev.atlassian.net/browse/CCPC-110

## Summary
This adds a new 'mug' to the `commcareConnect` plugin which is used by CommCare Connect projects to track service delivery. A delivery unit XML block is added to a form that represents a unit of service delivery. When this form is forwarded to CommCare Connect, the details in the XML block are processed to record the service delivery.

```xml
<deliver xmlns="http://commcareconnect.com/data/v1/learn" id="unit_one">
    <name>unit 1</name>
    <entity_id>123123</entity_id>
    <entity_name>23 Mvuramanga Road</entity_name>
</deliver>
```

image

## Feature Flag
CommCare Connect

## Safety Assurance

### Automated test coverage
Updated unit tests.

### QA Plan
None

### Safety story
The changes are limited to the CommCare Connect plugin which is only enabled based on the FF. The changes have unit test coverage and were tested locally.

### Rollback instructions

### DIFF
--- a/src/commcareConnect.js
+++ b/src/commcareConnect.js
@@ -29,6 +29,20 @@ define([
                 "nodeID",
             ],
         },
+        logicSection = {
+            slug: "logic",
+            displayName: gettext("Logic"),
+            help: {
+                title: gettext("Logic"),
+                text: gettext("Use logic to control when questions are asked and what answers are valid. " +
+                    "You can add logic to display a question based on a previous answer, to make " +
+                    "the question required or ensure the answer is in a valid range."),
+                link: "https://confluence.dimagi.com/display/commcarepublic/Common+Logic+and+Calculations"
+            },
+            properties: [
+                'relevantAttr',
+            ]
+        },
         baseSpec = {
             xmlnsAttr: {
                 presence: "optional",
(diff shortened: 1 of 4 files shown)


src/commcareConnect.js
```diff
@@ -222,7 +300,9 @@ define([
         parseBindElement: function (form, el, path) {
             let mug = form.getMugByPath(path);
             if (!mug) {
+                // check each mugConfig to see if this is path matches
```
millerdev, line 303: Not sure of correct wording here. Is this suggestion correct?
```suggestion
                // check each mugConfig to see if this path matches
```

millerdev (approved):

millerdev:
The description of this PR and also the linked Jira ticket are very sparse in detail of what a `DeliverUnitMug` is or why it is needed. Is there a document somewhere that has more information about that? Personally, I think it is preferable to add details like that to the public PR.

calellowitz:
I think we want to add the entity tracking we previously discussed to these blocks, if possible. At this point a simple "entity_id" and "entity_name" that take xpath expressions so that they can be set based on form data would be sufficient, but I am not sure the added level of effort for that.

snopoke:
> I think we want to add the entity tracking we previously discussed to these blocks, if possible. At this point a simple "entity_id" and "entity_name" that take xpath expressions so that they can be set based on form data would be sufficient, but I am not sure the added level of effort for that.

I've added in `entity_id` and `entity_name` as optional xpath fields (see updated screenshot).

calellowitz (approved):

PR #1108, opened 2024-02-06
Add Micro Image Mug

zandre-eng:
## Summary

Link to ticket [here](https://dimagi.atlassian.net/browse/SC-3346).

Related PR that creates relevant feature flag can be found [here](https://github.com/dimagi/commcare-hq/pull/34087).

This PR is part of a larger change set in which micro case images are added to CommCare behind a new feature flag.

## Feature Flag

`CASE_MICRO_IMAGE`

## Product Description

A new "Micro-Image" question type is added under the "Multimedia Capture" group:
Screenshot from 2024-02-06 16-25-16

## Safety Assurance

- [X] If QA is part of the safety story, the "Awaiting QA" label is used
- [X] I have confidence that this PR will not introduce a regression for the reasons below

The added code sits behind the new feature flag, and so will only execute if the given FF is enabled.

### Automated test coverage

Automated tests exist to ensure properties of new `MicroImage` mug type are correct.

### QA Plan

None

### Rollback instructions

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -386,15 +386,7 @@ define([
                     "FieldList"
                 ]
             },
-            {
-                group: ["Image", gettext('Multimedia Capture')],
-                questions: [
-                    "Image",
-                    "Audio",
-                    "Video",
-                    "Signature"
-                ]
-            },
+            mediaGroup,
             {
                 group: ["Trigger"],
                 questions: ["Trigger"],
--- a/src/core.js
+++ b/src/core.js
@@ -349,7 +349,7 @@ define([
                 "Signature"
             ]
         };
-        if (this.opts().features['case_micro_image']) {
+        if (this.opts().features.case_micro_image) {
             mediaGroup.questions.push("MicroImage");
         }
 
--- a/src/core.js
+++ b/src/core.js
@@ -340,6 +340,19 @@ define([
      *              - questions: [mugType, ...]
      */
     fn.getQuestionGroups = function () {
+        let mediaGroup = {
+            group: ["Image", gettext('Multimedia Capture')],
+            questions: [
+                "Image",
+                "Audio",
+                "Video",
+                "Signature"
+            ]
+        };
+        if (this.opts().features['case_micro_image']) {
+            mediaGroup.questions.push("MicroImage");
+        }
+
         return [
             {
                 group: ["Text"],
--- a/src/mugs.js
+++ b/src/mugs.js
@@ -1445,6 +1445,18 @@ define([
         }
     });
 
+    var MicroImage = util.extend(Audio, {
+        typeName: gettext('Micro-Image'),
+        isTypeChangeable: false,
+        icon: 'fa fa-camera',
+        tagName: 'input',
+        mediaType: "image/*", /* */
+        init: function (mug, form) {
+            Audio.init(mug, form);
+            mug.p.appearance = "micro-image";
+        }
+    });
+
     var Video = util.extend(Audio, {
         typeName: gettext('Video Capture'),
         icon: 'fa fa-video-camera',
@@ -1879,6 +1891,7 @@ define([
                 "Geopoint": Geopoint,
                 "Group": Group,
                 "Image": Image,
+                "MicroImage": MicroImage,
                 "Int": Int,
                 "Long": Long,
                 "MSelect": MSelect,
--- a/tests/exporter.js
+++ b/tests/exporter.js
@@ -42,7 +42,10 @@ define([
                         done();
                     }
                 },
-                features: {rich_text: false},
+                features: {
+                    rich_text: false,
+                    case_micro_image: true
+                },
             });
         });
 
--- a/tests/questionTypes.js
+++ b/tests/questionTypes.js
@@ -162,6 +162,11 @@ define([
                 path: 'question22/question23/',
                 nodeId: 'question7'
             }, {
+                //clickBeforeAdd: "question32",
+                type: 'MicroImage',
+                path: 'question22/question23/',
+                nodeId: 'question35',
+            }, {
                 clickBeforeAdd: "question19", // insert before question22
                 type: 'Group',
                 nodeId: 'question21'
@@ -172,9 +177,6 @@ define([
                 attrs: {
                     repeat_count: 2
                 }
-            }, {
-                type: 'MicroImage',
-                nodeId: 'question35'
             }
         ];
 
--- a/tests/questionTypes.js
+++ b/tests/questionTypes.js
@@ -188,6 +188,7 @@ define([
                             done();
                         }
                     },
+                    features: { case_micro_image: true }
                 });
             });
 
--- a/tests/questionTypes.js
+++ b/tests/questionTypes.js
@@ -172,6 +172,9 @@ define([
                 attrs: {
                     repeat_count: 2
                 }
+            }, {
+                type: 'MicroImage',
+                nodeId: 'question35'
             }
         ];
 
@@ -720,6 +723,16 @@ define([
         it("should not write the type", function () {
             util.loadXML(SELECT1_HELP_WITH_TYPE_XML);
             util.assertXmlEqual(util.call('createXML'), SELECT1_HELP_XML);
+        });
+    });
+
+    describe("Micro-Image question", function () {
+        it("should have correct options set", function () {
+            util.loadXML("");
+            var microImage = util.addQuestion("MicroImage", "microimage");
+            assert.strictEqual(microImage.options.mediaType, 'image/*');
+            assert.strictEqual(microImage.options.tagName, 'input');
+            assert.strictEqual(microImage.p.appearance, 'micro-image');
         });
     });
 });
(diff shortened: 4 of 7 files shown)


src/core.js
```diff
@@ -340,6 +340,19 @@ define([
      *              - questions: [mugType, ...]
      */
     fn.getQuestionGroups = function () {
+        let mediaGroup = {
```
SmittieC, line 343: Unless I'm misunderstanding it, should `mediaGroup` not be injected into the object being returned? (and replace the similar one already in there)

millerdev (commented):
Vellum has very thorough test coverage, but I don't see new tests here. Maybe the [`Image` mug tests](https://github.com/dimagi/Vellum/blob/24d34182bf6efd42cae93fbd0a5200fcb6678981/tests/questionTypes.js#L697-L709) could be used as an inspiration for testing this new mug type?

src/core.js
zandre-eng: Good catch! I was porting these changes over from HQ code (as I was testing there) and this must have slipped through. This has been fixed (77bd00a1a7cb01de518011f3339e052f30e3bf65)

zandre-eng:
> Vellum has very thorough test coverage, but I don't see new tests here. Maybe the [Image mug tests](https://github.com/dimagi/Vellum/blob/24d34182bf6efd42cae93fbd0a5200fcb6678981/tests/questionTypes.js#L697-L709) could be used as an inspiration for testing this new mug type?

@millerdev That is an excellent idea, thanks for the example reference. I'll add some tests.

src/mugs.js
```diff
@@ -1445,6 +1445,18 @@ define([
         }
     });
 
+    var MicroImage = util.extend(Audio, {
```
mkangia, line 1448: Can this extend from Image instead of Audio?

mkangia (commented):
Looking good.
Definitely agree on adding tests.

src/mugs.js
mkangia: I believe that is happening because of the `spec` of the Image which would need to over written though that makes it quite dirty.

Extending Audio for Image is confusing to read, may be it should be a `BaseMedia` object that should be extended for all. Though I see that is what we have been doing elsewhere as well. So not something you should address in this PR.

zandre-eng: > may be it should be a `BaseMedia` object that should be extended for all

This is a nice idea, and I think would help make things more readable. But I agree, since there is already pre-existing classes extending from `Audio`, this change is better handled in a separate PR.

tests/questionTypes.js
```diff
@@ -188,6 +188,7 @@ define([
                             done();
                         }
                     },
+                    features: { case_micro_image: true }
```
Charl1996, line 193: Why is it needed to add `case_micro_image` here? (I don't have any context on how vellum works, but seeing this here and no other existing features makes me wonder whether I'm missing something)

zandre-eng: The reason for adding `case_micro_image` here is because we hide the new mug behind a feature flag. We check whether a domain has the necessary flag enabled by looking at the list of features sent by [this](https://github.com/dimagi/commcare-hq/blob/a331be04b86657e62c4ce19cfd4b21571eafc81b/corehq/apps/app_manager/views/formdesigner.py#L311) HQ function. Since the HQ function doesn't get called when running tests, we need to mock having the feature flag enabled for the tests.

Charl1996: Ah, I see. Thanks for clarifying.

tests/exporter.js
```diff
@@ -42,7 +42,10 @@ define([
                         done();
                     }
                 },
-                features: {rich_text: false},
+                features: {
+                    rich_text: false,
+                    case_micro_image: true
```
mkangia, line 47: Can you share why was this changed?
(Not well versed with vellum tests so hard to put this all together)

tests/questionTypes.js
mkangia: So we need to set it true for all tests by default?

tests/questionTypes.js
```diff
@@ -161,6 +161,11 @@ define([
                 type: 'AndroidIntent',
                 path: 'question22/question23/',
                 nodeId: 'question7'
+            }, {
+                //clickBeforeAdd: "question32",
```
mkangia, line 165: Can you expand on this comment?

mkangia: why did we move the `MicroImage` block up?

tests/static/all_question_types.xml
```diff
@@ -204,13 +204,14 @@
...
-					<upload ref="/data/question22/question23/question29" mediatype="image/*" appearance="signature"/>
-					<input ref="/data/question22/question23/question7" appearance="intent:question7"/>
+					<upload ref="/data/question22/question23/question24" mediatype="image/*" jr:imageDimensionScaledMax="250px" />
+					<upload ref="/data/question22/question23/question25" mediatype="audio/*" />
+					<upload ref="/data/question22/question23/question26" mediatype="video/*" />
+					<input ref="/data/question22/question23/question27" />
+					<secret ref="/data/question22/question23/question28" />
+					<upload ref="/data/question22/question23/question29" mediatype="image/*" appearance="signature" />
+					<input ref="/data/question22/question23/question7" appearance="intent:question7" />
+					<input ref="/data/question22/question23/question35" mediatype="image/*" appearance="micro-image" />
```
mkangia, line 214: nit: the addition of question35 got hidden in the changes to add extra space before the closing tag here.
Good to keep those in separate commits so relevant changes don't get missed.

mkangia:
Thanks @zandre-eng 
Mostly non blocking feedback. I don't have good expertise with vellum so just asked questions mostly.

tests/questionTypes.js
zandre-eng: No, this only needs to be doneeor tests that include the new `MicroImage` mug type. Currently, only the tests I have added use this new mug type.

tests/exporter.js
zandre-eng: Please refer to [this](https://github.com/dimagi/Vellum/pull/1108#discussion_r1481566734) comment. Essentially, this was added to mock the feature flag being enabled for the test.

tests/questionTypes.js
zandre-eng: Ah, that must have slipped through when I was working on this test. Will remove this comment.

> why did we move the MicroImage block up?

This was purely to help make it more readable by putting it in the same position on the test as it is in the `TEST_XML` file.

tests/static/all_question_types.xml
zandre-eng: Good point. GitHub doesn't make it very obvious which changes are just spacing changes and which are new lines of code. Will keep this in mind next time.

zandre-eng:
> So we need to set it true for all tests by default?

@mkangia This doesn't have to be set to true for all tests, only the ones that use the new `MicroImage` mug type. For more context, please refer to [this](https://github.com/dimagi/Vellum/pull/1108#discussion_r1481566734) comment.

mkangia (approved):
Looks good to me.
Though good to wait for Daniel's feedback on this as he has worked more closely with Vellum.

millerdev (commented):
Looks good, although I'm struggling to understand what a "micro image" is and how it differs from a normal image. Is the only difference that it has a custom `appearance` attribute in the form XML?

Instead of adding a new question type, would it work to simply enter the custom value into the _Appearance Attribute_ under _Advanced_ properties of a normal _Image_?

image

My question bears partly on how widely we expect this feature to be used, and whether it is planned to go GA at some point in the future? If it's a one-off for a project or two then maybe it would make sense to wait on implementing a custom mug type and feature flag until it is more widely used?

zandre-eng:
@millerdev

> I'm struggling to understand what a "micro image" is and how it differs from a normal image.

Micro images are intended to be highly compressed small images that users capture of a person's face. This image is saved as a case property, which is then made viewable in the case list/details view with custom case tiles. For more context, you can have a look at [the tech spec](https://docs.google.com/document/d/14yDNK6RyQGB-8H6DioGgoj0c2mwJuI8nWck4zFMYFAs/edit?usp=sharing) for this feature.

> Is the only difference that it has a custom appearance attribute in the form XML?

Apart from the appearance change, we also change the `tagName`. Unlike an `Image` mug which has an `upload` tag, we use an `input` tag for the `MicroImage` mug. The reason for this difference is that we want to be able to save the micro image as a case property, which is not possible when using an `upload` tag. 

> Instead of adding a new question type, would it work to simply enter the custom value into the Appearance Attribute under Advanced properties of a normal Image?

We would also need to change the `tagName` if we use a normal Image for the micro image. We could potentially dynamically change the tag on init if the appearance is `micro-image`, however this feels like a less clean solution. But let me know if you think this would work better.

> how widely we expect this feature to be used, and whether it is planned to go GA at some point in the future?

This is a great question. I had a chat with Mary regarding this, and whilst it is only a few projects for the time being, there are definitely plans to GA this in the future. We also intend to do some follow up work after the initial feature is released, such as making it more user friendly to add to a case list.

millerdev (approved):
Thanks for answering my questions.

PR #1110, opened 2024-02-20
update yarn.lock

biyeun:
## Summary
Update `yarn.lock` so that build can complete. This was the output after running `yarn install` locally.

## Safety Assurance

### Automated test coverage
Yes

### QA Plan
No

### Safety story
Safe change as this is only affecting the `yarn.lock` file. need to update otherwise the `vellum-to-hq` command won't complete.

### Rollback instructions


biyeun:
https://github.com/dimagi/Vellum/pull/1111

PR #1111, opened 2024-02-20
remove yarn pin from package.json

biyeun:
## Summary
This removes the `yarn` pin from `package.json`. This is not good practice to do.

This needs to be done so that the `vellum-to-hq` command can finish packaging vellum for HQ

## Safety Assurance

### Automated test coverage
yes

### QA Plan
no

### Safety story
safe. just updates dependency on `yarn`. tests will fail if this is a bad idea.

### Rollback instructions

### DIFF
--- a/.github/workflows/metrics.yml
+++ b/.github/workflows/metrics.yml
@@ -25,9 +25,7 @@ jobs:
           3.9
     - name: Install dependencies
       run: |
-        yarn_version=$(grep -Eo '^yarn@[^:]+' ./yarn.lock)
-        echo "lock file requires: $yarn_version"
-        npm install "$yarn_version"
+        npm install "yarn@^1"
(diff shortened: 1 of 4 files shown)


millerdev:
Looks like js-xpath should be upgraded to 0.0.8. It is what is requiring yarn, per yarn.lock:
```
xpath@dimagi/js-xpath#v0.0.7:
  version "0.0.7"
  resolved "https://codeload.github.com/dimagi/js-xpath/tar.gz/cc482ce3cd8635ce311d9b4c82dba24998492a46"
  dependencies:
    biginteger "^1.0.3"
    yarn "1.22.10"
```
Once that's upgraded the yarn dependency should go away entirely.

millerdev (approved):

PR #1124, opened 2024-09-16
Upgraded requirejs from 2.36 to 2.37

orangejenny:
## Summary
https://dimagi.atlassian.net/browse/SAAS-15983

HQ was done in https://github.com/dimagi/commcare-hq/pull/34917

## Safety Assurance

### Automated test coverage

There is test coverage, I'll also smoke test on staging.

### QA Plan

no

### Rollback instructions

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
         "langcodes": "dimagi/langcodes#master",
         "markdown-it": "12.3.2",
         "require-css": "0.1.10",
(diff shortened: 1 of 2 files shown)


gherceg (approved):
I always miss CVEs for vellum. Thank you! Remind me, would vellum end up using the version of RequireJS set in HQ, or are the versions of their dependencies independent of one another?

PR #1129, opened 2024-11-12
Remove grunt - not referenced by scripts or docs

millerdev:
This may have been used to automatically run tests when source files changed. [Original PR](https://github.com/dimagi/Vellum/pull/566).

## Safety Assurance

### Automated test coverage

No.

### Safety story

This was a dev dependency that was likely only used to run tests in a local dev environment. It's being removed since that usage was never documented, and the version is outdated.

### Rollback instructions


orangejenny:
Looks fine, but I think you should remove `Gruntfile.js`, too.

orangejenny (approved):

PR #1137, opened 2025-02-27
Updated limit for non-question atwho widgets

orangejenny:
## Summary
https://dimagi.atlassian.net/browse/SAAS-16391

This raises the limit from the [default](https://dimagi.atlassian.net/browse/SAAS-16391) of 5 to 10, which is the [same](https://github.com/dimagi/Vellum/blob/b193bd450901efd6e419f952f4e42548b037b175/src/atwho.js#L160) used for the question dropdown (this list of questions that shows up when you type `#form`). Ten is bigger than working memory (which is 5-7 items), so it's long enough that people are less likely to look at it, grasp all the items, and feel like something is missing.

## Safety Assurance

### Automated test coverage

Not of this behavior.

### QA Plan

No.

### Safety story
Trivial config change.

### Rollback instructions

### DIFF
--- a/src/atwho.js
+++ b/src/atwho.js
@@ -55,6 +55,7 @@ define([
         var options = {
             at: "",
             data: choices,
+            limit: 10,
             maxLen: Infinity,
             suffix: "",
             tabSelectsMatch: false,


gherceg (approved):
Nice one!

PR #1138, opened 2025-03-05
Display warning for invalid XPath expression

millerdev:
Invalid XPath expressions are now flagged with a warning message. Questions with warnings are highlighted in the question tree, and the save button displays an error on mouse-over.

image

https://dimagi.atlassian.net/browse/SAAS-16376

## Safety Assurance

### Automated test coverage

Yes.

### Safety story

Simply adds a warning when an XPath expression is not valid. Invalid XPath expressions were previously not highlighted in any special way, although the logic manager could not parse them.

### Rollback instructions

### DIFF
--- a/src/logic.js
+++ b/src/logic.js
@@ -7,12 +7,16 @@ define([
     _,
     util
 ) {
-    var EXTERNAL_REF = "#";
+    var EXTERNAL_REF = "#",
+        INVALID_XPATH = "#invalid/xpath ";
 
     function LogicExpression (exprText, xpathParser) {
         this._text = exprText || "";
         this._xpathParser = xpathParser;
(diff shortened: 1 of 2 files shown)


nospame:
I don't think I'm understanding how this works. Is `#invalid/xpath ` already set somewhere else that this code is able to read it and display the error?

millerdev:
> Is `#invalid/xpath ` already set somewhere else that this code is able to read it and display the error?

Yes, in the `richText` module.

https://github.com/dimagi/Vellum/blob/c5a5cb73ce8899b8787d2138eaa441e4dc74a1bd/src/richText.js#L681-L685

nospame (approved):

PR #1144, opened 2025-04-22
Upgraded version of node used in github actions

orangejenny:
## Summary
https://dimagi.atlassian.net/browse/SAAS-17438

Cherry-picked from webpack migration, which adds `css-loader`, which requires node minimum 18.12.0. See [here](https://github.com/dimagi/Vellum/actions/runs/14499301585/job/40674880030) for a test run that failed on a proof of concept branch.

I also updated setup-node to v4 because that's what HQ does.

## Safety Assurance

### Automated test coverage

It is tests.

### QA Plan

no

### Safety story
Only affects developer-facing workflows.

### Rollback instructions

### DIFF
--- a/.github/workflows/metrics.yml
+++ b/.github/workflows/metrics.yml
@@ -16,9 +16,9 @@ jobs:
       DATADOG_APP_KEY: ${{ secrets.DATADOG_APP_KEY }}
     steps:
     - uses: actions/checkout@v3
-    - uses: actions/setup-node@v3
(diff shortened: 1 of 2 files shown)


millerdev (approved):

gherceg (approved):
Moving package.json before the installing yarn is a bit hacky, but given the PR and commit message providing context, and the scope of the change, seems fine to me!

PR #1145, opened 2025-04-22
Jls/swan song

orangejenny:
Ignore, just opening for tests.


orangejenny:
nevermind, forgot that tests don't run at all in Github actions due to node incompatibility

PR #1146, opened 2025-04-23
Reduce usage of jsdiff

orangejenny:
## Summary
This replaces the usage of `jsdiff` in the "Show overwrite warning" popup with usage of `diff_match_patch`.

`jsdiff` is one of the few libraries that's only on bower, not npm, which means it's one of the libraries keeping Vellum on yarn. I was hoping to remove it altogether but didn't realize that it's also used in a [test utility](https://github.com/dimagi/Vellum/blob/c109eaa51f203c4e14b23add33d2d1042c763a4a/tests/utils.js#L110-L115) to create diffs for test results like XML snippets and text representations of jstree schemas.

## Product Description
This changes the look and feel of the diff. I think it's an improvement, but I'm biased.

### Before
Screenshot 2025-04-23 at 11 03 16 AM

### After
Screenshot 2025-04-23 at 11 13 10 AM

## Safety Assurance

### Automated test coverage

Added in this PR

### QA Plan

Not requesting QA

### Safety story
This is a limited change to a minor workflow, and it has test coverage.

### Rollback instructions

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -2,7 +2,7 @@ define([
     'require',
     'json!langCodes',
     'underscore',
-    'jsdiff',
+    'diff-match-patch',
     'vellum/markdown',
     'vellum/xml',
     'jquery',
@@ -11,7 +11,7 @@ define([
     require,
     langCodes,
     _,
-    jsdiff,
+    diff_match_patch,
     markdown,
     xml,
     $
@@ -310,7 +310,7 @@ define([
 
     that.parseXML = xml.parseXML;
 
-    that.xmlDiff = function (localForm, serverForm, opts) {
+    that.htmlXMLDiff = function (localForm, serverForm) {
         function cleanForDiff (value) {
             // convert leading tabs to spaces
             value = value.replace(/^\t+/mg, function (match) {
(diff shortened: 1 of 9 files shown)


src/util.js
```diff
@@ -322,25 +322,15 @@ define([
...
-        patch = patch.replace(/^Index:/,
-            opts.not ? gettext("XML should not be equivalent") : gettext("XML mismatch")
-        );
-        return patch;
+
+        const dmp = new diff_match_patch(),
+            diff = dmp.diff_main(localForm, serverForm);
+
+        let html = dmp.diff_prettyHtml(diff);
+        html = html.replaceAll("&para;<br>", "<br>");
```
gherceg, line 332: Is this effectively you overriding the output of `prettyHtml` based on preference? If so, is it worth noting that or does that seem clear as is?

src/util.js
```diff
@@ -322,25 +322,15 @@ define([
             }
             return value;
         }
-        opts = opts || {};
-        if (opts.normalize_xmlns) {
-            var xmlns = $(xml.parseXML(serverForm)).find('data').xmlAttr('xmlns');
-            localForm = localForm.replace(/(data[^>]+xmlns=")(.+?)"/,
-                                    '$1' + xmlns + '"');
-        }
```
gherceg, line 330: Just confirming that it was the case that nowhere called `xmlDiff` with opts provided and therefore this code is irrelevant, right?

gherceg (approved):
Looks great! Regarding the usage in the test utility, do you plan to remove that in a separate PR or is this change sufficient for your needs?

src/util.js
orangejenny: Yep, the only call to `xmlDiff` is the one updated in this PR.

src/util.js
orangejenny: Good point, this is a good place for a comment - 321bc1da

This is how things look without doing this replacement 😱 
Screenshot 2025-04-23 at 4 17 46 PM

millerdev (approved):

orangejenny:
> Regarding the usage in the test utility, do you plan to remove that in a separate PR or is this change sufficient for your needs?

I'm done for now (once tests pass...). I started working on this because I hit a wall with https://github.com/dimagi/Vellum/pull/1144 and started thinking I might need to get Vellum off of yarn in order to use the latest node. I found a workaround for that (and it's beautiful code, as you noted), so I'm going to stop working on removing jsdiff, just wanted to merge the one cohesive part that's finished.

gherceg (approved):

PR #1152, opened 2025-05-02
Moved crypto-js from lib to package.json

orangejenny:
## Summary
Get this library into package.json, where our tooling can see it.

## Safety Assurance

### Automated test coverage

I think so. 

### QA Plan

no

### Safety story
I tested locally that Velum loads and that [this line](https://github.com/dimagi/Vellum/blob/138b7fa33f23ab0016963f1539ac0f114d6a6be8/src/core.js#L2331) still generates something.

### Rollback instructions


yarn.lock
```diff
@@ -422,6 +422,11 @@ cross-fetch@3.1.5:
   dependencies:
     node-fetch "2.6.7"
 
+crypto-js@4.2.0:
+  version "4.2.0"
```
gherceg, line 426: It looks like the sha1.js has version CryptoJS v3.0.2, so is this also an upgrade? And have you looked into what this upgrade entails?

orangejenny: Excellent point, I hadn't noticed. I went to the [release notes](https://www.npmjs.com/package/crypto-js#release-notes) and think it's fine to do the upgrade. The breaking change in 4.0 is for environments that don't have the native crypto module. It's well-supported at this point ([docs](https://developer.mozilla.org/en-US/docs/Web/API/Crypto)).

gherceg (approved):

PR #1155, opened 2025-05-13
Switch from RequireJS to Webpack

orangejenny:
## Technical Summary
https://dimagi.atlassian.net/browse/SAAS-17488

HQ PR: https://github.com/dimagi/commcare-hq/pull/36419

### Safety story
Large change. Relying on QA.

### Automated test coverage
This change affects most parts of Vellum. Overall, test coverage in Vellum is good enough to cover a lot of this PR but probably not every single behavior.

### QA Plan
https://dimagi.atlassian.net/browse/QA-7671

### Rollback instructions

### Labels & Review

### DIFF
--- a/webpack/webpack.common.js
+++ b/webpack/webpack.common.js
@@ -103,7 +103,6 @@ module.exports = {
         alias: {
             'ckeditor': path.resolve(BASE_PATH, 'lib/ckeditor/ckeditor.js'),
             'ckeditor-jquery': path.resolve(BASE_PATH, 'lib/ckeditor/adapters/jquery.js'),
-            'CryptoJS': path.resolve(BASE_PATH, 'lib/sha1'),
             'diff-match-patch': path.resolve(BASE_PATH, 'lib/diff_match_patch'),
             'save-button': path.resolve(BASE_PATH, 'lib', 'SaveButton.js'),
             'jquery.vellum': path.resolve(BASE_PATH, 'src', 'main'),


webpack/webpack.common.js
```diff
@@ -0,0 +1,112 @@
...
+    },
+    module: {
+        rules: [
+            {
+                test: /\.xml$/,
+                type: 'asset/source',
+            },
+            {
+                test: /\.html$/,
+                type: 'asset/source',
```
orangejenny, line 27: I'd prefer to use something like `underscore-template-loader` here, but it breaks a number of tests in ways that I haven't yet figured out. I don't want to block this PR on resolving those.

MartinRiese (approved):

webpack/webpack.common.js
```diff
@@ -0,0 +1,116 @@
...
+    },
+    module: {
+        rules: [
+            {
+                test: /\.xml$/,
+                type: 'asset/source',
+            },
+            {
+                test: /templates\/.*\.html$/,
+                loader: path.resolve('src/template-loader.js'),
```
millerdev, line 27: Now that we've fixed the issues in tests, could this be replaced with `underscore-template-loader`?

src/dateformats.js
```diff
@@ -1,8 +1,10 @@
 define([
     'jquery',
-    'tpl!vellum/templates/date_format_menu'
+    'underscore',
+    'vellum/templates/date_format_menu.html'
 ], function(
     $,
+    _,
```
millerdev, line 7: I don't see where `_` is referenced in this file. Is it a template vestige?

millerdev (approved):
Looking great! :trophy: 

Any chance you'd consider squashing https://github.com/dimagi/Vellum/pull/1155/commits/994aa3f04863268daf5e45db8720edd9584744c0 and https://github.com/dimagi/Vellum/pull/1155/commits/62666d1861392fba580a3869729aadc145acade8 into a single commit? I know it would mean a rebase, but I think it would clean up the PR nicely.

src/dateformats.js
orangejenny: 21c46756

src/core.js
```diff
@@ -2318,8 +2319,7 @@ define([
 
         if (saveType === 'patch') {
             checkForConflict = true;
-            var diff_match_patch = require('diff-match-patch'),
```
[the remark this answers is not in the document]
orangejenny: Given that this library is already included in [src/util](https://github.com/dimagi/Vellum/blob/e9c0f920ce1807cf9f95700076e4268a49fb52db/src/util.js#L5), I don't believe that this code will affect load time.

src/dateformats.js
orangejenny: 21c46756

orangejenny:
Rebased as requested.

millerdev (approved):

PR #1157, opened 2025-05-16
Riese/rm ckditor 2

MartinRiese:
### Rollback instructions

- [ ] This PR can be reverted after deploy with no further considerations

### Labels & Review
- [ ] Risk label is set correctly
- [ ] The set of people pinged as reviewers is appropriate for the level of risk of the change


src/richText.js
```diff
@@ -204,34 +204,27 @@ define([
         input.promise = new Promise((resolve) => {
           resolveEditorPromise = resolve;
         });
-        var NOTSET = {};
-        var newval = NOTSET;  // HACK work around async get/set
         const wrapper = {
             getValue: function (callback) {
```
millerdev, line 208: Is the `callback` parameter still needed since `getValue` and `setValue` are no longer async?

The same question applies to `setValue` below.

MartinRiese: I could only find references to this in the test code. But it is proving to be tricky to remove it. So I started a follow up [PR](https://github.com/dimagi/Vellum/pull/1168) for it.

tests/richText.js
```diff
@@ -765,19 +606,6 @@ define([
                     util.assertXmlEqual(call('createXML'), BURPEE_XML);
                 });
 
-                it("cursor should be at end of input on focus", function () {
```
millerdev, line 768: I saw some issues related to cursor position at the end of the input when I was playing around on staging. Would it be possible to adapt this test to the new editor?

tests/richText.js
```diff
@@ -883,18 +715,129 @@ define([
...
+                                $(".popover").remove();
+                            }
+                        });
+                        setTimeout(function() {
+                            bubble.mouseenter();
+                        }, editorUpDelay);
+                    });
+                });
+
+                // it("should expand tree on click show in question list", function (done) {
```
millerdev, line 800: Is this intended to be uncommented or removed?

src/undo.js
millerdev, line 1: This seems more elaborate than [undomanager.js](https://github.com/dimagi/Vellum/blob/6d90688bd943e56560596dd411d6d494c034a86a/src/undomanager.js), but looks like it does something similar. Do we need both? Probably out of scope to use this nice new `UndoStack` in places that currently use `UndoManager`...?

src/richText.js
```diff
@@ -917,24 +943,27 @@ define([
...
                         x = pos.left,
                         y = pos.top + $(this).height();
                     $("#" + dateFormatID).click(function () {
-                        $imgs.popover('hide');
+                        $element.popover('hide');
                         dateformats.showMenu(x, y, function (format) {
-                            $this.attr("data-date-format", format);
-                            editor.fire("saveSnapshot");
+                            $widget.attr("data-date-format", format);
+                            // editor.fire("saveSnapshot");
```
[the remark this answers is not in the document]
MartinRiese: yes. you are right

tests/undo.js
```diff
@@ -0,0 +1,147 @@
...
+define([
+    'chai',
+    'vellum/undo'
+], function (
+    chai,
+    undo
+) {
+    const assert = chai.assert;
+
+    describe("The UndoStack", function () {
```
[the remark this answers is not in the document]
MartinRiese: thanks

src/richText.js
```diff
@@ -305,66 +331,65 @@ define([
...
-    var realDataTransfer = CKEDITOR.plugins.clipboard.dataTransfer;
-    richTextDataTransfer.prototype = realDataTransfer.prototype;
-    CKEDITOR.plugins.clipboard.dataTransfer = richTextDataTransfer;
 
-    /**
-     * Set selection in CKEditor
-     */
-    function ckSelect(editor, index, length) {
+        input.data("ckwrapper", wrapper);
+        resolveEditorPromise(); // probably can just set it to a resolved promise to beging with or remove the code that waits for it.
```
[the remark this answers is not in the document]
MartinRiese: Instead of resolving it here. I could just created on that is already resolved. But I think I will just leave it as is and remove the comment.

src/undo.js
```diff
@@ -0,0 +1,188 @@
...
+         *                                               of adding new items
+         * @returns {boolean} - true if the push was successful, false if ignored due to
+         *                     an ongoing update operation
+         */
+        push: function (item, accumulateUntilPause = true) {
+            if (this.isUpdating) {
+                return false;
+            }
+            if (accumulateUntilPause && this.undoItems.length > 0 &&
+                this.lastPushTime !== null && Date.now() - this.lastPushTime < PUSH_INTERVAL) {
```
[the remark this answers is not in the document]
Jtang-1: oh ya i missed that conditional check

src/richText.js
```diff
@@ -136,15 +142,26 @@ define([
...
+                    range.setStartAfter(leadingSpaceNode);
+                }
+
+                const fragment = htmlToFrament(content);
+                range.insertNode(fragment);
+                range.collapse(false);
+
+                if (insertSpaces) {
+                    const trailingSpaceNode = document.createTextNode(ZERO_WIDTH_SPACE);
+                    range.insertNode(trailingSpaceNode);
```
Jtang-1, line 158: if we also insert space at the end, then couldn't this result in two spaces between elements?

src/richText.js
```diff
@@ -56,6 +56,22 @@ define([
 ){
     var FORM_REF_REGEX = /^#form\//;
     var INVALID_PREFIX = "#invalid/xpath ";
+    var ZERO_WIDTH_SPACE = "\u200B";
+
+    function htmlToFrament(html) {
```
Jtang-1, line 61: nit: typo

src/richText.js
```diff
@@ -196,7 +213,90 @@ define([
...
+                    }
+                }
+                if (nextNode && nextNode.nodeType === 3 && nextNode.nodeValue.startsWith(ZERO_WIDTH_SPACE)) {
+                    nextNode.nodeValue = nextNode.nodeValue.slice(1);
+                    if (nextNode.length === 0) {
+                        nextNode.remove();
+                    }
+                }
+
+                console.log('span removed');
```
Jtang-1, line 297: there's a couple print statements in here that can be deleted

src/richText.js
MartinRiese: Yes. that is by design. Always having one space to the left and one to the right makes the logic for removing and moving the cursor easier.

src/undo.js
MartinRiese: I could see the UndoManager using the new undo stack. But it is tied to the mugs which made is hard for me to reuse.

src/less-style/inputs.less
```diff
@@ -33,7 +33,7 @@
         box-shadow: inset 0 1px 1px rgba(0,0,0,.075), 0 0 8px rgba(82,168,236,.6);
     }
     p {
-        margin: 0; /* CKEditor line spacing */
+        margin: 0;
```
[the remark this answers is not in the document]
MartinRiese: removed

tests/richText.js
MartinRiese: Not sure anymore. It is passing. putting it back in

src/undo.js
```diff
@@ -156,6 +176,22 @@ define([], function () {
...
+                value = v;
+            });
+            return value;
+        },
+        reset: function () {
+            this.undoStack.peek((item) => {
+                if (item !== null) {
+                    this.element.innerHTML = item.value;
+                    setCursorPosition(this.element, item.cursor);
+                    // sendInputEvent(this.element);
```
millerdev, line 191: Can be removed?

tests/richText.js
```diff
@@ -350,9 +350,9 @@ define([
         describe("convert value with output and escaped HTML", function () {
             var items = [
                     ['<h1><output value="#form/text" /></h1>',
-                     '&lt;h1&gt;{text}&lt;/h1&gt;'],
+                     '&lt;h1&gt;​{text}​&lt;/h1&gt;'],
                     ['<output value="#form/text" /> <tag /> <output value="#form/othertext" />',
-                     '{text} &lt;tag /&gt; {othertext}'],
+                     '​{text}​ &lt;tag /&gt; ​{othertext}​'],
```
millerdev, line 352: Would it make sense to add comments on these lines to let future readers know they contain ZWSP characters before and after `{...}`?

tests/richText.js
millerdev: I think it's conventional for the cursor to be either at the place where it last was or at the end. At the beginning seems less than idea, but I'm not going to block the PR on this issue.

millerdev (commented):
I didn't get a chance yet (but am planning) to test drive it on staging, but this is looking fantastic! Thanks for all of the work you've put into it. :medal_sports:

src/richText.js
```diff
@@ -196,7 +213,90 @@ define([
...
         inputElement.addEventListener('input', function(e) {
+            const nonEditableSpans = inputElement.querySelectorAll('span[contenteditable="false"]');
+            const spansToRemove = [];
+
+            nonEditableSpans.forEach(span => {
+                const prevNode = span.previousSibling;
+                const nextNode = span.nextSibling;
+
+                const zwspBeforeMissing = !prevNode || prevNode.nodeType !== 3 || !prevNode.nodeValue.endsWith(ZERO_WIDTH_SPACE);
+                const zwspAfterMissing = !nextNode || nextNode.nodeType !== 3 || !nextNode.nodeValue.startsWith(ZERO_WIDTH_SPACE);
```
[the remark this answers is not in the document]
MartinRiese: yes. replaced it with Node.TEXT_NODE

tests/richText.js
MartinRiese: Fair. Added

tests/databrowser.js
```diff
@@ -145,9 +145,9 @@ define([
                     calc = $("[name=property-calculateAttr]"),
                     sessionUri = CASE_DATA[0].uri,
                     casedbUri = CASE_DATA[1].uri,
-                    editor = calc.ckeditor().editor,
-                    widget = util.getWidget('property-calculateAttr');
-                widget.input.promise.then(function () { 
+                    widget = util.getWidget('property-calculateAttr'),
+                    editor = widget.input.data("ckwrapper");
```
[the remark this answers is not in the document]
MartinRiese: replaced with "editorWrapper"

tests/atwho.js
```diff
@@ -186,7 +186,7 @@ define([
             }
 
             it("should show questions with #form", function() {
-                displayAtwho('#form', function(mug) {
+                displayAtwho('#form\u200B', function(mug) {
```
millerdev: Why did this change? Is it related to placing the cursor at the end on focus?

MartinRiese: Yes. I added a focus handler that moves the cursor one before the last character which should always be a ZWSP now. In the test it ended up putting it between the r and m of form causing the list of tags to be different. Adding it here just makes sure the text also ends with ZWSP and put the cursor after the m resulting it the right set of tags.

millerdev (commented):
I was ready to approve, but then noticed removed spaces in _"should insert expression into expression at ..."_ tests. If you can answer that I think it's good to go.

PR #1158, opened 2025-05-20
Make sure that yarn is built properly before running tests

orangejenny:
## Technical Summary
Followup for https://github.com/dimagi/Vellum/pull/1155

I realized that depending on which config of webpack you've most recently run with, running tests may fail with an error about require not being defined. This PR makes sure that the proper build files are available when you run via either the `test` script or the Makefile.

### Safety story
It's all about tests.

### Automated test coverage

It's tests.

### QA Plan

No

### Rollback instructions

### Labels & Review

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,4 +25,5 @@ _tar:
 	tar -czf vellum.tar.gz -C _build .
 
 _test:
+	yarn testbuild
 	npm test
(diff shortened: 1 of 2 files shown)


MartinRiese (approved):

orangejenny:
Noting that this one doesn't need to be "released" to HQ because it doesn't affect any application code. It should, however, fix problems with running `vellum-to-staging` on HQ.

PR #1176, opened 2025-10-03
Add support for HTML messages

mjriley:
## Product Description
Should have no direct effect yet, just adds backend support for HTML messages.

## Technical Summary
This PR adds HTML support for mug messages. Now, if a message object is added with an `html` property, that value will not be escaped, and thus will preserve any HTML. If the message object is instead a string (as its been previously) then the string is escaped before being displayed.

The intent behind this change was to allow us to include help links in message text, although this is not yet used.

## Feature Flag
None

### Safety story

Verified locally with custom data. Existing behavior (with normal strings) is preserved, and no code uses the new functionality yet, so the impacts of this change should be minimal.

### Automated test coverage

Added two new tests in _core.js_

### QA Plan

No QA

### Rollback instructions

### Labels & Review

### DIFF
--- a/src/mugs.js
+++ b/src/mugs.js
@@ -671,7 +671,7 @@ define([
             return _.isEmpty(this.messages);
         },
         getMessageText: function(message) {
-            return (message && message.hasOwnProperty("html")) ? message.html : message;
+            return (message && message.hasOwnProperty("markdown")) ? message.markdown: message;
         }
     };
 
--- a/src/widgets.js
+++ b/src/widgets.js
@@ -805,8 +805,8 @@ define([
             if (seen.hasOwnProperty(messageKey)) { return; }
             seen[messageKey] = true;
             let htmlMessage = "";
-            if (msg.message.hasOwnProperty("html")) {
-                htmlMessage = msg.message.html;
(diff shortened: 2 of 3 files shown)


millerdev (commented):
Looks like this caused a bunch of tests to fail, which may be able to be fixed relatively easily (not sure).

However, can we use markdown links instead of supporting arbitrary HTML? I thought there were security issues such as XSS if we allow arbitrary untrusted HTML. Or maybe I'm missing something and the source of the HTML is always trusted?

mjriley:
> Looks like this caused a bunch of tests to fail, which may be able to be fixed relatively easily (not sure).
> 
> However, can we use markdown links instead of supporting arbitrary HTML? I thought there were security issues such as XSS if we allow arbitrary untrusted HTML. Or maybe I'm missing something and the source of the HTML is always trusted?

Thanks for the suggestion. At least for now, I only need links, so I've changed over the functionality to markdown to keep us a bit more secure. I believe all feedback should be addressed. I was running all tests, but I'm guessing Chrome was using stale JS, and so the tests were using the old code and still passing.

millerdev (approved):

PR #1214, opened 2026-03-27
Turn Case Type into a required dropdown populated with existing case types

jingcheng16:
## Product Description

The Case Type field is now a dual-mode widget: a dropdown for static case types, or an XPath field for dynamic case type expressions.

- Dropdown mode (default): Populated with existing case types from the data dictionary. Create actions allow typing new case types (spaces auto-converted to underscores). Input is validated, and reserved case types (`commcare-user` for non-update actions, `user-owner-mapping-case`) are rejected.

- XPath mode: A xpath widget supports dynamic expressions like `cond()` that compute case types at runtime. 

Toggle between modes via "Use an xpath expression" / "Use a dropdown" links below the field.

Behavior on load:

| create/case_type bind value | Mode shown | Bind on save |
|--------|--------|--------|
| 'household' (quoted literal) | Dropdown: household | 'household' |
| "household" (double-quoted) | Dropdown: household | 'household' |
| /data/case_type_val (xpath) | XPath: /data/case_type_val | /data/case_type_val |
| cond(...) (dynamic xpath) | XPath: cond(...) | cond(...) |
| worker_role (bare word) | XPath: worker_role (prompt user to fix) | worker_role | 
| empty | Dropdown: falls back to vellum:case_type | from dropdown

Mode switching:
- Switching saves the current value and restores it when switching back
- XPath mode is only available when Create action is active
- Custom case types created via tags are removed from dropdown options when Create is deselected

Before
Screenshot 2026-03-27 at 11 55 08 AM

After
https://github.com/user-attachments/assets/53e219f8-f8cc-42fe-85e6-4820c8be102d

The generated XML remains identical to before — no changes to the form structure.
 
## Technical Summary

Ticket: https://dimagi.atlassian.net/browse/SAAS-19459

Upon investigation, ~137 out of 15,309 create action save to case questions with `create/case_type` use xpath expressions (not literals). These include dynamic patterns like `cond()` that compute case types based on form data. See https://docs.google.com/spreadsheets/d/1_8fXNNN1L9dda5PnugTwNC40ZpjTgq1MhoPYNGk1ceI/edit?gid=1491268561#gid=1491268561

Architecture — two spec properties, one UI slot:
- `case_type` (dropdown) — static case type string, stored in `vellum:case_type` and `create/case_type` bind as a quoted literal
- `caseTypeXPath` (xpath widget) — dynamic xpath expression, stored directly in `create/case_type` bind
- Only one is visible at a time; toggle links switch between them
- Validation is cross-linked: each field passes when the other has a value

Key changes:
- `parseBindElement`: Quoted literals (`'household'`) → `case_type` (dropdown). Non-quoted expressions → `caseTypeXPath` (xpath field), clears `case_type`
- `caseTypeDropdownWidget`: select2 dropdown with tags for Create, "Use an xpath expression" toggle link, removes custom tags when Create is deselected
- `caseTypeXpathWidget`: wraps widgets.xPath, "Use a dropdown" toggle link, validates bare words as likely literals
- `switchToXpathMode` / `switchToDropdownMode`: save/restore values when toggling, cross-validate both fields
- `getBindList` / `dataChildFilter` / `getCaseSaveData`: generate case_type from whichever field is active
- Case type validation: [\w-]+ regex, `commcare-user` restricted to update-only, `user-owner-mapping-case` fully reserved

[HQ change ](https://github.com/dimagi/commcare-hq/pull/37533) that passes `existingCaseTypes` to the `saveToCase` plugin options will be deployed independently before this — sending extra options that Vellum doesn't consume yet is harmless.

## Feature Flag

`VELLUM_SAVE_TO_CASE`

### Safety story

- The generated XML is identical to before — only the UI for entering `case_type` changed
- Existing forms with literal case types load and save identically
- Existing forms with xpath expressions now display in the xpath field instead of being dropped — strictly better backward compatibility
- If an old form has different values for the top section's Case Type field and the Create section's Case Type property, the new form's top section's Case Type field will use the bind value from `/create/case_type` 
  
### Automated test coverage

 - Existing round-trip tests (create_property.xml, create_2_property.xml, logic_test.xml) verify XML is unchanged
 - New tests "create/case_type backward compatibility" added
### QA Plan

- Load form with literal `create/case_type` → dropdown shown with value
- Load form with xpath `create/case_type` → xpath field shown with expression
- Toggle between modes → values saved and restored
- Create new case type, deselect Create → custom type removed from dropdown
- Type bare word in xpath field → validation warning about literal value
- Save in each mode → correct XML output
- `commcare-user` with Create → error; with Update only → allowed

### Rollback instructions

### Labels & Review

### DIFF
--- a/src/saveToCase.js
+++ b/src/saveToCase.js
@@ -157,6 +157,14 @@ function caseTypeDropdownWidget(mug, opts) {
 
     function initSelect2() {
         var value = widget.input.val();
+        // When Create is turned off, remove custom case types that
+        // aren't in the data dictionary — they were only valid for creation.
+        if (!createsCase(mug) && value && !_.contains(existingCaseTypes, value)) {
+            widget.input.find('option[value="' + value + '"]').remove();
+            widget.input.val('');
+            value = '';
+            mug.p.case_type = '';
+        }
         if (widget.input.data('select2')) {
             widget.input.select2('destroy');
         }
@@ -336,7 +344,7 @@ var slugToProp = {
                     if (!val && createsCase(mug)) {
                         return gettext("Case Type is required");
                     }
-                    if (!CASE_TYPE_REGEX.test(val)) {
+                    if (val && !CASE_TYPE_REGEX.test(val)) {
                         return gettext("Case types can only include the characters a-z, 0-9, '-' and '_'");
                     }
                     if (val === 'commcare-user') {
--- a/src/saveToCase.js
+++ b/src/saveToCase.js
@@ -353,8 +353,12 @@ var slugToProp = {
                     if (mug.p.case_type || !createsCase(mug)) {
                         return 'pass';
                     }
-                    if (!mug.p.caseTypeXPath) {
+                    var val = mug.p.caseTypeXPath;
+                    if (!val) {
                         return gettext("Case Type is required");
+                    }
+                    if (CASE_TYPE_REGEX.test(val)) {
+                        return gettext("This looks like a literal value. Use the dropdown instead, or wrap it in quotes like '" + val + "'.");
                     }
                     return 'pass';
                 },
--- a/src/saveToCase.js
+++ b/src/saveToCase.js
@@ -148,10 +148,7 @@ function caseTypeDropdownWidget(mug, opts) {
         if (val && /\s/.test(val)) {
             widget.setValue(val.replace(/\s/g, '_'));
         }
-        // User chose a value directly; drop any stashed xpath reference
-        mug.p.caseTypeXPath = null;
         super_updateValue();
-        // Re-validate the xpath field since its validation depends on case_type
         mug.validate('caseTypeXPath');
     };
 
--- a/src/saveToCase.js
+++ b/src/saveToCase.js
@@ -373,7 +373,15 @@ var slugToProp = {
 
             var actions = [];
             if (createsCase(mug)) {
-                actions.push(simpleNode('create', makeColumns(mug.p.createProperty)));
+                // Include a case_type column in the create action subtree when mug.p.case_type
+                // is set so that the tree shape matches legacy forms that has case_type
+                // under create section alongside other properties.
+                var createProps = {};
+                if (mug.p.case_type) {
+                    createProps.case_type = {};
+                }
+                _.extend(createProps, mug.p.createProperty);
+                actions.push(simpleNode('create', makeColumns(createProps)));
             }
 
             if (updatesCase(mug)) {
@@ -422,6 +430,14 @@ var slugToProp = {
                     ret = ret.concat({
                         nodeset: mug.absolutePath + "/case/@case_id",
                         calculate: mug.p.case_id
+                    });
+                }
+                // Emit /case/create/case_type from mug.p.case_type so XForm matches legacy
+                // save-to-case layout (create subtree) while the UI keeps Case Type top-level.
+                if (mug.p.case_type) {
+                    ret.push({
+                        nodeset: mug.absolutePath + "/case/create/case_type",
+                        calculate: mug.p.case_type
                     });
                 }
                 ret = ret.concat(generateBinds('create', mug.p.createProperty));
@@ -634,11 +650,20 @@ $.vellum.plugin("saveToCase", {}, {
                 if (mug && mug.__className === "SaveToCase") {
                     if (matchRet[2]) {
                         var prop = matchRet[2],
-                            pKey = {
+                            action = matchRet[1];
+
+                        // Skip create/case_type bind — case_type is set by parseDataNode
+                        // from vellum:case_type. We still consume the bind to prevent it
+                        // from going into createProperty.
+                        if (action === "create" && prop === "case_type") {
+                            return;
+                        }
+
+                        var pKey = {
                                 create: "createProperty",
                                 update: "updateProperty",
                                 index: "indexProperty",
-                            }[matchRet[1]];
+                            }[action];
 
                         if (!mug.p[pKey]) {
                             mug.p[pKey] = {};
--- a/src/saveToCase.js
+++ b/src/saveToCase.js
@@ -150,7 +150,7 @@ function caseTypeDropdownWidget(mug, opts) {
     // Keep parsed Case Type values visible in the dropdown even when they are
     // not present in configured case type options.
     if (mug.p.case_type && !_.contains(existingCaseTypes, mug.p.case_type)) {
-        existingCaseTypes = existingCaseTypes.concat([mug.p.case_type]);
+        existingCaseTypes.push(mug.p.case_type);
     }
     opts.defaultOptions = existingCaseTypes.map(function (ct) {
         return { text: ct, value: ct };
--- a/src/saveToCase.js
+++ b/src/saveToCase.js
@@ -145,6 +145,8 @@ var propertyWidget = function (mug, options) {
         return widget;
     };
 
+var CASE_TYPE_REGEX = /^[\w-]+$/;
+
 function caseTypeDropdownWidget(mug, opts) {
     var existingCaseTypes = opts.vellum.data.saveToCase?.existingCaseTypes || [];
     // Keep parsed Case Type values visible in the dropdown even when they are
@@ -243,6 +245,21 @@ var slugToProp = {
                 visibility: 'visible',
                 presence: 'required',
                 widget: caseTypeDropdownWidget,
+                validationFunc: function (mug) {
+                    var val = mug.p.case_type;
+                    if (val && !CASE_TYPE_REGEX.test(val)) {
+                        return gettext("Case types can only include the characters a-z, 0-9, '-' and '_'");
+                    }
+                    if (val === 'commcare-user') {
+                        if (createsCase(mug) || closesCase(mug) || indexesCase(mug)) {
+                            return gettext("'commcare-user' cases can only be updated, not created, closed, or linked");
+                        }
+                    }
+                    if (val === 'user-owner-mapping-case') {
+                        return gettext("This is a reserved case type. Please choose another name.");
+                    }
+                    return 'pass';
+                },
             },
             "case_id": {
                 lstring: gettext("Case ID"),
--- a/src/saveToCase.js
+++ b/src/saveToCase.js
@@ -673,10 +673,14 @@ $.vellum.plugin("saveToCase", {}, {
                         var prop = matchRet[2],
                             action = matchRet[1];
 
-                        // Skip create/case_type bind — case_type is set by parseDataNode
-                        // from vellum:case_type. We still consume the bind to prevent it
-                        // from going into createProperty.
+                        // Consume /case/create/case_type into the top-level Case Type field.
+                        // Prefer a non-empty bind over vellum:case_type
                         if (action === "create" && prop === "case_type") {
+                            var caseTypeBindValue = el.xmlAttr("calculate") || '',
+                                stripped = caseTypeBindValue.replace(/^'(.*)'$/, '$1');
+                            if (stripped) {
+                                mug.p.case_type = stripped;
+                            }
                             return;
                         }
 
--- a/src/saveToCase.js
+++ b/src/saveToCase.js
@@ -550,6 +550,12 @@ var slugToProp = {
     };
 
 $.vellum.plugin("saveToCase", {}, {
+    init: function () {
+        var opts = this.opts().saveToCase || {};
+        this.data.saveToCase = {
+            existingCaseTypes: opts.existingCaseTypes || [],
+        };
+    },
     getAdvancedQuestions: function () {
         return this.__callOld().concat(["SaveToCase"]);
     },
--- a/src/saveToCase.js
+++ b/src/saveToCase.js
@@ -151,6 +151,8 @@ function caseTypeDropdownWidget(mug, opts) {
         // User chose a value directly; drop any stashed xpath reference
         mug.p.caseTypeXPath = null;
         super_updateValue();
+        // Re-validate the xpath field since its validation depends on case_type
+        mug.validate('caseTypeXPath');
     };
 
     function initSelect2() {
@@ -210,6 +212,8 @@ function caseTypeXpathWidget(mug, opts) {
         var val = $.trim(widget.getValue());
         mug.p.caseTypeXPath = val || null;
         super_updateValue();
+        // Re-validate the dropdown since its validation depends on caseTypeXPath
+        mug.validate('case_type');
     };
 
     widget.postRender = function () {
@@ -252,6 +256,8 @@ function switchToXpathMode(mug, dropdownWidget, $dropdownRow) {
 
     $dropdownRow.hide();
     $xpathRow.show();
+    mug.validate('case_type');
+    mug.validate('caseTypeXPath');
 }
 
 function switchToDropdownMode(mug, xpathWidget, $xpathRow) {
@@ -277,6 +283,8 @@ function switchToDropdownMode(mug, xpathWidget, $xpathRow) {
 
     $xpathRow.hide();
     $dropdownRow.show();
+    mug.validate('case_type');
+    mug.validate('caseTypeXPath');
 }
 
 var slugToProp = {
--- a/src/saveToCase.js
+++ b/src/saveToCase.js
@@ -841,8 +841,9 @@ $.vellum.plugin("saveToCase", {}, {
                                 stripped = caseTypeBindValue.replace(/^(['"])(.*)\1$/, '$2');
                             if (stripped && stripped === caseTypeBindValue) {
                                 // No quotes stripped — xpath expression;
-                                // stash for resolution in handleMugParseFinish
+                                // show in xpath field, override vellum:case_type
                                 mug.p.caseTypeXPath = caseTypeBindValue;
+                                mug.p.case_type = '';
                             } else if (stripped) {
                                 // Route create/case_type to the Case Type dropdown.
                                 mug.p.case_type = stripped;
--- a/src/saveToCase.js
+++ b/src/saveToCase.js
@@ -33,15 +33,15 @@ function usesCases(mug) {
 
 // Try to resolve an xpath expression to a string literal.
 // Handles:
-//   - Quoted strings:  'household'  →  household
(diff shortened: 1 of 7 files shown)


src/saveToCase.js
```diff
@@ -673,10 +673,14 @@ $.vellum.plugin("saveToCase", {}, {
...
                             action = matchRet[1];
 
-                        // Skip create/case_type bind — case_type is set by parseDataNode
-                        // from vellum:case_type. We still consume the bind to prevent it
-                        // from going into createProperty.
+                        // Consume /case/create/case_type into the top-level Case Type field.
+                        // Prefer a non-empty bind over vellum:case_type
                         if (action === "create" && prop === "case_type") {
+                            var caseTypeBindValue = el.xmlAttr("calculate") || '',
+                                stripped = caseTypeBindValue.replace(/^'(.*)'$/, '$1');
```
kaapstorm: I think this line strips surrounding `'` from literals in the Calculate field. But are there any instances where the Calculate field has a calculation / XPath expression to set the case type?

Is there a possibility that we could accidentally set the name of a case type to an XPath expression? In which case, should we be evaluating `el.xmlAttr("calculate")` instead of trusting that it is always a string literal?

millerdev (commented):
Reviewed up to 4f3558ec8b341433ef40c39f860e3efb8ba7a1be. Will continue with the rest later.

src/saveToCase.js
```diff
@@ -373,7 +373,15 @@ var slugToProp = {
 
             var actions = [];
             if (createsCase(mug)) {
-                actions.push(simpleNode('create', makeColumns(mug.p.createProperty)));
+                // Include a case_type column in the create action subtree when mug.p.case_type
+                // is set so that the tree shape matches legacy forms that has case_type
+                // under create section alongside other properties.
```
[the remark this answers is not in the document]
jingcheng16: I intentionally ask AI to add those comment and have rephrased them myself make sure they conveyed what I tried to convey. I think the way we handle case_type might be confusing for people who doesn't know the history of case_type fields and I hope those comments would be helpful for those people.

millerdev (commented):
LGTM. Curious about the [investigation on Norman's question](https://github.com/dimagi/Vellum/pull/1214#discussion_r3011395060).

jingcheng16:
@millerdev and @kaapstorm 
Thanks Norman for raising a crucial question, after pulling data from production I did find a bunch of stc questions refer a hidden value as case type. I pushed some new changes and also updated PR description to include this.
It took me a while to think about how to deal with each scenario.
I have more changes I want to make, and will ping both of you again when this PR is ready for review again. Thank you!

jingcheng16:
@millerdev and @kaapstorm This is ready for review again!

src/saveToCase.js
```diff
@@ -471,7 +471,7 @@ var slugToProp = {
                 if (mug.p.case_type) {
                     ret.push({
                         nodeset: mug.absolutePath + "/case/create/case_type",
-                        calculate: mug.p.case_type
+                        calculate: "'" + mug.p.case_type + "'"
```
millerdev: Is it always true that `mug.p.case_type` will not contain `'` character(s)? I assume so, but wanted to make sure you thought about it.

src/saveToCase.js
```diff
@@ -31,6 +31,25 @@ function usesCases(mug) {
...
 
+// Try to resolve an xpath expression to a string literal.
+// Handles:
+//   - Quoted strings:  'household'  →  household
+//   - Path references:  /data/q     →  follow the question's calculateAttr
+function resolveToLiteral(form, expr) {
+    if (!expr) {
+        return null;
+    }
+    var quotedMatch = expr.match(/^'([^']*)'$/);
```
millerdev: I believe XPath expressions support single quotes (`'`) or double quotes (`"`) as string delimiters, but this will only work for single quotes. Is that intentional?

src/saveToCase.js
```diff
@@ -652,6 +674,15 @@ $.vellum.plugin("saveToCase", {}, {
                 return value === inner;
             });
         }
+        if (mug.p._caseTypeCalc) {
+            var resolved = resolveToLiteral(mug.form, mug.p._caseTypeCalc);
+            if (resolved) {
+                mug.p.case_type = resolved;
+            } else {
+                // drop the stashed xpath reference if it's not resolvable
+                mug.p._caseTypeCalc = null;
```
millerdev: This is starting to feel complicated and a bit hacky; especially the part with `mug.p._caseTypeCalc`. Can you explain in what scenarios `mug.p._caseTypeCalc` will have a value? Are you confident that the value will never get out of sync with the value in `mug.p.case_type`?

src/saveToCase.js
```diff
@@ -31,6 +31,25 @@ function usesCases(mug) {
...
+function resolveToLiteral(form, expr) {
+    if (!expr) {
+        return null;
+    }
+    var quotedMatch = expr.match(/^'([^']*)'$/);
+    if (quotedMatch) {
+        return quotedMatch[1] || null;
+    }
+    var refMug = form.getMugByPath(expr);
+    if (refMug && refMug.p.calculateAttr) {
```
millerdev: It seems wrong to discard whatever value is in `expr` if it does not resolve to a mug path. What if the user mis-typed the mug path? It seems like a bad UX to throw away the wrong value they typed.

src/saveToCase.js
jingcheng16: `'` is a special character, it is impossible to save a case type with `'`, it won't pass the validation

src/saveToCase.js
jingcheng16: ah this is my oversight. I thought it is impossible to have double quotes, because the `calculate` attr of `bind` element already has double quotes around the value. But I just tested it, for double quote value, it will become   `<bind calculate="&quot;case_type&quot;" />`. Fixed in e252633

src/saveToCase.js
jingcheng16: `_caseTypeCalc` has a value when the `create/case_type` bind contains an xpath reference (e.g. `/data/case_type_val`) rather than a quoted literal. This is a pattern where a hidden value question holds the actual case type, like:
```
<bind nodeset="/data/case_type_val" calculate="'household'" />                                                                                                                                               
<bind nodeset="/data/q1/case/create/case_type" calculate="/data/case_type_val" />
```
  At parse time, we resolve `/data/case_type_val` → follow the reference → find 'household' → set `mug.p.case_type` = "household" (for the dropdown) and `mug.p._caseTypeCalc` = "/data/case_type_val" (to preserve the original reference in the bind on save). Store the reference in `_caseTypeCalc` is intentional so if user go to `case_type_val` question to change case type, it will reflected at run time.
I'm aware that the two values could get out of sync. When user changes the hidden value's calculate during the same session, the dropdown would still show the old resolved value. But the dropdown's value is only used to populate case metadata, which means it only impacts the Case Summary/Form Summary page, so I think it is not a big deal if they go out of sync. In our current implementation, because there is no validation for both `mug.p.case_type` and `/create/case_type`, they can still be different. 

`mug.p.case_type` re-syncs on form reload (parse runs again).
                                                                                                                                                                                                               
`_caseTypeCalc` is cleared when the user selects a new value from the dropdown. At that time, we will never need `_caseTypeCalc` anymore.

Sorry I know it is complicated but I cannot think of a better way to handle this.

src/saveToCase.js
jingcheng16: I discard the value if it is not a literal value, and is not a mug path. Because those value would be invalid. Also based on the data I pulled from production, `/create/case_type` can be empty, literal, a mug path, or bare words without quotes around it which is invalid and will evaluate to empty at run time. So the unresolvable expression would be broken regardless...why it is a bad UX to throw away? Discarding those will prompt user to select a valid case type when the form is opened. But if the form is never opened, then the xml is untouched and will remain the same.

src/saveToCase.js
millerdev: I believe APIs or something like that allow setting case types without validation, and therefore some case types exist that contain special characters. I noticed this in a recent investigation for work I was doing on the Data Dictionary.

Not sure if we should try to support them here, but wanted to make sure you know about it.

src/saveToCase.js
millerdev: > why it is a bad UX to throw away?

Because it is discarding some work that was done by the user. It may be a seemingly trivial amount of work, but whatever they typed could be as simple as a reminder/note to come back and fix later. If it is discarded their note is lost and they may forget to go back and fix it.

A better UX may be to display a validation error rather than modifying/discarding user input.

src/saveToCase.js
millerdev: I'm not understanding how much control the user has over the value of `mug.p.case_type` (and possibly `mug.p._caseTypeCalc`). Can `mug.p.case_type` can contain any arbitrary XPath expression typed by the user, or is it limited to only literals and mug paths that can only be selected from a controlled list, for example?

src/saveToCase.js
```diff
@@ -31,6 +31,25 @@ function usesCases(mug) {
...
 
+// Try to resolve an xpath expression to a string literal.
+// Handles:
+//   - Quoted strings:  'household' or "household"  →  household
+//   - Path references:  /data/q     →  follow the question's calculateAttr
+function resolveToLiteral(form, expr) {
+    if (!expr) {
+        return null;
+    }
+    var quotedMatch = expr.match(/^(['"])([^'"]*)\1$/);
```
millerdev: Is there a test that covers double-quoted case type?

jingcheng16: Just added in 87c4626

src/saveToCase.js
jingcheng16: Thanks for flagging. If it's invalid, we should ask the user to fix it when user load the form. 😞

src/saveToCase.js
jingcheng16: `mug.p.case_type` is limited to plain string values. It can only be set from the dropdown or by resolving a bind value to a literal during parsing. User cannot type xpath into it.
`mug.p._caseTypeCalc` is never user-editable. It's only set during XML parsing when the original bind contained an xpath reference. When the user selects a new value from the dropdown, `_caseTypeCalc` is cleared. So the user only interacts with the dropdown; `_caseTypeCalc` is purely internal bookkeeping for preserving the original bind expression.

src/saveToCase.js
millerdev: I'd like to put a pin in this and possibly come back to it as a future enhancement, but don't want to block the entire PR on it.

millerdev (approved):

jingcheng16:
@millerdev This is ready to review again. I now changed to the new dual mode and we no longer discard any values. Let me know if you have any questions! Thank you! 💟

src/saveToCase.js
```diff
@@ -126,6 +127,174 @@ var propertyWidget = function (mug, options) {
...
+                return { id: term, text: term };
+            },
+        });
+        widget.input.val(value).trigger('change.select2');
+    }
+
+    widget.postRender = function () {
+        initSelect2();
+        var $dropdownRow = widget.input.closest('.widget'),
+            $toggleLink = addModeToggle($dropdownRow, gettext('Use an xpath expression'), function () {
```
millerdev:
```suggestion
            $toggleLink = addModeToggle($dropdownRow, gettext('Select case type with XPath'), function () {
```

src/saveToCase.js
```diff
@@ -126,6 +127,174 @@ var propertyWidget = function (mug, options) {
...
+        var val = $.trim(widget.getValue());
+        mug.p.caseTypeXPath = val || null;
+        super_updateValue();
+        // Re-validate the dropdown since its validation depends on caseTypeXPath
+        mug.validate('case_type');
+    };
+
+    widget.postRender = function () {
+        var $xpathRow = widget.input.closest('.widget');
+        addModeToggle($xpathRow, gettext('Use a dropdown'), function () {
```
millerdev:
```suggestion
        addModeToggle($xpathRow, gettext('Select case type from a list'), function () {
```

src/saveToCase.js
```diff
@@ -149,7 +149,7 @@ function caseTypeDropdownWidget(mug, opts) {
             widget.setValue(val.replace(/\s/g, '_'));
         }
         // User chose a value directly; drop any stashed xpath reference
-        mug.p._caseTypeCalc = null;
+        mug.p.caseTypeXPath = null;
```
millerdev: AI comment: `case_type` dropdown `updateValue` unconditionally nulls `caseTypeXPath`

> Every dropdown change sets `mug.p.caseTypeXPath = null`, even if the user is in dropdown mode and `caseTypeXPath` is already null. This is mostly harmless, but it also wipes `caseTypeXPath` when the dropdown is programmatically updated during mode switching (e.g., `dropdownWidget.handleChange()` in `switchToDropdownMode` at line 289 triggers `updateValue`). The `_savedCaseTypeXPath` stash at line 272 prevents data loss, but the flow is convoluted and easy to break.

I'm having a hard time deciding if this is valid. Feel free to disregard if you think it is not.

tests/saveToCase.js
millerdev, line 1: AI comment: No test for the xpath-to-dropdown-to-xpath round-trip

> The tests cover parsing and initial visibility, but there's no test that verifies: switch to xpath mode, enter a value, switch back to dropdown, switch back to xpath — and the xpath value is preserved. Given the complexity of the stash/restore logic, this is worth covering.

AI comment: Missing test for generated XML with xpath case type

> The old test "should preserve xpath reference in generated bind" was removed and replaced with a UI visibility test. There's no longer a test verifying that the XML output contains `calculate="/data/case_type_val"` when `caseTypeXPath` is set. This is a regression in test coverage for the serialization path.

Did Claude hallucinate that old test ^^? I can't find it, so this might not be valid.

src/saveToCase.js
```diff
@@ -126,6 +127,174 @@ var propertyWidget = function (mug, options) {
...
+        }
+        if (mug.p.caseTypeXPath && !mug.p.case_type) {
+            $dropdownRow.hide();
+        }
+        mug.on('property-changed', function (e) {
+            if (e.property === 'useCreate') {
+                initSelect2();
+                $toggleLink.toggle(createsCase(mug));
+                if (!createsCase(mug) && mug.p.caseTypeXPath) {
+                    switchToDropdownMode(mug, null, $dropdownRow.next('.widget'));
```
millerdev, line 197: AI comment: No test for `createsCase` toggle while in xpath mode

> The `useCreate` deselect path (line 201) force-switches back to dropdown mode when xpath is active. This behavior isn't tested.

src/saveToCase.js
```diff
@@ -126,6 +127,174 @@ var propertyWidget = function (mug, options) {
...
+        });
+        if (!createsCase(mug) || !mug.p.caseTypeXPath || mug.p.case_type) {
+            $xpathRow.hide();
+        }
+    };
+    return widget;
+}
+
+function addModeToggle($row, text, onClick) {
+    var $link = $('<a href="#" style="font-size:12px;margin-top:4px;display:inline-block;" />')
```
millerdev: AI comment: Inline styles on toggle link. Prefer a CSS class for consistency with the rest of the codebase.

src/saveToCase.js
```diff
@@ -166,7 +335,49 @@ var slugToProp = {
...
+                deserialize: mugs.deserializeXPath,
+                validationFunc: function (mug) {
+                    if (mug.p.case_type || !createsCase(mug)) {
+                        return 'pass';
+                    }
+                    var val = mug.p.caseTypeXPath;
+                    if (!val) {
+                        return gettext("Case Type is required");
+                    }
+                    if (CASE_TYPE_REGEX.test(val)) {
```
millerdev, line 372: AI comment: The `caseTypeXPath` validation checking `CASE_TYPE_REGEX.test(val)` is clever but could false-positive on valid xpath fragments that happen to be alphanumeric (e.g., `count`). The heuristic seems reasonable for now but worth documenting with a comment.

millerdev (commented):
Looking good.

Copy/paste does not work correctly for Case Type when in XPath mode. If I type `/data/text` in the Case Type Xpath input (referencing a "text" question), then copy the Advanced Case Actions question and paste it, the resulting question's Case Type has an invalid xpath value. Paste into Excel shows something like:

Image

It seems like copy/paste thinks `caseTypeXPath` is Display Text rather than XPath. Not sure what's causing that.

---

I saw some inconsistency with clicking the Case Action buttons, but I am unable to reproduce now: When "Create" Case Action is selected, case type selection is lost on click "Close" or "Index". Probably don't worry about it if you can't reproduce.

src/saveToCase.js
jingcheng16: I think the code is harmless, but it is true that those clear is unnecessary and meaningless. Fixed in a4d0347

tests/saveToCase.js
jingcheng16: No test for the xpath-to-dropdown-to-xpath round-trip: added in  5293946
The old test might exists. But since we now support dual mode, there is no need to preserve xpath reference, we just keep it in caseTypeXpath. "should show xpath field when create/case_type bind is an xpath" should cover the scenario that this AI comment trying to cover.

src/saveToCase.js
jingcheng16: Added in 107aa4c

src/saveToCase.js
jingcheng16: I'm confused. Is `count` a valid xpath? Isn't it has to be `count(...)` ?

src/saveToCase.js
jingcheng16: Fixed in 62eb570

jingcheng16:
Hi @millerdev I've address all your feedbacks. Let me know if I missed anything.

> Copy/paste does not work correctly for Case Type when in XPath mode.

734ca27 should fix this! Thanks for finding it!!

But I do noticed other two issues, unrelated to changes in this PR and I'd like to address them in seperate PR:
1. Even though the question has selected "Create", when copy paste, there is an validation error against Case ID
Screenshot 2026-04-08 at 3 04 23 PM

2. When "Create" Case Action is selected, case type selection is lost on click "Close" or "Index". I can reproduce it.

Both can be reproduced on `jc/save-to-case-base`

millerdev (approved):
Fine to address those other issues in a seperate PR.

PR #1216, opened 2026-03-29
Hide Date Modified and User ID in collapsible advanced section

jingcheng16:
## Product Description

Move Date Modified and User ID fields into a collapsible "Advanced" section in the Save to Case question type.
- Fields are hidden by default behind a "View Advanced" / "Hide Advanced" toggle
- Section auto-expands if any field has a warning message
 
Screenshot 2026-03-28 at 10 55 42 PM

Screenshot 2026-03-28 at 10 56 41 PM

## Technical Summary

Ticket: https://dimagi.atlassian.net/browse/SAAS-19463

- Added a reusable `collapse_toggle.html` template using Bootstrap 3's native collapse component
- Added `widgets.util.addCollapseToggle(slug, options)` helper in `widgets.js` for wiring up any section as a B3 collapsible
- CSS uses B3's `.collapsed` class to toggle label/icon visibility — no JS event handlers needed
- Moved `date_modified` and `user_id` into a new "advanced" section in `sectionData`
## Feature Flag

`VELLUM_SAVE_TO_CASE`

### Safety story

Only affects the Save to Case question type UI layout — no data model changes.
### QA Plan

- Manually verify the toggle shows/hides Date Modified and User ID
- Verify auto-expand when a validation message exists on either field

### Rollback instructions

### Labels & Review

### DIFF
--- a/src/less-style/question-props.less
+++ b/src/less-style/question-props.less
@@ -121,6 +121,34 @@
                 }
             }
         }
+
+        .fd-question-fieldset[data-slug='advanced'] {
+            legend {
+                display: none;
+            }
+        }
+
+        .fd-collapse-toggle {
+            padding: 10px 0;
+            &:has(button.collapsed) {
+                margin-bottom: 20px;
+            }
+            button {
+                background: none;
+                border: none;
+                padding: 0;
+                color: @cc-brand-mid;
+                &:hover {
+                    color: darken(@cc-brand-mid, 10%);
+                }
+                &.collapsed .fd-collapse-hide {
+                    display: none;
+                }
+                &:not(.collapsed) .fd-collapse-show {
+                    display: none;
+                }
+            }
+        }
     }
 
     .fd-chips {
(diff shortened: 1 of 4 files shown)


kaapstorm (commented):
I don't know enough about Vellum to approve this, but it looks good to me.

millerdev (commented):
Is the section selector not available for this question type?

Image

While I think the section selector could be improved, I would prefer us to use it consistently across question types rather than inventing a new way for "Advanced" question properties to be displayed as a group. Thoughts?

millerdev (approved):
Don't love having two advanced sections. Approving because my suggestion was overridden.

jingcheng16:
@millerdev Thanks. I will merge it.
Think about this again, I still think this provides the best UI for our case. Otherwise, where should we place this "Advanced" section? Right after "Case Action" field so it is easy to find? But it is weird to have an "Advanced" section above those Case Action section. I hope the Create/Update/Close/Index section can be close to "Case Action" field.

If we place the "Advanced" section at the bottom of the question, and allow the pills to keep insert new sections between the top section and the Advanced section, it's also weird. Advanced section is jumping around...

PR #1217, opened 2026-03-31
Fix case management diff to handle duplicate and modified mappings

millerdev:
## Product Description
Fixes the case management diff algorithm to correctly handle duplicate question mappings and modified mappings. Previously, modifying a mapping (e.g., changing `update_mode`) produced an "update" operation that downstream consumers may not handle. Now, modifications are represented as a delete + add pair, which is more explicit and consistent. Additionally, duplicate mappings (the same question path mapped multiple times) are now diffed correctly, only adding or removing the exact surplus or deficit rather than misidentifying them.

## Technical Summary
- Removed the "updates" concept from the diff schema; modified mappings are now expressed as delete + add pairs. This change requires [dm/form-action-diff](https://github.com/dimagi/commcare-hq/pull/37543), specifically https://github.com/dimagi/commcare-hq/pull/37543/commits/4eda9cff6696f8e4887acd2abb6b2a7a0aced2f4, for proper interoperability with HQ. However, it will have no adverse affects on user who do not have the `FORMBUILDER_SAVE_TO_CASE` feature flag enabled if this is merged before the HQ branch. Update: the HQ branch has been merged.
- Replaced the `question_path`-based matching logic with an exact-match counting algorithm that correctly handles duplicates and tracks how many copies of each item exist in baseline vs. incoming. The new diff algorithm is also [used in HQ](https://github.com/dimagi/commcare-hq/pull/37543/commits/fd9fa9b6d1ecb4a5e97041527b84f44b61ca7236).

:blowfish: Review by commit.

## Feature Flag
`FORMBUILDER_SAVE_TO_CASE`

### Safety story
The diff algorithm is purely functional with no side effects. The change is well-covered by new tests exercising duplicate, triplicate, and cross-property duplicate scenarios. All changed functionality is behind the feature flag.

### Automated test coverage
New test cases added covering: duplicate mappings, triplicate mappings, duplicates with differing `update_mode`, and duplicates across different case properties. Existing tests updated to reflect the new delete + add representation of modifications.

### Rollback instructions

### DIFF
--- a/src/caseDiff.js
+++ b/src/caseDiff.js
@@ -6,7 +6,6 @@ export function compareCaseMappings (baseline, incoming) {
     // the original with the incoming mappings to produce a diff.
     const additions = {};
     const deletions = {};
-    const updates = {};
     const allKeys = new Set([...Object.keys(baseline), ...Object.keys(incoming)]);
 
     allKeys.forEach(key => {
@@ -16,7 +15,8 @@ export function compareCaseMappings (baseline, incoming) {
                 if (!match) {
                     push(key, item, additions);
                 } else if (!_.isEqual(match, item)) {
-                    push(key, item, updates);
+                    push(key, match, deletions);
+                    push(key, item, additions);
                 }
             });
             baseline[key].forEach(item => {
@@ -41,9 +41,6 @@ export function compareCaseMappings (baseline, incoming) {
     }
     if (Object.keys(deletions).length) {
         diff.delete = deletions;
-    }
-    if (Object.keys(updates).length) {
-        diff.update = updates;
     }
     return diff;
 }
--- a/src/caseDiff.js
+++ b/src/caseDiff.js
@@ -12,7 +12,7 @@ export function compareCaseMappings (baseline, incoming) {
         if (Object.hasOwn(baseline, key) && Object.hasOwn(incoming, key)) {
             const cache = {};
             incoming[key].forEach(item => {
-                const num = countExactMataches(item, baseline[key], incoming[key], cache);
+                const num = countExactMatches(item, baseline[key], incoming[key], cache);
                 if (num === null || num > 0) {
                     push(key, item, additions);
                 } else if (num < 0) {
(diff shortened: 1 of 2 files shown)


millerdev:
Re: https://github.com/dimagi/Vellum/pull/1217#discussion_r3015994876

I recalculated algorithmic complexity and came up with slightly different, and I believe more accurate, calculation plus an additional pathological scenario that seems pretty unlikely.

```
O((base + inc) * ((base[k] + inc[k]) ** 2))  # not actually squared because of caching
O((M + N) * ((1 + 2 + 2*log(2)) * 2) * 2) - common case
O(14.4 * (M + N)) - common case
O((M + N) * (200 ** 2) - pathological (100 duplicate questions per case property)?
```

Not adding to the comment thread above to hopefully not confuse Copilot or cause it to get too eager and go making undesired changes.

src/caseDiff.js
```diff
@@ -12,15 +12,15 @@ export function compareCaseMappings (baseline, incoming) {
         if (Object.hasOwn(baseline, key) && Object.hasOwn(incoming, key)) {
             const cache = {};
             incoming[key].forEach(item => {
-                const num = countExactMataches(item, baseline[key], incoming[key], cache);
+                const num = countExactMatches(item, baseline[key], incoming[key], cache);
```
jingcheng16, line 15: or rename to `countExactMatcha` 🍵

jingcheng16:
> multiple mappings existed for the case property -> question path pair. That scenario is not possible to create in Vellum, but could be created externally.

Do you mean identical mappings can be created?
Like "one" -> /data/one ; "one" -> /data/one ?
I tried in Case Management Tab, I'm not able to save such duplicate mappings.

src/caseDiff.js
jingcheng16: jk

millerdev:
> I tried in Case Management Tab, I'm not able to save such duplicate mappings.

There was a bug (now fixed) in the Case Management tab that made it possible to create them. As far as I know it's no longer possible.

jingcheng16 (approved):

PR #1218, opened 2026-04-07
Fix test failure

jingcheng16:
## Technical Summary

This PR is to fix `src/widgets.js: line 946, col 9, Function declarations should not be placed in blocks. Use a function expression or move the statement to the top of the outer function.`

The test failure is introduced by https://github.com/dimagi/Vellum/pull/1216. Because this PR is not based on master so it doesn't have test run automatically. I only realized the error after I merged the PR.

I've now pushed https://github.com/dimagi/Vellum/commit/f5ce7e8a9788446bc43835d3f91c851e3bc45955 to `jc/save-to-case-base`, so this won't happen in the future.

### Rollback instructions

### Labels & Review

### DIFF
--- a/src/widgets.js
+++ b/src/widgets.js
@@ -943,11 +943,11 @@ function addCollapseToggle(slug, options) {
     $section.before(collapse_toggle($.extend({collapseId: collapseId}, options)));
 
     if (options.mug) {
-        function expandIfMessages() {
+        var expandIfMessages = function () {
             if ($content.find('.messages').children().length) {
                 $content.collapse('show');
             }
-        }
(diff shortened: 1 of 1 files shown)


gherceg (approved):

millerdev (commented):
Sorry for [bad recommendation](https://github.com/dimagi/Vellum/pull/1216#discussion_r3011864440), @jingcheng16. Clearly I forgot that declaring a function there is not legal.

jingcheng16:
@millerdev No worries! I should run tests before I merge it!

PR #1219, opened 2026-04-07
Add "locked" widget to lock question properties

nospame:
## Product Description
Adds a "Locked" checkbox, which disables all properties on a question from being edited. Also adds messages when a question is locked and uneditable or a group contains any locked questions.
image

image

image

## Technical Summary
https://dimagi.atlassian.net/browse/SAAS-19544
This is the initial set of changes for Locked Admin questions, essentially an MVP. Specifically, this:
- Removes access to previous vellum:lock functionality. This was never UI-accessible and only partially functional, and deemed OK to remove by product leadership
- Adds the locked widget to all questions & groups, reading from and writing to the question's `<bind>` XML node
- Uses the added locked property to determine which question widgets should be locked. In this iteration it is all of them, though in the future we may add more selective options, which informed the use of "all" on the XML attribute rather than a boolean
- Makes sure locked questions cannot be re-ordered or deleted from the form
- Handles Select questions by applying the locked property to any of their children and removing the "Add Choice" button
- Handles groups by checking whether a group contains any locked child questions, and if so:
  - Locks the question ID and position of the group
  - Ensures the group cannot be deleted

AI (@claude Opus 4.6) was used to plan, debug, and write tests included in this pull request.

## Feature Flag
Not explicitly feature flagged but will be when HQ development begins. For now, the main changes here are not accessible without manually enabling the "lock" plugin.

### Safety story
Developing as a plugin means core code is relatively unaffected, so blast radius would be limited to those with the plugin enabled. Tested locally and (soon) on staging. Added/updated automated tests for all key functionality here.

### Automated test coverage
Added/updated automated testing in formdesigner.lock.js

### QA Plan
This will get QA holistically with the other elements of Locked Admin questions.

### Rollback instructions

### Labels & Review

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1179,7 +1179,7 @@ fn.checkMove = function (srcId, srcType, dstId, dstType, position) {
     }
     if (position === 'inside') { position = 'into'; } // normalize for Vellum
 
-    var locked = !this.isMugPathMoveable(sourceMug.hashtagPath);
+    var locked = !this.isMugPathMoveable(sourceMug, sourceMug.hashtagPath);
     if (locked) {
         if (position === 'into' || position === 'last' || position === 'first') {
             return sourceMug.parentMug === targetMug;
@@ -2486,11 +2486,11 @@ fn.isMugRemoveable = function (mug, path) {
     return mug.options.isRemoveable;
 };
 
-fn.isPropertyLocked = function (mugPath, propertyPath) {
+fn.isPropertyLocked = function (mug, mugPath, propertyPath) {
     return false;
 };
 
-fn.isMugPathMoveable = function (mugPath) {
+fn.isMugPathMoveable = function (mug, mugPath) {
     return true;
 };
 
--- a/src/core.js
+++ b/src/core.js
@@ -1172,23 +1172,15 @@ fn.getRelativePosition = function (mug, position) {
 
 fn.checkMove = function (srcId, srcType, dstId, dstType, position) {
     var form = this.data.core.form,
-        targetMug = form.getMugByUFID(dstId),
         sourceMug = form.getMugByUFID(srcId);
     if (!sourceMug) {
         return false;
     }
-    if (position === 'inside') { position = 'into'; } // normalize for Vellum
-
-    var locked = !this.isMugPathMoveable(sourceMug);
-    if (locked) {
-        if (position === 'into' || position === 'last' || position === 'first') {
-            return sourceMug.parentMug === targetMug;
-        } else {
-            return sourceMug.parentMug === targetMug.parentMug;
-        }
+    if (this.isMugPathMoveable(sourceMug)) {
+        return true;
     }
 
-    return true;
+    return srcId === dstId;
 };
 
 fn.onFormChange = function (mug) {
--- a/src/core.js
+++ b/src/core.js
@@ -1179,7 +1179,7 @@ fn.checkMove = function (srcId, srcType, dstId, dstType, position) {
     }
     if (position === 'inside') { position = 'into'; } // normalize for Vellum
 
-    var locked = !this.isMugPathMoveable(sourceMug, sourceMug.hashtagPath);
+    var locked = !this.isMugPathMoveable(sourceMug);
     if (locked) {
         if (position === 'into' || position === 'last' || position === 'first') {
             return sourceMug.parentMug === targetMug;
@@ -2069,7 +2069,7 @@ fn.getMugToolbar = function (mug, multiselect) {
         $baseToolbar = $(question_toolbar({
             comment: multiselect ? '' : richText.sanitizeInput(mug.p.comment),
             isDeleteable: mugs && mugs.length && _.every(mugs, function (mug) {
-                return _this.isMugRemoveable(mug, mug.hashtagPath);
+                return _this.isMugRemoveable(mug);
             }),
             isCopyable: !multiselect && mug.options.isCopyable,
             sections: multiselect ? [] : _.chain(_this.getSections(mug))
@@ -2128,7 +2128,7 @@ fn.getQuestionTypeChanger = function (mug) {
         }
         return ret;
     };
-    var changeable = this.isMugTypeChangeable(mug, mug.hashtagPath);
+    var changeable = this.isMugTypeChangeable(mug);
 
     var $questionTypeChanger = $(question_type_changer({
         currentQuestionIcon: mug.getIcon(),
@@ -2482,19 +2482,19 @@ fn.getMugSpec = function () {
     return mugs.baseSpecs;
 };
 
-fn.isMugRemoveable = function (mug, path) {
+fn.isMugRemoveable = function (mug) {
     return mug.options.isRemoveable;
 };
 
-fn.isPropertyLocked = function (mug, mugPath, propertyPath) {
+fn.isPropertyLocked = function (mug, propertyPath) {
     return false;
 };
 
-fn.isMugPathMoveable = function (mug, mugPath) {
+fn.isMugPathMoveable = function (mug) {
     return true;
 };
 
-fn.isMugTypeChangeable = function (mug, mugPath) {
+fn.isMugTypeChangeable = function (mug) {
     return mug.options.isTypeChangeable;
 };
 
--- a/src/core.js
+++ b/src/core.js
@@ -1177,7 +1177,7 @@ fn.checkMove = function (srcId, srcType, dstId, dstType, position) {
         return false;
     }
     if (!this.isMugPathMoveable(sourceMug)) {
-        var srcParentId = sourceMug.parentMug ? sourceMug.parentMug.ufid : '#';
+        var srcParentId = sourceMug.parentMug?.ufid || '#';
         return srcParentId === dstId;
     }
     return true;
--- a/src/core.js
+++ b/src/core.js
@@ -1176,11 +1176,11 @@ fn.checkMove = function (srcId, srcType, dstId, dstType, position) {
     if (!sourceMug) {
         return false;
     }
-    if (this.isMugPathMoveable(sourceMug)) {
-        return true;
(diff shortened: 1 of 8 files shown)


src/core.js
```diff
@@ -1179,7 +1179,7 @@ fn.checkMove = function (srcId, srcType, dstId, dstType, position) {
     }
     if (position === 'inside') { position = 'into'; } // normalize for Vellum
 
-    var locked = !this.isMugPathMoveable(sourceMug.hashtagPath);
+    var locked = !this.isMugPathMoveable(sourceMug, sourceMug.hashtagPath);
```
millerdev: ~Would it be hard to change `isMugPathMoveable` to accept a single `mug` argument? It simplifies the signature, and the function can reference `mug.hashtagPath` or whatever other properties it needs with that.~ I see this was done in a later commit.

src/lock.js
```diff
@@ -3,58 +3,32 @@
...
-            propertyPath === 'nodeID') 
-        {
-            return true;
-        } else if (lock === 'value' && propertyPath.indexOf('Itext') === -1) {
-            return true;
-        }
-
-        return false;
+    isPropertyLocked: function (mug, mugPath, propertyPath) {
+        return this.__callOld() || mug.p.locked;
```
millerdev: ~`mug.p.locked` is a string (`'all'`), which should work fine as a boolean value here. Is this and related code below expected to change to handle other values in the future? If not, maybe consider changing it to a boolean?~ Also changed in a later commit.

src/core.js
```diff
@@ -1172,23 +1172,15 @@ fn.getRelativePosition = function (mug, position) {
...
-            return sourceMug.parentMug === targetMug;
-        } else {
-            return sourceMug.parentMug === targetMug.parentMug;
-        }
+    if (this.isMugPathMoveable(sourceMug)) {
+        return true;
     }
 
-    return true;
+    return srcId === dstId;
```
millerdev: Did behavior change here? The relationship between `srcId` and `dstId` is not obvious, and requiring them to be equal seems unintuitive/possibly wrong. Can you explain why this is correct?

tests/formdesigner.lock.js
```diff
@@ -44,47 +45,112 @@ describe("The Lock plugin", function() {
         return call('isMugTypeChangeable', getMug(path));
     }
 
+    it("includes 'locked' in the main properties after 'requiredAttr'", function () {
```
millerdev, line 48: When and to whom will this new property be visible? I think most people using the form builder will not know or care about locked questions, and should not have it taking up space on their screen. If it is visible to more than the small set of people who know what it is and are using it frequently, then I think we should consider moving it to another section (possibly Advanced).

tests/formdesigner.lock.js
```diff
@@ -44,47 +45,112 @@ describe("The Lock plugin", function() {
...
-    it("disallows renaming a locked node", function () {
-        assert(locked('#form/node_locked', 'nodeID'));
-        assert(locked('#form/value_locked', 'nodeID'));
-        assert.isFalse(locked('#form/none_locked', 'nodeID'));
-        assert.isFalse(locked('#form/normal', 'nodeID'));
+    it("sets vellum:lock='all' on rawBindAttributes when locked", function () {
+        const mug = getMug('/data/unlocked');
+        mug.p.locked = true;
+        assert.equal(mug.p.rawBindAttributes['vellum:lock'], 'all');
+        mug.p.locked = false;
```
millerdev: Should this be in a `finally` block or a `beforeEach` function? Seems like it could cause other tests to fail if the above assertion fails.

src/lock.js
```diff
@@ -3,58 +3,125 @@
...
+            serialize: () => {},
+            deserialize: () => {},
+            setter: function (mug, attr, value) {
+                if (value === true) {
+                    mug.p.rawBindAttributes[LOCKED_BIND_ATTR] = 'all';
+                } else {
+                    delete mug.p.rawBindAttributes[LOCKED_BIND_ATTR];
+                }
+                mug.p.set(attr, value);
+                mug.form.getChildren(mug).forEach(child => _this.handleMugParseFinish(child));
```
millerdev: `handleMugParseFinish` is a parse lifecycle function, so calling it here seems strange since this is not part of the parse lifecycle. Wonder if it would make more sense to make a separate method that can be called here and also by `handleMugParseFinish` so this doesn't have to reuse a function that could do parse-related things that don't make sense in here?

tests/formdesigner.lock.js
```diff
@@ -29,60 +32,208 @@ function beforeFn(done) {
...
+
+            mug.p.locked = true;
+            assert(choice.p.locked);
+            mug.p.locked = false;
+            assert(!choice.p.locked);
+        });
+
+        it("prevents moving choices into a locked select", function () {
+            const src = getMug('/data/unlocked_select/choice1'),
+                dst = getMug('/data/locked_select');
```
millerdev: nit
```suggestion
            const src = getMug('/data/unlocked_select/choice1');
            const dst = getMug('/data/locked_select');
```

tests/formdesigner.lock.js
nospame: The lock plugin will be enabled for Advanced and Enterprise plans, and the "Locked" widget will be always-visible only to users with the locked questions user role permission (to be added in HQ), so most people using the form builder will be unlikely to see it. Discussed on Slack and it sounds like this positioning is what we want - highly visible for those who can use it or are affected by it.

src/core.js
nospame: Thanks for calling this out, it is an incorrect change in behavior and would prevent almost all moves. I was confusing `srcID` as referring to the original parent of the node being checked, when it's actually the node itself.

In digging here I also found that `position` as passed into `checkMove` is always an index, not a text string, which is why the previous checks for various strings weren't doing any good in my testing.

Updated and fixed logic here: [71aec29](https://github.com/dimagi/Vellum/pull/1219/commits/71aec29370d977aeac0cfd0335bacb464443760c)

tests/formdesigner.lock.js
nospame: Good call: https://github.com/dimagi/Vellum/pull/1219/commits/b691ba7724708200aba3dd81968993376e81302d

tests/formdesigner.lock.js
nospame: [b6475b3](https://github.com/dimagi/Vellum/pull/1219/commits/b6475b301b84ced6519f11ccd863ac83f8993987)

tests/formdesigner.lock.js
millerdev: Sounds good. Wonder if it would make sense to additionally limit it to apps in a linked apps configuration? It has no use in non-linked apps, right?

Although maybe that would be problematic because initial form building could be done before linked apps are configured?

I'm not asking for any other changes at this time, given what you've said.

src/core.js
```diff
@@ -1172,22 +1172,14 @@ fn.getRelativePosition = function (mug, position) {
...
-
-    var locked = !this.isMugPathMoveable(sourceMug.hashtagPath);
-    if (locked) {
-        if (position === 'into' || position === 'last' || position === 'first') {
-            return sourceMug.parentMug === targetMug;
-        } else {
-            return sourceMug.parentMug === targetMug.parentMug;
-        }
+    if (!this.isMugPathMoveable(sourceMug)) {
+        var srcParentId = sourceMug.parentMug ? sourceMug.parentMug.ufid : '#';
```
millerdev: nit: I became more familiar with the [_optional chaining_](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) operator recently, and I love its conciseness.

```suggestion
        var srcParentId = sourceMug.parentMug?.ufid || '#';
```

millerdev (approved):

src/lock.js
nospame: That makes sense. I expect there will be other UI-related things to do when setting 'locked' in a future PR as well, and this sets a better precedent. b89e74d

src/core.js
nospame: I remember reading about this (or reviewing some code using it) before HQ was on webpack and couldn't support it yet. I like it. cf8e47d

tests/formdesigner.lock.js
nospame: The expectation is this could potentially be useful for a single project space not using linked apps, where they simply need to protect certain items from being edited by lower-privileged app editors.

But yes, this will have the most utility in a future version of linked apps, where an upstream app editor needs to protect against editing by various downstream app editors.

src/core.js
```diff
@@ -1172,22 +1172,14 @@ fn.getRelativePosition = function (mug, position) {
...
-    var locked = !this.isMugPathMoveable(sourceMug.hashtagPath);
-    if (locked) {
-        if (position === 'into' || position === 'last' || position === 'first') {
-            return sourceMug.parentMug === targetMug;
-        } else {
-            return sourceMug.parentMug === targetMug.parentMug;
-        }
+    if (!this.isMugPathMoveable(sourceMug)) {
+        var srcParentId = sourceMug.parentMug?.ufid || '#';
+        return srcParentId === dstId;
```
millerdev, line 1181: This still looks like a slight behavior change from what this function did before this PR, but I assume you've tested it and it behaves acceptably.

millerdev (approved):

src/core.js
nospame: It is somewhat, because the previous behavior didn't (as far as I'm concerned) work. In practice, `isMugPathMoveable` only gets touched by the lock plugin ([in core](https://github.com/dimagi/Vellum/blob/72e82a5a7c3560137ffe6fda59491aea9c6fc88f/src/core.js#L2493-L2495) it's always `true` and no other plugins use it at present) so this won't have any practical effect until the lock plugin is enabled.

PR #1224, opened 2026-04-16
Nudge to assign a case name on registration forms

millerdev:
## Product Description

Registration forms require a question to be mapped to the case `name` property, but it's easy to forget. This change adds a few gentle nudges to help form builders remember:

- **Inline nudge** in the Case Management sidebar reminds users to name their case whenever a registration form has no question mapped to `name`.
- **Save-button warning** on hover surfaces the same issue when attempting to save, along with an **Auto-assign case name** button that maps the first question to `name` (creating a hidden `case-name` calculation question if the first question is already mapped to another property).
- **Auto-assigned case name** is highlighted with a green success message to show what happened when the button is clicked.

The existing general "Case Management" learning nudge continues to show on non-registration forms and on registration forms that already have a name mapped.

Validation error

---

Case name nudge

---

Auto-assigned name hint

---

https://dimagi.atlassian.net/browse/SAAS-19568

## Technical Summary

- New `preSaveValidation` plugin hook on `core` returns an array of warning messages contributed by plugins; `caseManagement` uses it to report the missing case name.
- `validateForSave` in `core.js` was refactored to aggregate multiple alerts into a single popover (previously only the first error was shown).
- The save-button popover trigger changed from `hover` to `manual` so users can move the mouse into the popover to click action buttons without it dismissing; a small hide timer keeps the UX feeling like hover.
- `is_registration_form` is read from the `caseManagement` plugin options (supplied by HQ) to gate the new nudge.

:blowfish: Review by commit

## Feature Flag

`FORMBUILDER_SAVE_TO_CASE`

### Safety story

- Change is mostly additive on the UI layer.
- Auto-assign creates a new `DataBindOnly` question with `calculate="uuid()"` only when the first child question already has a `caseProperty`; otherwise it reuses the existing question. In both cases the user can undo via the usual edit flow before saving.
- Registration forms that already map a question to `name` see no behavior change.

### Automated test coverage

Yes.

### QA Plan

No QA.

### Rollback instructions


src/core.js
```diff
@@ -162,12 +156,36 @@ fn.init = function () {
...
     });
+    const saveButtonUi = _this.data.core.saveButton.ui;
+    let hideTimeout;
+    function showPopover() {
+        cancelHide();
+        saveButtonUi.popover('show');
+    }
+    function scheduleHide() {
+        hideTimeout = setTimeout(function () {
+            saveButtonUi.popover('hide');
```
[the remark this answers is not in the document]
millerdev: 46d9f600397babf0cb26ea3c743974b207c61f58

src/caseManagement.js
```diff
@@ -385,6 +389,56 @@ class CaseMapMaintainer {
...
+    if (!first) {
+        return;
+    }
+    let targetMug = first;
+    if (first.p.caseProperty) {
+        targetMug = form.createQuestion(first, 'before', 'DataBindOnly', true);
+        targetMug.p.nodeID = form.generate_question_id('case-name');
+        targetMug.p.calculateAttr = 'uuid()';
+    }
+    targetMug.p.caseProperty = 'name';
```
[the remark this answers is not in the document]
millerdev: b8cfb2bee88066b515668a603f319203636a69d2

src/templates/case_management_auto_assign_name.html
```diff
@@ -0,0 +1,5 @@
+<div>
+  <button class="btn btn-default btn-xs fd-auto-assign-case-name">
```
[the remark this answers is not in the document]
millerdev: 76312d723c34e952ffa6d69a2fcca562bee5a419

millerdev:
Going to make a few more small tweaks. Hold off on reviewing until further notice.

millerdev:
Ready for review.

jingcheng16:
Have some questions before reviewing the code:

> Dismissing it with « suppresses the reminder going forward.

Curious why dismiss icon is « instead of x ?

> Save-button warning on hover surfaces the same issue when attempting to save

Save is blocked. Is this intentional? I have mixed feelings about this change. As I've also flagged in slack, I encounter an error when assign a question to 'name' property, then the save is blocked and I'm not able to test anything. Blocking save is a kind of dangerous move. I remember we talked about this before, and at that time you said even if a form has error we still want user be able to save it.

> If the inline nudge has been dismissed, the save-button warning includes a » button to re-enable inline reminders.

This is almost invisible... I'd played around the feature on staging several times before I start reviewing the PR and I never noticed it....

> Registration forms require a question to be mapped to the case name property, but it's easy to forget.

If it is forgotten, usually we default the first question in the form to the case name property...I see your other PR to stop this behavior...I wonder if the whole design is a bit aggressive.

millerdev:
> Curious why dismiss icon is « instead of x ?

Because it is not a permanent dismissal. The nudge can be re-enabled by clicking the opposite » button in the save pop-out. I want it to be possible to disable this nudge for advanced users who find it noisy and annoying, but also want people who accidentally click the « button to be able to restore the nudge if they prefer that. It's intentionally subtle so as not to be too annoying.

> If it is forgotten, usually we default the first question in the form to the case name property...I see your other PR to stop this behavior...I wonder if the whole design is a bit aggressive.

Have you reviewed the updated tech spec to see why we decided to do it this way? You can find a link to it in the description of [this ticket](https://dimagi.atlassian.net/browse/SAAS-19568), which I forgot to link in the PR description (sorry! I will add it).

jingcheng16 (approved):

src/templates/case_management_auto_assigned_name.html
```diff
@@ -0,0 +1,8 @@
+<div class="alert alert-success fd-nudge">
+    <button type="button" class="close" data-dismiss="alert"
+            aria-label="<%-gettext("Close")%>">
+      &times;
+    </button>
+    <i class="fa fa-check-circle"></i>
+    <%-gettext("Automatically assigned.")%>
+</div>
```
jingcheng16: nit: miss a new line at the end of the file

jingcheng16 (approved):

src/templates/case_management_auto_assigned_name.html
millerdev: https://github.com/dimagi/Vellum/pull/1224/commits/d1d4cce3574167092500ca04f0e78857fd91b346

jingcheng16 (approved):

src/caseManagement.js
```diff
@@ -201,19 +198,44 @@ class CaseMappingsBuilder {
...
     }
+}
 
-    buildQuestionMappingsFromCaseMappings(caseMappings) {
-        const mappingsByQuestion = {};
-        Object.entries(caseMappings).forEach(([caseProperty, questions]) => {
+/**
+ * Add case mappings to plugin data
+ *
+ * Mappings for unknown questions are not added to the plugin data.
```
jingcheng16, line 206: I understand what you're doing in this commit. But do you know why unknown questions are existed in the first place? Especially in my case, I create a new registration form, and create a new form by editing xml, I didn't delete any questions, why there will be an unknown question?

jingcheng16 (approved):

src/caseManagement.js
millerdev: Yes, the default mapping set is `{"name": [{"question_path": null, "edit_mode": "always"}]}` because `<OpenCaseAction>.name_update` is required and always exists, even when there is no assigned question.

I also may make an HQ PR to exclude that mapping when sending to Vellum, but thought it was good to have Vellum handle the scenario gracefully as well.

PR #1225, opened 2026-04-16
Improve Create Section: add open case condition and dedicated fields

jingcheng16:
## Product Description

The old Create section required users to manually enter case_name, case_type, and owner_id as generic key-value properties, each with a Property Name, Calculation, and an optional Relevant (update condition) field. We’ve already reconciled case_type with a top-level dropdown. This PR is to:
- make case_name and owner_id dedicated single fields that only require dragging and dropping a question reference - removing the need to type property names manually.
- when the action is create, user no longer needs to select update action as well to put other case properties, user can add other case properties in "Case Properties To Create"

Before
Screenshot 2026-04-16 at 10 22 08 AM

After
Screenshot 2026-04-16 at 10 18 52 AM

### Backward compatibility

Legacy forms that used per-property `relevant` on `case_type`, `case_name`, or `owner_id` create binds are automatically migrated:
- `case_type` and `case_name` relevants are promoted to **Open Case Condition** (combined with `and` if they differ).
- `owner_id` relevant is absorbed if it matches the promoted condition, otherwise kept as **Owner ID Condition**.
- Legacy forms with both `<create>` and `<update>` sections merge update properties into the Create section.

See the [design decision document](https://docs.google.com/document/d/1a2S1b1spUumzGV82FrrloZ01GER5qkAnwtjg12JhAto/edit?tab=t.2g3s56t3ix7z#bookmark=id.dfsjehh4860) for the full rationale on how `relevant` fields are handled.

## Technical Summary

Ticket: https://dimagi.atlassian.net/browse/SAAS-19457

### Why `aec64e3` (XPath validation) is needed

The xPath widget's built-in logic reference tracking (via the logic manager) validates expressions during interactive editing — but it does NOT validate on form load. When a legacy form with an invalid XPath in a promoted `relevant` is loaded, the logic manager silently fails to track references without producing an error. The `validationFunc` in `aec64e3` catches this by running `form.xpath.parse()` during `mug.validate()`, which fires when the mug is displayed. The message text matches the logic manager's message (`"Invalid XPath expression."` with period) so `getMessages` deduplicates them during interactive editing.

## Feature Flag

`VELLUM_SAVE_TO_CASE`

### Safety story

Tested locally multiple times. I also test on [production forms that has per-property relevant](https://docs.google.com/spreadsheets/d/1R5zTAz36bLehx-p35k3OFRqK9TI2By_mjSSPKYIAGHw/edit?gid=0#gid=0) to make sure them will load as expected so user's workflow won't be disrupted. 

### Automated test coverage

New test suites:
- **"dedicated create fields"**
- **"create section backward compatibility"**

### QA Plan

- Load existing production forms with per-property relevant → verify Open Case Condition is populated, no per-property relevant in output
- Create new form with Open Case Condition → verify case-level bind in XML
- Verify bubbles display correctly for promoted conditions
- Verify invalid XPath shows exactly one warning message

### Rollback instructions

### Labels & Review

### DIFF
--- a/src/saveToCase.js
+++ b/src/saveToCase.js
@@ -486,23 +486,17 @@ var slugToProp = {
                 validationFunc: function (mug) {
                     if (mug.p.useCreate) {
                         var props = _.without(_.keys(mug.p.createProperty), ""),
-                            required = ["case_name"],
-                            optional = ["owner_id"],
-                            legal = _.union(required, optional),
-                            illegalProps = _.difference(props, legal),
-                            requiredProps = _.intersection(props, required),
+                            reserved = ["case_type", "case_name", "owner_id"],
+                            reservedUsed = _.intersection(props, reserved),
                             invalidProps = _.filter(props, function(p) {
                                 return !VALID_PROP_REGEX.test(p);
                             });
 
-                        if (requiredProps.length !== required.length) {
+                        if (reservedUsed.length > 0) {
                             return util.format(
-                                gettext("You must include {columns} columns to create a case"),
-                                {columns: required.join(", ")}
+                                gettext("{props} cannot be added here. Use the dedicated fields above."),
+                                {props: reservedUsed.join(", ")}
                             );
-                        } else if (illegalProps.length > 0) {
-                            return gettext("You can only use the following properties:") +
-                                " " + legal.join(', ');
                         } else if (invalidProps.length > 0) {
                             return util.format(
                                 gettext("{props} are invalid properties"),
@@ -632,8 +626,16 @@ var slugToProp = {
                 actions.push(simpleNode('create', makeColumns(createProps)));
             }
 
-            if (updatesCase(mug)) {
-                actions.push(simpleNode('update', makeColumns(mug.p.updateProperty)));
+            // <update> from extra create properties or standalone update action
+            var updateProps = {};
+            if (createsCase(mug) && mug.p.createProperty) {
+                _.extend(updateProps, _.omit(mug.p.createProperty, ""));
+            }
+            if (updatesCase(mug) && mug.p.updateProperty) {
+                _.extend(updateProps, _.omit(mug.p.updateProperty, ""));
+            }
+            if (!_.isEmpty(updateProps)) {
+                actions.push(simpleNode('update', makeColumns(updateProps)));
             }
 
             if (closesCase(mug)) {
@@ -711,7 +713,7 @@ var slugToProp = {
                     }
                     ret.push(ownerBind);
                 }
-                ret = ret.concat(generateBinds('create', mug.p.createProperty));
+                ret = ret.concat(generateBinds('update', mug.p.createProperty));
             }
             if (updatesCase(mug)) {
                 ret = ret.concat(generateBinds('update', mug.p.updateProperty));
--- a/src/saveToCase.js
+++ b/src/saveToCase.js
@@ -901,6 +901,45 @@ var slugToProp = {
         ]
     };
 
+function promoteStashedCreateBindRelevants(mug) {
+    var caseTypeRelevant = mug.p._caseTypeRelevant,
+        caseNameRelevant = mug.p._caseNameRelevant,
+        ownerIdRelevant = mug.p._ownerIdRelevant;
+    delete mug.p._caseTypeRelevant;
+    delete mug.p._caseNameRelevant;
+    delete mug.p._ownerIdRelevant;
+
+    var caseTypeAndNameRelevants = _.compact(
+        _.uniq([caseTypeRelevant, caseNameRelevant])
+    );
+    if (caseTypeAndNameRelevants.length && !mug.p.openCaseCondition) {
+        mug.p.openCaseCondition = caseTypeAndNameRelevants.join(" and ");
+    }
+
+    if (ownerIdRelevant) {
+        var redundantWithCaseTypeOrName = _.contains(
+            caseTypeAndNameRelevants,
+            ownerIdRelevant
+        );
+        var redundantWithOpenCase = ownerIdRelevant === mug.p.openCaseCondition;
+        if (!redundantWithCaseTypeOrName && !redundantWithOpenCase) {
+            mug.p.ownerIdCondition = ownerIdRelevant;
+        }
+    }
+}
+
+function mergeUpdatePropertiesIntoCreateAfterParse(mug) {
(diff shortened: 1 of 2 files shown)


src/saveToCase.js
```diff
@@ -810,6 +901,45 @@ var slugToProp = {
         ]
     };
 
+function promoteStashedCreateBindRelevants(mug) {
+    var caseTypeRelevant = mug.p._caseTypeRelevant,
+        caseNameRelevant = mug.p._caseNameRelevant,
+        ownerIdRelevant = mug.p._ownerIdRelevant;
+    delete mug.p._caseTypeRelevant;
+    delete mug.p._caseNameRelevant;
+    delete mug.p._ownerIdRelevant;
```
millerdev: I don't see this `delete mug.p.property` pattern elsewhere in Vellum. Have you confirmed that it works? Might be better to use `mug.p.set('property')`, which deletes `'property'` from the mug's properties.

millerdev (commented):
I ran out of time and quickly reviewed the final tests commit, so should spend a bit more time on that, but otherwise looks good.

src/saveToCase.js
jingcheng16: ahhh, sorry, this is confusing. `_caseTypeRelevant`, `_caseNameRelevant`, `_ownerIdRelevant` technically are not spec properties on mug, they're just plain JS property on `MugProperties` object. `mug.p.set('attr')` won't work because it deletes from `__data`. But I can see it is confusing because the pattern is `mug.p.attr`... I've now moved them to `mug._stashedCreateBindRelevants` to avoid confusion. 60a076b

tests/saveToCase.js
```diff
@@ -584,4 +583,198 @@ describe("The SaveToCase module", function() {
...
+
+        it("should reject reserved properties in createProperty", function () {
+            util.loadXML("");
+            var mug = util.addQuestion("SaveToCase", "mug");
+            mug.p.useCreate = true;
+            mug.p.caseName = "/data/name";
+            mug.p.createProperty = {
+                case_name: { calculate: "/data/name" },
+            };
+            assert.notEqual(mug.spec.createProperty.validationFunc(mug), "pass");
```
millerdev, line 605: How hard would it be to check for the specific expected error here rather than "not pass"?

Example:

https://github.com/dimagi/Vellum/blob/5c0c6eda1fea810fe8f3a68fce3cb7ab3ed0e251/tests/form.js#L99

There is at least one other assertion in this file that might also benefit from that technique.

tests/saveToCase.js
```diff
@@ -584,4 +583,198 @@ describe("The SaveToCase module", function() {
...
+        it("should merge update into createProperty on parse when both create property and update property exist", function () {
+            util.loadXML(LOGIC_TEST_XML);
+            var mug = util.getMug("save_to_case_create");
+            assert.equal(mug.p.caseName, "'name'");
+            assert.equal(mug.p.useCreate, true);
+            assert.equal(mug.p.useUpdate, false);
+            assert.deepEqual(_.without(_.keys(mug.p.createProperty), ""), ["p1", "p2"]);
+        });
+
+        it("should preserve user-provided ownerIdCondition even when openCaseCondition has same condition", function () {
```
millerdev, line 635: Given this behavior, why not always move `owner_id` relevant to `ownerIdCondition` rather than merging the two as outlined in the spec?

> If `case_name` + `owner_id` or `case_type` + `owner_id` share the same relevant → move to Open Case Condition. XML changes.

Seems like that could simplify the implementation and maybe not be quite as surprising for users who (for example) accidentally set the both `case_name` + `owner_id` relevants to the same expression when they intended to only set `owner_id` relevant?

tests/saveToCase.js
```diff
@@ -584,4 +583,198 @@ describe("The SaveToCase module", function() {
...
+                $xml.find('bind[nodeset="/data/save_to_case/case/create/case_name"]').attr('relevant')
+            );
+            assert.notOk(
+                $xml.find('bind[nodeset="/data/save_to_case/case/create/owner_id"]').attr('relevant')
+            );
+        });
+    });
+
+    describe("openCaseCondition promotion", function () {
+        it("should preserve user-provided ownerIdCondition even when openCaseCondition has same condition", function () {
```
millerdev, line 757: How is this test different from the one above on line 635?

```
it("should preserve user-provided ownerIdCondition even when openCaseCondition has same condition", ...
```

millerdev (approved):
Responses to all comments can be implemented in a follow-up PR if you agree with them but would prefer merge this PR ASAP.

tests/saveToCase.js
jingcheng16: Yes it will simplify the implementation. But I believe the intention for user to leave same relevant for both `case_name` and `owner_id` or `case_type` and `owner_id` is to open the case conditionally and I want the new UI will accurately reflect that.
> maybe not be quite as surprising for users who (for example) accidentally set the both case_name + owner_id relevants to the same expression when they intended to only set owner_id relevant?

This is possible. I wonder how well our user remember the value they typed in all mugs...For users who intentionally set both relevant the same expression and intend to make them just open case condition, if we always move `owner_id` relevant to `ownerIdCondition` will we surprise user by open case condition and owner id condition be the same?

tests/saveToCase.js
jingcheng16: With my changes in https://github.com/dimagi/Vellum/pull/1227, now each field carry its own validation, the property level will never return the reserved-name error, it will only return an `errorSummary` which is an generic string: “One or more properties above have errors. Fix the highlighted fields.” . And I find checking against the generic string is no different than checking pass.

tests/saveToCase.js
jingcheng16: nice catch. I'll address in a follow up PR.

jingcheng16: https://github.com/dimagi/Vellum/pull/1232

tests/saveToCase.js
jingcheng16: I guess you're right. The scenario I don't like is, if user explicitly type the same condition in our new UI for Open Case Condition and Owner ID Condition, we still wipes out Owner ID Condition. I will address is in a follow up PR to always move `owner_id` relevant to `ownerIdCondition`

jingcheng16: follow up in https://github.com/dimagi/Vellum/pull/1237

PR #1227, opened 2026-04-23
repeater card UI for Create/Update/Index properties

jingcheng16:
## Product Description

Vellum has a recurring widget pattern for editing a list of structured items. Every time it shows up, it's rendered as a gray background card containing three or four rows, each a full-width <input> whose only clue about what it represents is the placeholder text. No labels. Little visual grouping that says "these three fields belong together". No indication and no validation of which fields are required vs. optional. 

**Before**
Screenshot 2026-04-23 at 12 12 02 AM

This PR is to replace the old gray card with a new card:
- Field labe is now on the left instead of as placeholder
- Calculation and Condition are now both fully-supported Xpath field
- Errors are now attached to individual fields inside the card with a summary message at the end of the list

**Create Section**

Screenshot 2026-04-23 at 12 00 38 AM

**Update Section**

Screenshot 2026-04-23 at 12 11 17 AM

**Index Section**

Screenshot 2026-04-23 at 12 03 43 AM

Inline Validation + Widget-level message at the end of the list
Screenshot 2026-04-23 at 12 12 43 AM

## Technical Summary

Ticket: https://dimagi.atlassian.net/browse/SAAS-19666
The ticket doesn't has any useful content. It is just created to track my work.

Design Spec: https://docs.google.com/document/d/1a2S1b1spUumzGV82FrrloZ01GER5qkAnwtjg12JhAto/edit?tab=t.2g3s56t3ix7z#bookmark=id.kiknyxtcwm8j

Renaming mentioned in the tech spec will be done in a follow up PR

Review by commit

I tried my best to improve the readability of the code and admit there are still parts that even me find not very intuitive. I had to use a lot of real form to test the feature, come up with edge case to test the feature to make sure it won't break anything.

I had to repeat some validation code for both inline validation and data-side validation. Because inline validation is  DOM-driven, it won't work when the save to case mug is not the current selected mug. I cannot use data-side validation only, which only validate at property level while I want the inline effect.

## Feature Flag

`VELLUM_SAVE_TO_CASE`

### Safety story

- **Data model is unchanged.** `mug.p.createProperty` / `updateProperty` / `indexProperty` JSON shape is identical; all existing XML fixture round-trips (`loadXML(UPDATE_PROPERTY_XML)` → `assertXmlEqual(call("createXML"), UPDATE_PROPERTY_XML)`) pass untouched.
- Tested using real form on prod env.

### Automated test coverage

new tests across `tests/saveToCase.js` and `tests/widgets.js` 

### QA Plan

Manual spot-checks worth doing in a dev env:
1. Load an existing form with SaveToCase mugs (Create + Update + Index populated). Verify UI renders cards identically to before — same fields, same values — and a round-trip save produces identical XML.
2. Click "Add property" in each section. Type a name, type an xpath. Verify inline validation fires for bad names/expressions and clears once fixed. Save; reload; verify the row persists.
3. Drag a question from the tree onto an xpath field inside a card. Verify it inserts as a bubble (rich-text) or string (plain) as appropriate.
4. Remove a card that has validation errors. Verify the widget-level alert + tree-triangle + save-tooltip clear when the errored card is gone.
5. Try the example form pattern from a customer: one SaveToCase creates a case, another indexes it via `/data/.../case/@case_id`. Verify no spurious "Unknown question" warning appears.

### Rollback instructions

Data shapes are unchanged; a revert brings back the old UI without any migration. Saved XMLs produced on the new UI are indistinguishable from ones produced on the old.

### Labels & Review


src/widgets.js
```diff
@@ -965,6 +934,351 @@ function setWidget($el, widget) {
...
+ *     `displayXPathEditor` (the modal launcher). May include
+ *     `leftPlaceholder`, `rightPlaceholder`, `autocompleteChoices`, and
+ *     `lstring` (all forwarded to the modal).
+ * @param {Object} context - Per-call glue supplied by the caller:
+ *   @param {Function} context.getValue - Returns the current value to seed
+ *       the modal with.
+ *   @param {Function} context.setValue - Called with the user's committed
+ *       value when the modal is dismissed with a save.
+ *   @param {Function} context.onDone - Called after `setValue` so the
+ *       caller can react (e.g. fire a change event, re-validate).
```
millerdev: Why is `onDone` needed if it is called immediately after `setValue`? Seems valuable to remove `onDone` to simplify the interface; callers can provide an implementation of `setValue` that calls `onDone` if they need that feature.

src/templates/nested_dropdown_field.html
```diff
@@ -0,0 +1,40 @@
+<%
+  // Nested dropdown field template used by repeater-card rows,
+  // with fallthrough rendering for a current value that isn't in the list
+  // (e.g. parsed from saved XML that references a case type no longer
+  // defined, or a relationship value outside `child`/`extension`). Both
+  // cases we still want to show to the user rather than silently discard.
```
millerdev: nit
```suggestion
  // defined, or a relationship value outside `child`/`extension`). In both
  // cases we still want to display it rather than silently discarding it.
```

src/templates/widget_repeater_card.html
```diff
@@ -0,0 +1,103 @@
...
+  //                   objects, or a resolver `(mug, opts) => [...]`.
+  //   extraValidator — optional per-field validator, returns an error
+  //                   string or null (see `validateField`).
+  //
+  var fields = cardConfig.fields,
+      rootClass = cardConfig.rootClass,
+      cardHeaderText = cardConfig.cardHeaderText;
+%>
+<div class="fd-repeater-list">
+  <% _.each(props, function (v, k) { %>
```
millerdev: Single-letter variable names are not ideal when their scope spans many lines.

src/templates/widget_repeater_card.html
```diff
@@ -0,0 +1,103 @@
...
+  //                   (`validateField`, `getValue`) to find the field in
+  //                   the DOM, and often grep'd by identifier.
+  //   widget        — "input" (default) | "xpath" | "dropdown".
+  //   required      — adds data-required="true", enables the inline
+  //                   "Required" validator in widgets.js `validateField`.
+  //   placeholder   — plain-text + xpath modes only.
+  //   options       — dropdown mode only: array of strings, {value, label}
+  //                   objects, or a resolver `(mug, opts) => [...]`.
+  //   extraValidator — optional per-field validator, returns an error
+  //                   string or null (see `validateField`).
```
millerdev: This is a lot to take in. Nearly as many lines of docstring as the template! I wonder if it could be said more succinctly? ~Possibly not since it describes a rather complicated nested data structure.~

I like this description better: it's shorter and I find it more readable:

https://github.com/dimagi/Vellum/blob/e6f6b9550ae562a68c808e982df2280aa9d50048/src/widgets.js#L968-L991

src/widgets.js
```diff
@@ -960,6 +964,167 @@ function enableAutocompleteOnInput($input, mug, options) {
     });
 }
 
+// -------------------------------------------------------------------------
+// Repeater card widget — a compound-list property (as opposed to a scalar
```
millerdev: Wondering if "Repeater" is the best word to use here, given that we also have a _Repeat Group_ question type, which is an entirely different thing. Alternate suggestions: "Cards widget" or "Card list widget"

millerdev: What do you think of breaking widgets.js into multiple files similar to how mugs.js was split up? It's getting pretty big. Maybe a simple place to start would be to move this new widget to its own module/file?

src/widgets.js
```diff
@@ -960,6 +964,167 @@ function enableAutocompleteOnInput($input, mug, options) {
...
+
+    // Read values straight from the DOM using `cardConfig.fields` — no
+    // per-mug `getValue` override needed. Each card's key is the
+    // `isIdentifier` field's value; each card's body is
+    // {valueKey: fieldValue, ...}.
+    widget.getValue = function () {
+        var currentValues = {};
+        widget.input.find('.' + cardConfig.rootClass).each(function () {
+            var $card = $(this), key = null, entry = {};
+            _.each(cardConfig.fields, function (f) {
```
millerdev: Single letter variable names are not great. Please use a more descriptive name.

src/widgets.js
```diff
@@ -960,6 +964,167 @@ function enableAutocompleteOnInput($input, mug, options) {
...
+                var val = readFieldValue($card.find('.' + f.fieldClass));
+                if (f.isIdentifier) { key = val; }
+                else if (f.valueKey) { entry[f.valueKey] = val; }
+            });
+            if (key !== null) { currentValues[key] = entry; }
```
millerdev: Is it possible for `key` to be `undefined` here? Does this do the right thing if it is? Maybe it could be simplified to
```suggestion
            if (key) { currentValues[key] = entry; }
```
Which would ignore all keys like `null`, `undefined`, `0`, and `""`.

src/widgets.js
```diff
@@ -960,6 +964,167 @@ function enableAutocompleteOnInput($input, mug, options) {
...
+        e.preventDefault();
+        $(this).closest('.fd-repeater-card').remove();
+        widget.handleChange();
+    };
+
+    widget.addProperty = function (e) {
+        e.preventDefault();
+        var currentValues = widget.getValue();
+        // If there's already a blank card, focus it instead of adding another.
+        if (!("" in currentValues)) {
```
millerdev: I think this would be safer as `!Object.hasOwn(currentValues, "")`

The `in` operator matches any item in the object prototype (very sad).

src/saveToCase.js
```diff
@@ -44,100 +40,132 @@ function usesCases(mug) {
...
+// Shared Calculation + Condition fields for Create and Update configs.
+var SAVE_PROPERTY_CALC_FIELD = {
+        label: gettext("Calculation"), fieldClass: "fd-update-property-source",
+        valueKey: "calculate", widget: "xpath", required: true,
+    },
+    SAVE_PROPERTY_RELEVANT_FIELD = {
+        label: gettext("Condition"), fieldClass: "fd-update-property-relevant",
+        valueKey: "relevant", widget: "xpath",
+        placeholder: gettext("Optional — leave blank to always apply"),
+    };
```
millerdev, line 83: These would be easier to read if wrapped to one item per line.

src/saveToCase.js
```diff
@@ -44,100 +40,132 @@ function usesCases(mug) {
...
+var CREATE_CARD_CONFIG = {
+        rootClass: "fd-update-property",
+        cardHeaderText: gettext("Case property"),
+        addLabel: gettext("Add property"),
+        errorSummary: gettext("One or more properties above have errors. Fix the highlighted fields."),
+        requiresAtLeastOne: false,
+        emptyStateMessage: null,
+        fields: [
+            {label: gettext("Property Name"), fieldClass: "fd-update-property-name",
+             isIdentifier: true, required: true},
```
millerdev: Wrap with one item per line. The same applies in several places below.

src/saveToCase.js
```diff
@@ -44,100 +40,132 @@ function usesCases(mug) {
...
+        fields: [
+            {label: gettext("Relationship Identifier"), fieldClass: "fd-index-property-name",
+             isIdentifier: true, required: true},
+            {label: gettext("Referenced Case ID"), fieldClass: "fd-index-property-source",
+             valueKey: "calculate", widget: "xpath", required: true},
+            {label: gettext("Referenced Case Type"), fieldClass: "fd-index-property-case-type",
+             valueKey: "case_type", widget: "dropdown",
+             placeholder: gettext("Select from existing case types"),
+             options: function (mug, opts) {
+                 return (opts.vellum.data.saveToCase || {}).existingCaseTypes || [];
```
millerdev: nit
```suggestion
                 return opts.vellum.data.saveToCase?.existingCaseTypes || [];
```

src/saveToCase.js
```diff
@@ -44,100 +40,132 @@ function usesCases(mug) {
...
 
-        return widget;
-    };
+// Apply `fn` to every xpath string, and overwrite in place if the returned
+// value differs. Used by `updateLogicExpressions` to rewrite paths when a
+// referenced question is renamed.
+function rewriteCardXPaths(rowMap, keys, fn) {
+    forEachCardXPath(rowMap, keys, function (row, key) {
+        var next = fn(row[key]);
+        if (next !== row[key]) { row[key] = next; }
```
millerdev: Is `row[key]` always a simple value that can be compared with `!==`? This would not work if it were an object or array (`{}` or `[]`), for example. Consider using `!_.isEqual(next, row[key])` if deep equality is necessary.

src/saveToCase.js
```diff
@@ -44,100 +40,132 @@ function usesCases(mug) {
...
+    });
+}
+
+// Walk the nested card data and detect required-missing or invalid-xpath
+// errors; when anything's broken, the parent widget's validationFunc
+// returns `cardConfig.errorSummary` as the widget-level message.
+function hasNestedCardError(mug, propMap, fields) {
+    var broken = false;
+    _.each(propMap || {}, function (v, name) {
```
millerdev: Single-letter variable name `v`. Please use a more descriptive name.

The same below for `f`.

tests/saveToCase.js
```diff
@@ -777,4 +777,204 @@ describe("The SaveToCase module", function() {
...
+
+            // Click the Add button inside the Update section.
+            var $updateSection = $("[name='property-updateProperty']");
+            $updateSection.find(".fd-add-property").trigger("click");
+
+            $list = $(".fd-update-property").filter(".fd-repeater-card");
+            assert.equal($list.length, 1, "Add creates one blank card");
+            assert.equal(
+                $list.find(".fd-update-property-name").val(), "",
+                "new card's name is empty"
```
millerdev: nit
```suggestion
                "new card's name should be empty"
```

Other tests added in this commit have similar assertion messages that seem confusing if the assertion fails.

tests/saveToCase.js
```diff
@@ -777,4 +777,204 @@ describe("The SaveToCase module", function() {
...
+
+        it("should add a blank card when the Add button is clicked", function () {
+            util.loadXML("");
+            util.addQuestion("SaveToCase", "mug", {
+                case_id: "/data/meta/caseID",
+                useUpdate: true,
+            });
+            util.clickQuestion("mug");
+            var $list = $(".fd-update-property").filter(".fd-repeater-card");
+            assert.equal($list.length, 0, "no cards initially");
```
millerdev:
```suggestion
            assert.equal($list.length, 0, "unpexected initial card(s)");
```

tests/saveToCase.js
```diff
@@ -777,4 +777,204 @@ describe("The SaveToCase module", function() {
...
+
+            // Click the Add button inside the Update section.
+            var $updateSection = $("[name='property-updateProperty']");
+            $updateSection.find(".fd-add-property").trigger("click");
+
+            $list = $(".fd-update-property").filter(".fd-repeater-card");
+            assert.equal($list.length, 1, "Add creates one blank card");
```
millerdev:
```suggestion
            assert.equal($list.length, 1, "Add should create one blank card");
```

src/saveToCase.js
```diff
@@ -40,6 +40,34 @@ function usesCases(mug) {
 }
 
 
+// Inline per-field validators. Each returns an error string to show under
+// the field, or null when the value passes. `validateField` in widgets.js
+// picks them up via the field config's `extraValidator` key.
+function validatePropertyNameChars(val) {
+    if (val && !/^[a-z0-9_-]+$/i.test(val)) {
+        return gettext("Use only letters, numbers, '-' and '_'");
```
millerdev: In addition to these allowed characters, case properties must start with a letter. The regex I am planning to use in the caseManagement  plugin is `/^[a-z][\w-]*/i`

src/saveToCase.js
```diff
@@ -40,6 +40,34 @@ function usesCases(mug) {
...
+    }
+    return null;
+}
+
+function validateCreatePropertyName(val) {
+    if (val && _.contains(["case_type", "case_name", "owner_id"], val)) {
+        return gettext("Use the dedicated field above instead.");
+    }
+    if (val && val.startsWith("name")) {
+        return gettext("Use the case_name field above instead.");
```
millerdev, line 56: This will match a value like "name123", which is not the case name. Should it fail validation?

src/widgets.js
millerdev, line 1: Finding it hard to believe that the three tests added in 01a0911c61e4e57ff7bbf69017d573f42de4cd2c cover all of the changes introduced in this file in the same commit. Please take a look and confirm that there is good test coverage for all of the changes.

tests/saveToCase.js
```diff
@@ -227,7 +227,7 @@ describe("The SaveToCase module", function() {
                 relationship: "notchildorextension",
             }
         };
-        assert.strictEqual(mug.spec.indexProperty.validationFunc(mug), "Relationship must be child or extension");
+        assert.notEqual(mug.spec.indexProperty.validationFunc(mug), "pass");
```
millerdev, line 230: I think the old way of testing this error was better. Checking for `!== "pass"` means it could be any error, even the wrong one. Maybe use `assert.match(error, /must be child or extension/)` to only match the important part of the message?

millerdev (commented):
Reviewed up to 01a0911c61e4e57ff7bbf69017d573f42de4cd2c; ran out of time when I got to the tests in that commit (did not review them yet).

src/saveToCase.js
millerdev: Edit: didn't get that regexp quite right. Here's what I ended up with: `/^[a-z][\w-]*$/i`

src/logic.js
```diff
@@ -268,6 +268,27 @@ LogicManager.prototype = {
...
+                        var suffix = pathString.substring(ancestorPath.length);
+                        if (candidate.options.isValidSubPath &&
+                            candidate.options.isValidSubPath(suffix)) {
+                            refMug = candidate;
+                        }
+                        break;
+                    }
+                    ancestorPath = ancestorPath.replace(/\/[^/]*$/, '');
+                }
+            }
```
millerdev: This adds a big chunk of code to an even larger function. I know you said it gets reverted later, so maybe there is a better place for this comment, but consider moving this code to its own function?

src/saveToCase.js
```diff
@@ -174,6 +174,39 @@ function rewriteCardXPaths(rowMap, keys, fn) {
...
+// (and any non-interactive code path that calls `mug.validate()`) sees
+// the spec's `validationFunc`, not the DOM. This helper reads mug.p
+// directly so the same per-field rules (Required, XPath syntax,
+// extraValidator) apply regardless of touched state.
+function hasRepeaterCardFieldError(mug, cardMap, cardConfig) {
+    var fields = cardConfig.fields;
+    var hasError = false;
+    _.each(cardMap || {}, function (card, cardId) {
+        if (hasError) { return; }
+        var cardIsEmpty = !cardId && _.every(card, function (v) { return !v; });
```
millerdev: nit
```suggestion
        var cardIsEmpty = !cardId && _.every(card, (v => !v));
```

src/saveToCase.js
```diff
@@ -174,6 +174,39 @@ function rewriteCardXPaths(rowMap, keys, fn) {
...
+// validator is touched-state gated and DOM-driven, so it only catches
+// errors for cards the user has actually interacted with. `isFormValid`
+// (and any non-interactive code path that calls `mug.validate()`) sees
+// the spec's `validationFunc`, not the DOM. This helper reads mug.p
+// directly so the same per-field rules (Required, XPath syntax,
+// extraValidator) apply regardless of touched state.
+function hasRepeaterCardFieldError(mug, cardMap, cardConfig) {
+    var fields = cardConfig.fields;
+    var hasError = false;
+    _.each(cardMap || {}, function (card, cardId) {
```
millerdev: Rather than using `_.each` with a `hasError` flag, consider using `_.find` and returning true when an error is found. `_.each` is less efficient because it visits every item while `_.find` will stop iterating when an error is found. The same applies to the inner use of `_.each` below.

src/saveToCase.js
```diff
@@ -174,6 +174,39 @@ function rewriteCardXPaths(rowMap, keys, fn) {
     });
 }
 
+// Data-side mirror of widgets.js `validateField`. The widget's inline
+// validator is touched-state gated and DOM-driven, so it only catches
+// errors for cards the user has actually interacted with. `isFormValid`
+// (and any non-interactive code path that calls `mug.validate()`) sees
+// the spec's `validationFunc`, not the DOM. This helper reads mug.p
+// directly so the same per-field rules (Required, XPath syntax,
+// extraValidator) apply regardless of touched state.
```
millerdev: Is this large comment useful? It looks AI-written, and I didn't find it very informative. Maybe it could be condensed?

millerdev (commented):
There is a lot of clearly AI-generated code in this PR that doesn't follow coding best practices (large comments, single letter variable names, etc.). I'd like to see that cleaned up as much as possible. I'm finding the verbosity of the code makes it difficult to understand what's going on in this PR.

src/templates/widget_repeater_card.html
jingcheng16: Sorry about that. I think I fixed most of single-letter variable... but looks like I still missed a lot... 834c632

src/templates/widget_repeater_card.html
jingcheng16: I didn't even realized the comment in widget.js and the template overlapped... My head has been dizzy last week. I checked some templates in this repo and none of them has comments, so I removed all comments in the `widget_repeater_card.html` and consolidate it into `widget.js`. I removed the usage example because 1) if I keep it then this comment will be too long 2) user can find existing examples in the code
c253bf0

src/widgets.js
jingcheng16: I think it sounds great! ad0cf13

src/widgets.js
jingcheng16: 38d67e1

src/saveToCase.js
jingcheng16: 4c47419

src/saveToCase.js
jingcheng16: 4c47419

src/saveToCase.js
jingcheng16: This function is later deleted. But I've addressed all single letter variables except for `e` (stands for event).

src/saveToCase.js
jingcheng16: Yes it will always be an xpath string, so it will never be an object or array. I also renamed some variables here to help with readability: 3ad7139

src/widgets.js
jingcheng16: I don't want `""` to be ignored, it is ok that user fill in `calculate` or `relevant` first, and those two values should be preserved, waiting for user to fill in the identifier field later.
`key` ( I've renamed it to `cardIdentifier`) only ends up as `null` if there's no `isIdentifier: true` field in cardConfig.fieldSpecs at all. ( which is a misconfiguration and shouldn't happen )

src/widgets.js
jingcheng16: I realized I should move nestedXPathField too: 658be66f8f1a150a220aa28169ca5ac765519de8

src/widgets.js
jingcheng16: d759c2f

src/widgets.js
jingcheng16: Renamed in two steps: 1f6e3bf84ccc2575e6529906fb11d44be5237a9d and 1e7d53660f0c9c343d67d46e3a284f3045524161

src/widgets.js
jingcheng16: Nice catch. I merged `setValue` and `onDone` to done because it is called `done` before extraction. 03d9349f4fa1f995bb60b7a9132e0d9fe3a0f89b

tests/saveToCase.js
jingcheng16: Thanks. I bulk improved them in f18a9818c73784c3e0bb27beb433403cce13d0da

tests/saveToCase.js
jingcheng16: Thanks. I bulk improved them in f18a9818c73784c3e0bb27beb433403cce13d0da

tests/saveToCase.js
jingcheng16: Thanks. I bulk improved them in f18a9818c73784c3e0bb27beb433403cce13d0da

src/saveToCase.js
jingcheng16: Thanks, I'll make a follow up PR to make this regex a constant once this PR is merged to `jc/save-to-case-base`: 922aa15dc4fb8895e4a07fc5eff05cf64103f688

src/saveToCase.js
jingcheng16: 🫤 an awkward overlook...7a4d2e77bfcc9ed294bdb620d354d73f3bdc2390

src/widgets.js
jingcheng16: Added more tests in 3a699e93e31ef4411505cc6a9f96ab390b285cef. Unknown path check is removed later so I didn't add test for that.

tests/saveToCase.js
jingcheng16: It's no longer possible to catch the specific error because those are rendered inline. On property level, it will always just be "One or more index properties above have errors. Fix the highlighted fields.", the errorSummary defined in card config.

src/logic.js
jingcheng16: I'm a bit confused. The comment and the code are reverted together...Are you suggesting that in the future, whenever I need to add comments in a big chunk of code, I should move the code to its own function, so the comment can also move along and won't make the already large function larger?

src/saveToCase.js
jingcheng16: I intentionally left it because I thought people might wondering given we already have inline validator, and we propogate inline validation to mug.messages, why we still need this data-side validation. But if you don't find it useful I can remove it.  feba4b1d01530d2284832cd282fe5c806433d9fb This commit also removed a lot of other comments. >_< AI wrote a lot of comments, the comments that I kept are those I find useful when I try to understand the code. I've also rephrased some of them to hopefully make them helpful for reviewer and maintainer. But now when I looked at them again I feel they are not that necessary.

src/saveToCase.js
jingcheng16: 💡 I replaced with `_.some` which will return false if not found. a37a6e2403d48f5d16cfcc53f8f376d134d7454a

src/saveToCase.js
jingcheng16: 71a4302c49f1ceb697ec720c5df01794027ddfee

tests/saveToCase.js
millerdev: That's disappointing. Seems like a different error could cause tests like this to pass when they should not.

src/logic.js
millerdev: Yes. In general try not to allow functions to become too big. It usually makes them hard to understand at a glance.

In this case it's fine to do nothing since this code was later reverted.

millerdev (approved):

PR #1229, opened 2026-04-23
UI Improvements for Locked Admin Questions

nospame:
## Product Description
Adds form tree indicators for question locked status, updates question property enabled state immediately based on locking/unlocking, and adds a "select all"-like lock and unlock for all questions contained in a group when it gets locked or unlocked. 

Not shown: removes "Edit Source XML" and "Edit Bulk Translations" options when a form contains any locked questions and the `edit_locked_questions` feature is false or absent.

https://github.com/user-attachments/assets/60d5d47c-4179-4f86-af6a-37133b84c172

## Technical Summary
https://dimagi.atlassian.net/browse/SAAS-19545
UI updates as noted in product description, as well as some small refactors for code readability and efficiency.

## Feature Flag
In HQ the `lock` plugin is enabled by LOCKED_ADMIN_QUESTIONS.

### Safety story
Tested functionality locally. The vast majority of changes here are limited to the lock plugin so any current issues would be constrained to only those with the feature flag enabled for testing purposes, with any remaining edge cases hopefully caught by QA.

### Automated test coverage
Automated testing is added or modified for the key functionality added in this PR.

### QA Plan
This feature will get QA as a whole before removing the feature flag and making it generally available.

### Rollback instructions

### Labels & Review

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1966,6 +1966,12 @@ fn.setTreeValidationIcon = function (mug) {
     }
 };
 
+/**
+ * Extension point for plugins to add inline icons next to mug name in the form tree.
+ * Should populate a data.extraIcons object on the mug's corresponding jstree node.
+ * 
+ * @param mug - The mug to add icons for.
+ */
 fn.setTreeExtraIcons = function (mug) {};
 
 fn._resetMessages = function (errors) {
--- a/src/core.js
+++ b/src/core.js
@@ -1028,6 +1028,10 @@ fn._createJSTree = function () {
                 var args = Array.prototype.slice.call(arguments),
                     node = this.parent.redraw_node.apply(this.inst, args);
                 obj = this.inst.get_node(obj);
+                // add any extra icons if present
+                if (node && obj.data?.extraIcons) {
+                    $(node).find('a > i').first().after(Object.values(obj.data.extraIcons));
+                }
                 // decorate node with error indicator if present
                 if (node && obj.data && obj.data.errors) {
                     $(node).find('a > i').first().after(obj.data.errors);
@@ -1529,6 +1533,7 @@ fn._populateTree = function (selectedHashtag) {
             if (!changed && mug.hasErrors()) {
                 _this.setTreeValidationIcon(mug);
             }
+            _this.setTreeExtraIcons(mug);
             _this.setTreeActions(mug);
         }
     });
@@ -1960,6 +1965,8 @@ fn.setTreeValidationIcon = function (mug) {
         this.jstree("redraw_node", node);
     }
 };
+
+fn.setTreeExtraIcons = function (mug) {};
 
 fn._resetMessages = function (errors) {
     var error, messages_div = this.$f.find('.fd-messages');
--- a/src/dataSourceWidgets.js
+++ b/src/dataSourceWidgets.js
@@ -190,7 +190,12 @@ function fixtureWidget(mug, options, labelText) {
     if (options.hasAdvancedEditor) {
         widget.getUIElement = function () {
             var query = widgets.util.getUIElementWithEditButton(
-                    widgets.util.getUIElement(widget.input, labelText),
+                    widgets.util.getUIElement(
+                        widget.input,
+                        labelText,
+                        !!widget.isDisabled(),
+                        widget.getHelp(),
+                    ),
                     function () {
                         mug.form.vellum.displaySecondaryEditor({
                             source: getSource(mug),
@@ -206,7 +211,8 @@ function fixtureWidget(mug, options, labelText) {
                                 }
                             }
                         });
-                    }
+                    },
+                    !!widget.isDisabled(),
                 );
             query.find(".fd-edit-button").text("...");
             return $("<div></div>").append(query);
--- a/src/lock.js
+++ b/src/lock.js
@@ -98,7 +98,11 @@ $.vellum.plugin("lock", {}, {
             help: gettext("A locked question cannot be edited, moved, or deleted from the form."),
             helpURL: "https://www.example.com",  // placeholder for public documentation
             serialize: () => {},
-            deserialize: () => {},
+            deserialize: (data, key, mug, context) => {
+                if (mug.p.rawBindAttributes && mug.p.rawBindAttributes[LOCKED_BIND_ATTR]) {
+                    delete mug.p.rawBindAttributes[LOCKED_BIND_ATTR];
+                }
+            },
             setter: function (mug, attr, value) {
                 if (value === true) {
                     mug.p.rawBindAttributes = mug.p.rawBindAttributes || {};
--- a/src/lock.js
+++ b/src/lock.js
@@ -60,7 +60,7 @@ $.vellum.plugin("lock", {}, {
     handleMugParseFinish: function (mug) {
         this.__callOld();
         if (_.contains(SELECT_CLASSES, mug.__className)) {
-            setControlOnlyChildrenLocked(mug);
+            propagateLockToControlOnlyChildren(mug);
         }
     },
     setTreeActions: function (mug) {
@@ -122,21 +122,19 @@ $.vellum.plugin("lock", {}, {
                 _this.setTreeExtraIcons(mug);
 
                 if (_.contains(SELECT_CLASSES, mug.__className)) {
-                    setControlOnlyChildrenLocked(mug);
+                    propagateLockToControlOnlyChildren(mug);
                 }
 
                 if (!mug.form.isLoadingXForm) {
                     if (_.contains(GROUP_CLASSES, mug.__className)) {
-                        cascadeLockToChildren(mug, value);
+                        cascadeLockToDescendants(mug, value);
                     } else if (_.contains(STATIC_SELECT_CLASSES, mug.__className)) {
                         _this.setTreeActions(mug);
                     }
                     if (_this.getCurrentlySelectedMug() === mug) {
                         _this.displayMugProperties(mug);
                     }
-                    if (mug.parentMug) {
-                        updateParentTreeIcons(mug.parentMug);
-                    }
+                    updateAncestorTreeIcons(mug);
                 }
             },
         };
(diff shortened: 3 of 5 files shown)


src/core.js
```diff
@@ -1981,6 +1966,8 @@ fn.setTreeValidationIcon = function (mug) {
     }
 };
 
+fn.setTreeExtraIcons = function (mug) {};
```
nospame, line 1975: I think I should either add a comment/docstring here or set up a basic data structure (since we expect to access `data.extraIcons`). Do reviewers have thoughts on one option or the other?

dannyroberts, line 1975: May be nice to add a comment here and above (where `obj.data.extraIcons` is accessed) to say this is a hook for a plugin to implement icons.

tests/formdesigner.lock.js
```diff
@@ -130,18 +130,6 @@ describe("The Lock plugin", function() {
...
-            const groupWithNestedLock = getMug('/data/group_with_nested_lock');
-            assert(call('_hasLockedChildren', groupWithNestedLock),
-                "group with deeply nested locked question should have locked children");
-            const subgroup = getMug('/data/group_with_nested_lock/subgroup');
-            assert(call('_hasLockedChildren', subgroup),
-                "subgroup directly containing locked question should have locked children");
-            const groupNoLock = getMug('/data/group_no_lock');
-            assert.isFalse(call('_hasLockedChildren', groupNoLock),
-                "group without locked children should return false");
-        });
```
dannyroberts, line 143: Would be nice to get a clear explanation for why these tests are safe/necessary to delete.

src/lock.js
```diff
@@ -46,6 +46,22 @@ $.vellum.plugin("lock", {}, {
...
+        const node = mug.ufid && this.jstree('get_node', mug.ufid);
+        if (!node) { return; }
+
+        let icon = null;
+        if (mug.p.locked) {
+            icon = treeLockIcon(mug);
+        }
+
+        if (node.data.extraIcons?.lock !== icon) {
+            node.data.extraIcons = node.data.extraIcons || {};
```
dannyroberts, line 83: Not a strong preference, but consider initializing `node.data.extraIcons` at an earlier stage so that it is always present and set to `{}` or `{lock: null}` or whatever the proper empty value is.

src/lock.js
```diff
@@ -46,6 +46,22 @@ $.vellum.plugin("lock", {}, {
         }
         this.__callOld();
     },
+    setTreeExtraIcons: function (mug) {
```
dannyroberts, line 72: Stray thought, maybe `syncTreeExtraIcons` or something like that would be a more descriptive name. `set` to me sounds like it takes a value and sets something to that value.

src/lock.js
```diff
@@ -138,8 +155,38 @@ function hasLockedChildren(mug) {
     );
 }
 
+function hasUnlockedChildren(mug) {
+    return mug.form.getChildren(mug).some(child =>
```
dannyroberts, line 203: Probably not worth doing anything about this, but we had a quick discussion that if full tree search ever becomes a performance problem here, that maybe breadth-first instead of depth-first might be heuristically faster.

tests/formdesigner.lock.js
```diff
@@ -214,6 +214,53 @@ describe("The Lock plugin", function() {
...
+        it("shows a lock icon on a locked group with no unlocked children", function () {
+            const icon = getLockIcon('/data/locked_group');
+            assert(icon, "expected icon on locked group");
+            assert.include(icon, 'fa-lock');
+            assert.notInclude(icon, 'fa-unlock');
+        });
+
+        it("shows an unlock icon on a locked group with unlocked children", function () {
+            const icon = getLockIcon('/data/locked_group_with_unlocked_children');
+            assert(icon, "expected lock icon on locked select");
```
dannyroberts: Evan pointed this out: should say "locked group"

src/lock.js
```diff
@@ -92,6 +92,12 @@ $.vellum.plugin("lock", {}, {
                 mug.p.set(attr, value);
                 mug.form.getChildren(mug).forEach(child => setLockedFromParent(child));
                 _this.setTreeExtraIcons(mug);
+
+                if (!mug.form.isLoadingXForm) {
+                    if (mug.parentMug) {
+                        updateParentTreeIcons(mug.parentMug);
+                    }
```
dannyroberts: Your choice, but I noticed you could replace these 3 lines with just

```js
updateParentTreeIcons(mug)
```

src/lock.js
```diff
@@ -41,9 +41,10 @@ $.vellum.plugin("lock", {}, {
         setLockedFromParent(mug);
     },
     setTreeActions: function (mug) {
-        if (mug.p.locked) {
-            mug.options.canAddChoices = false;
+        if (!mug.options.hasOwnProperty('_originalCanAddChoices')) {
+            mug.options._originalCanAddChoices = mug.options.canAddChoices;
```
dannyroberts: A comment might be helpful here. Basically `_originalCanAddChoices` is whether a question, forgetting whether it's locked or not, fundamentally has `.canAddChoices`, and we save it here while we kind of hijack `canAddChoices` to mean whether (due to a question being locked or otherwise) it should be displayed with the ability to add chocies.

src/lock.js
```diff
@@ -94,6 +95,9 @@ $.vellum.plugin("lock", {}, {
                 _this.setTreeExtraIcons(mug);
 
                 if (!mug.form.isLoadingXForm) {
+                    if (_.contains(["Select", "MSelect"], mug.__className)) {
```
dannyroberts: What's special about `["Select", "MSelect"]` is that these are questions that normally can take choices. Maybe make this a constant, and use in other places where you need to figure this out?

src/lock.js
```diff
@@ -196,10 +198,12 @@ function hasUnlockedChildren(mug) {
     );
 }
 
-function setLockedFromParent(mug) {
-    if (mug.parentMug && mug.options.isControlOnly) {
-        mug.p.set('locked', mug.parentMug.p.locked);
-    }
+function setControlOnlyChildrenLocked(mug) {
```
dannyroberts: Name is a bit confusing. Can you come up with a better one?

src/lock.js
```diff
@@ -113,13 +113,16 @@ $.vellum.plugin("lock", {}, {
...
+                _this.setTreeExtraIcons(mug);
+
                 if (_.contains(["Select", "MSelect"], mug.__className)) {
                     setControlOnlyChildrenLocked(mug);
                 }
-                _this.setTreeExtraIcons(mug);
 
                 if (!mug.form.isLoadingXForm) {
-                    if (_.contains(["Select", "MSelect"], mug.__className)) {
+                    if (mug.__className === "Group") {
```
dannyroberts: Does "Group" (here and below) capture every relevant container type? (Testing on the call, Evan said this might not cover repeat groups.)

Like my feedback for selects, probably worth making this list of container types a constant.

src/lock.js
nospame: I didn't manage to get farther than the very verbose `propagateLockedToControlOnlyChildren`, the reason being that this is useful to both select/multi-select questions with choices, and those that use lookup table references. That also came with a rename to clarify "children" vs "descendants" that I hope makes it somewhat clearer what's happening https://github.com/dimagi/Vellum/pull/1229/commits/22571501648c710a3f1ed77e4dbfea80248c54d2

src/lock.js
nospame: Updated to simply reference the class name to determine if we should touch `canAddChoices`. https://github.com/dimagi/Vellum/pull/1229/commits/70c6f937875d9c71cbc0b3b9f53e7ac07ac99c6a

src/lock.js
nospame: I agree in general, but `set` is also what we use for the preexisting `setTreeValidationIcon` (which would show an error icon in the same place) and I think existing naming is more consistent with that.

src/core.js
nospame: https://github.com/dimagi/Vellum/pull/1229/commits/f2560238284b0bfa0e612dff5f4229fbfc59a603

src/lock.js
nospame: Captured here https://github.com/dimagi/Vellum/pull/1229/commits/40f487d1158e561c1a608c6ecd3339c63fdec754

src/lock.js
nospame: Thank you for bringing this up. I don't think it's immediately clear that breadth-first would be faster here, but it does seem important to consider since forms can get pretty complex / nested and, like you said, if we start noticing any performance issues for heavy users of Locked Questions features.

tests/formdesigner.lock.js
nospame: You raised a good point which is something I'll put into practice going forward, that the better way to do this would be to update the test to use the new functions by exposing them on the lock.js plugin to ensure it still passes -- _then_ remove it if necessary.

In this case the test was removed, yes because the functionality is no longer exposed by lock.js, but also because it's already implicitly tested through "disallows moving a group that contains a locked question", "disallows deleting a group that contains a locked question", and "disallows changing ID of a group that contains a locked question" - plus additional tests that were added in later commits for the `fa-lock` / `fa-unlock` icons.

src/lock.js
nospame: Since this is just the concern of `setTreeExtraIcons` I don't think I'd want to set it anywhere outside this function. One option would be to initialize it on the `core` stub method - that requires getting the node via jstree just to set an empty `{}` and not do anything with it, which doesn't feel super useful to me. The other option I see might be just moving this a few lines up out of the conditional. Do you think that would be better?

dannyroberts (approved):

dannyroberts (approved):

src/dataSourceWidgets.js
```diff
@@ -206,7 +211,8 @@ function fixtureWidget(mug, options, labelText) {
                                 }
                             }
                         });
-                    }
+                    },
+                    !!widget.isDisabled(),
```
millerdev, line 215: nit: This argument is far away from the function name, and it takes a bit to parse that it's the third argument, not to mention what it does. What do you think of passing it in an `options` object?

`{isDisabled: !!widget.isDisabled()},`

which facilitates something more like keyword arguments. In case not obvious, it would change the signature of `getUIElementWithEditButton` to `($uiElem, editFn, options)`.

PR #1230, opened 2026-04-24
Hide case management for ineligible question types

millerdev:
## Product Description

The Case Management section is now hidden for question types that cannot meaningfully be mapped to case properties. Specifically, the section no longer appears for:

- Multimedia questions (Audio, Image, Video, etc.)
- Save to Case questions (the section didn't actually appear, but auto-assign could assign to it)
- Trigger (label) questions
- Group-type containers (Group, Repeat, Field List)
- Questions inside a Repeat group

The "auto-assign case name" action also now correctly skips these ineligible question types when picking the first question to use as the case name, and will insert a new hidden case-name question instead.

## Technical Summary

- Moved the logic that marks question types as `caseProperty: { presence: 'notallowed' }` out of `getMugTypes` and into a new `postInit` hook. This allows `caseProperty.presence` to be set on `SaveToCase`, which was not previously possible because the `saveToCase` plugin comes later in the call stack than `caseManagement` (depending on configuration).
- Added an explicit check in `getSectionDisplay` so that any question whose ancestor chain contains a `Repeat` returns an empty section. Previously an eligible question inside a Repeat still showed the Case Management UI.
- Updated `autoAssignName` so that when the first question is a type where `caseProperty` is not allowed (no spec or presence `!== 'optional'`), it inserts a new hidden `case-name` data node instead of attempting to reuse the ineligible question. Uses `getPresence` so mug-level overrides are respected.

https://dimagi.atlassian.net/browse/SAAS-19672
https://dimagi.atlassian.net/browse/SAAS-19673

## Feature Flag

`FORMBUILDER_SAVE_TO_CASE`

### Safety story

This change only affects UI visibility of the Case Management section and which question types can be auto-selected as the case name. The underlying form XML and existing case mappings are unchanged. Existing forms that already have case properties set on now-excluded types will continue to function; the UI simply stops offering new mappings for them.

### Automated test coverage

Yes.

### Rollback instructions

### DIFF
--- a/src/caseManagement.js
+++ b/src/caseManagement.js
@@ -458,6 +458,19 @@ $.vellum.plugin('caseManagement', {}, {
         initAutoAssignName(this);
     },
 
+    postInit: function() {
+        this.__callOld();
+        const types = this.data.core.mugTypes.normalTypes;
+        const exclude = {'Trigger': true, 'SaveToCase': true};
+        _(types).each((type, name) => {
+            if (Object.hasOwn(exclude, name) ||
+                    type.dataType === 'binary' ||  // multimedia: Audio, Image, ...
+                    type.tagName === 'group') {
+                type.spec.caseProperty = { presence: 'notallowed' };
+            }
+        });
+    },
+
     loadXML: function () {
         this.__callOld();
         const _this = this;
@@ -504,19 +517,6 @@ $.vellum.plugin('caseManagement', {}, {
             const writer = new XMLCaseMappingWriter(xmlWriter);
             writer.writeCaseMappingsElement(this.data.caseManagement);
         }
-    },
-
-    getMugTypes: function () {
-        const types = this.__callOld();
-        const excludedTypes = [
-            types.normal.Trigger,
-            types.normal.Group,
-            types.normal.Repeat,
-            types.normal.FieldList
-        ];
-        excludedTypes.forEach(excludedType => excludedType.spec.caseProperty = { presence: 'notallowed' });
-
-        return types;
     },
 
     getSections: function (mug) {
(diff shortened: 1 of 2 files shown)


nospame (approved):
Looks good to me, with a couple of questions.

> Existing forms that already have case properties set on now-excluded types will continue to function; the UI simply stops offering new mappings for them.

Is there no valid use case for case properties being set on these now-excluded types, or are we choosing not to offer it for other reasons (e.g., it would be too complex or confusing for most users)? 

And would those old case properties be able to be removed if someone finds they are causing issues?

src/caseManagement.js
```diff
@@ -458,6 +458,19 @@ $.vellum.plugin('caseManagement', {}, {
         initAutoAssignName(this);
     },
 
+    postInit: function() {
+        this.__callOld();
+        const types = this.data.core.mugTypes.normalTypes;
+        const exclude = {'Trigger': true, 'SaveToCase': true};
+        _(types).each((type, name) => {
+            if (Object.hasOwn(exclude, name) ||
```
jingcheng16, line 466: Why use a dictionary for exclude when all key is true?
What do you think of this for simplicity:
```suggestion
        const excludedTypeNames = ['Trigger', 'SaveToCase'];
        _(types).each((type, name) => {
            if (_.contains(excludedTypeNames, name) ||
```

jingcheng16 (approved):
Just a nit comment

src/caseManagement.js
millerdev: With only two types to exclude it probably doesn't matter, but using an array with `_.contains` is O(N) while `Object.hasOwn` is O(1). The dictionary value doesn't matter in this case. I'm going to keep the original implementation.

millerdev:
@nospame, I should have removed that AI-generated sentence from the safety story.

> Is there no valid use case for case properties being set on these now-excluded types...?

There is no valid use case.

> And would those old case properties be able to be removed if someone finds they are causing issues?

They could be removed in the Case Management tab, but not in the Form Builder.

PR #1231, opened 2026-04-27
Fix premature 'Case ID is required' warning on SaveToCase

jingcheng16:
## Product Description

Adding a SaveToCase question or clicking the Create chip on one for the first time fires the "Case ID is required." warning before the user had any chance to fill in the Case ID field. 
This PR is to fix it.

## Technical Summary

Review by commit. Technical reason for each change was added in commit messages.

## Feature Flag

`VELLUM_SAVE_TO_CASE`

### Safety story
It only change when validation messages are surfaced, the underlying validation are unchanged.

### Rollback instructions

### Labels & Review

### DIFF
--- a/src/saveToCase.js
+++ b/src/saveToCase.js
@@ -19,19 +19,19 @@ function validateXPath(mug, value) {
 }
 
 function createsCase(mug) {
-    return mug ? mug.p.useCreate : false;
+    return !!(mug && mug.p.useCreate);
 }
 
 function closesCase(mug) {
-    return mug ? mug.p.useClose : false;
+    return !!(mug && mug.p.useClose);
 }
 
 function updatesCase(mug) {
-    return mug ? mug.p.useUpdate : false;
+    return !!(mug && mug.p.useUpdate);
 }
 
 function indexesCase(mug) {
-    return mug ? mug.p.useIndex : false;
+    return !!(mug && mug.p.useIndex);
 }
 
 function usesCases(mug) {
--- a/src/saveToCase.js
+++ b/src/saveToCase.js
@@ -558,7 +558,7 @@ var slugToProp = {
                     if (mug.p[prop]) {                                                                                                                                                                                       
                         mug.p[prop] = false;                                     
                     }
-                    mug.form.vellum.collapseSection(slug, true);  
+                    mug.form.vellum.collapseSection(slug, true);
                 },
             },
             useCreate: {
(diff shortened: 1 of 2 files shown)


src/saveToCase.js
```diff
@@ -554,8 +554,11 @@ var slugToProp = {
                     mug.form.vellum.collapseSection(slug, false);
                 },
                 onDeselect: function (slug, mug) {
-                    mug.p[slugToProp[slug]] = false;
-                    mug.form.vellum.collapseSection(slug, true);
+                    var prop = slugToProp[slug];
+                    if (mug.p[prop]) {                                                                                                                                                                                       
+                        mug.p[prop] = false;                                     
+                    }
+                    mug.form.vellum.collapseSection(slug, true);
```
millerdev: What changed on this line? Looks like maybe some spaces were added at the end?

src/saveToCase.js
```diff
@@ -19,19 +19,19 @@ function validateXPath(mug, value) {
 }
 
 function createsCase(mug) {
-    return mug ? mug.p.useCreate : false;
+    return !!(mug && mug.p.useCreate);
```
millerdev: > when a chip is never selected, it's value will be `undefined`

Why does that matter? In a boolean context it has the same meaning as `false`.

```suggestion
    return mug?.p.useCreate;
```

or if a boolean is really important
```suggestion
    return !!mug?.p.useCreate;
```

The same applies to similar lines below.

millerdev (approved):
No blocking comments.

src/saveToCase.js
jingcheng16: ahhh yes some spaces...

src/saveToCase.js
jingcheng16: The boolean value matters because [it will be passed to tags](https://github.com/dimagi/Vellum/blob/ec3fce979b234d65fcba2f2fa252d83890a1882a/src/saveToCase.js#L253).
`tags: !!createsCase(mug)` also fix the problem.

jingcheng16: 20b337f5d9ab3e1c17260cd10d7822521fc5b5d2

src/saveToCase.js
jingcheng16: ed09c0450715e300dac8c9a053f704e49704291c

millerdev (approved):

PR #1234, opened 2026-05-01
Add learn more help link

jingcheng16:
## Product Description

As discussed in ga save to case channel, add a help link next to the "Advanced Case Actions" button
Screenshot 2026-05-01 at 10 51 38 AM

## Feature Flag

`VELLUM_SAVE_TO_CASE`

### Safety story

Safe, just add a link

### Rollback instructions

### Labels & Review

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -2117,7 +2117,7 @@ fn.getMugToolbar = function (mug, multiselect) {
         _this.refreshCurrentMug();
     });
     if (!multiselect) {
-        $baseToolbar.find('.btn-toolbar.pull-left')
+        $baseToolbar.find('.fd-toolbar-start')
             .prepend(this.getQuestionTypeChanger(mug));
         if (mug.p.comment) {
             $baseToolbar.find('.fd-question-comment').show();
--- a/src/core.js
+++ b/src/core.js
@@ -13,6 +13,7 @@ import form_errors_template from "vellum/templates/form_errors_template.html";
 import pre_save_alerts from "vellum/templates/pre_save_alerts.html";
 import question_fieldset from "vellum/templates/question_fieldset.html";
 import question_type_changer from "vellum/templates/question_type_changer.html";
+import question_help_link from "vellum/templates/question_help_link.html";
 import question_toolbar from "vellum/templates/question_toolbar.html";
 import alert_global from "vellum/templates/alert_global.html";
 import modal_content from "vellum/templates/modal_content.html";
@@ -2117,8 +2118,11 @@ fn.getMugToolbar = function (mug, multiselect) {
         _this.refreshCurrentMug();
     });
     if (!multiselect) {
-        $baseToolbar.find('.fd-toolbar-start')
-            .prepend(this.getQuestionTypeChanger(mug));
+        var $leftToolbar = $baseToolbar.find('.fd-toolbar-start');
+        $leftToolbar.prepend(this.getQuestionTypeChanger(mug));
+        if (mug.options.helpURL) {
+            $leftToolbar.append(question_help_link({helpURL: mug.options.helpURL}));
+        }
         if (mug.p.comment) {
             $baseToolbar.find('.fd-question-comment').show();
         }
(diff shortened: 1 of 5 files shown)


src/saveToCase.js
```diff
@@ -384,6 +384,7 @@ var slugToProp = {
     CASE_XMLNS = "http://commcarehq.org/case/transaction/v2",
     saveToCaseMugOptions = {
         typeName: 'Advanced Case Actions',
+        helpURL: 'https://dimagi.atlassian.net/wiki/spaces/cc/pages/2146608216/CommCare+Supported+Feature+Save+to+Case+Feature+Flag',
```
gherceg, line 387: Just confirming it is expected for this page to not be publicly accessible yet?

src/core.js
```diff
@@ -2117,8 +2118,11 @@ fn.getMugToolbar = function (mug, multiselect) {
         _this.refreshCurrentMug();
     });
     if (!multiselect) {
-        $baseToolbar.find('.fd-toolbar-start')
-            .prepend(this.getQuestionTypeChanger(mug));
+        var $leftToolbar = $baseToolbar.find('.fd-toolbar-start');
```
gherceg, line 2121: nit: use `let`? We use ES6 now right?

src/templates/question_toolbar.html
```diff
@@ -1,6 +1,6 @@
-<div class="fd-question-toolbar-container clearfix">
-    <% if (isCopyable) { %><div class="btn-toolbar pull-left"></div><% } %>
-    <div class="btn-toolbar pull-right">
```
gherceg, line 3: Looks like these get replaced in bootstrap 4 anyway so one less thing to do in a migration :)

gherceg (approved):
Nothing blocking.

src/less-style/question-props.less
```diff
@@ -10,6 +10,20 @@
         border-bottom: 1px solid @navbarBackground;
         padding: @contentPaddingVellum;
 
+        .fd-question-toolbar-container {
+            display: flex;
+            align-items: center;
+
+            .btn-toolbar {
+                display: flex;
+                align-items: center;
```
nospame, line 19: This is a very targeted style but it seems `.btn-toolbar` has some meaning in B3 already: https://getbootstrap.com/docs/3.3/components/#btn-groups-toolbar. Though it's not clear to me we were actually using it that way here.

src/templates/question_toolbar.html
```diff
@@ -1,6 +1,6 @@
...
+<div class="fd-question-toolbar-container">
```
nospame, line 1: Do we not need `clearfix` anymore? I guess maybe removing the floating `pull-left` and `pull-right` fixes it?

src/saveToCase.js
jingcheng16: Yes I'll replace it with a real page later. I haven't draft the documentation yet.

src/templates/question_toolbar.html
jingcheng16: Yes!!! That's part of my intention

src/core.js
jingcheng16: I think because most of Vellum code is using `var` so I keep using `var` everywhere. We can do a whole file refactor at some point.

nospame (approved):
Not sure the move to flexbox for this element seems necessary to me, but looks good.

src/templates/question_toolbar.html
jingcheng16: Yes!

src/less-style/question-props.less
jingcheng16: Sharp eye. [`btn-toolbar` will give its children `btn` or `btn-group` spacing via `margin-left: 5px` (with `-5px` on the toolbar itself)](https://github.com/twbs/bootstrap/blob/68b0d231a13201eb14acd3dc84e51543d16e5f7e/dist/css/bootstrap.css#L3748-L3760)... The override won't conflicts with its bootstrap style. And the scope is limited to `.fd-question-toolbar-container` so I think it is fine.

PR #1237, opened 2026-05-05
Always promote owner_id relevant to Owner ID Condition

jingcheng16:
## Product Description

Follow-up to #1225. When migrating legacy forms with per-property `relevant` on `owner_id`, always promote it to **Owner ID Condition** instead of absorbing it when it matches the promoted **Open Case Condition**.

## Feature Flag

`VELLUM_SAVE_TO_CASE`

### Safety story

Small, isolated change in the legacy-form migration path. The new behavior is strictly more conservative — it preserves an expression rather than dropping it.

### Automated test coverage

Updated tests in `tests/saveToCase.js` under `create section backward compatibility`:
- `should promote all-same relevant to openCaseCondition and ownerIdCondition`
- `should promote shared case_name+owner_id relevant to both conditions`
- `should output case-level relevant instead of per-property relevant after loading legacy form`

### QA Plan

- Load a legacy form where `case_type`/`case_name`/`owner_id` all share the same `relevant` → both Open Case Condition and Owner ID Condition should be populated with that expression.
- Set Open Case Condition and Owner ID Condition to the same expression in the new UI, save, reload → both fields should still be populated.

### Rollback instructions

### Labels & Review

### DIFF
--- a/src/saveToCase.js
+++ b/src/saveToCase.js
@@ -991,14 +991,7 @@ function promoteStashedCreateBindRelevants(mug) {
     }
 
     if (ownerIdRelevant) {
-        var redundantWithCaseTypeOrName = _.contains(
-            caseTypeAndNameRelevants,
-            ownerIdRelevant
-        );
-        var redundantWithOpenCase = ownerIdRelevant === mug.p.openCaseCondition;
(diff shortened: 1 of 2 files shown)


millerdev (approved):
Nice, I like how this turned out.

PR #1238, opened 2026-05-05
Remove reserved types check for Create section

jingcheng16:
## Product Description

Before
Screenshot 2026-05-05 at 12 04 06 PM
After the change, user will no longer see error messages.

At first I don't allow user to define `case_type`, `case_name`, `owner_id`, `name` again in the create section's property card. But after I did some compatibility check, I find we have a lot of forms already in this state. Since we allow it in the past, I don't want to raise error now.
The `case_name`, `case_type`, `owner_id` defined in property card will went to update node, and it will override whatever user defined in the dedicated field.
This change is just for backwards compatibility.

### Safety story

Safe, just remove some validation field so the validation rules matches the old rule.
### Rollback instructions

### Labels & Review

### DIFF
--- a/src/saveToCase.js
+++ b/src/saveToCase.js
@@ -48,10 +48,6 @@ function validatePropertyNameChars(val) {
     return null;
 }
 
-function validateCreatePropertyName(val) {
-    return validatePropertyNameChars(val);
-}
-
 function validateRelationshipChoice(val) {
     if (val && val !== "child" && val !== "extension") {
         return gettext("Relationship must be child or extension.");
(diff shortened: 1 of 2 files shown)


nospame (approved):
Should we consider giving users a notification that they should use the dedicated field and possibly deprecating this option (maybe by a `visible_if_present` attribute on those fields), or is it fine if they prefer to do it this way?

jingcheng16:
> Should we consider giving users a notification that they should use the dedicated field and possibly deprecating this option (maybe by a visible_if_present attribute on those fields), or is it fine if they prefer to do it this way?

It's fine if they prefer to do it this way. The value in the property card will just override the dedicated field.

PR #1239, opened 2026-05-05
Update wiki link

jingcheng16:
## Product Description

Replace a placeholder link with a real link

### Rollback instructions

### Labels & Review


nospame (approved):
This link says it's "restricted" - I assume that's just because it's on the public wiki but not published yet.

jingcheng16:
@nospame Thanks for catching it! I. just made it public

PR #1241, opened 2026-05-07
Allow index references in case properties

millerdev:
## Product Description
In the Case Management tab of the Form Builder, case property fields previously only accepted a hard-coded set of index prefixes (`parent`, `host`, and `user` when usercase was enabled). Any other index path was flagged with a validation warning, creating visual noise for forms that legitimately reference deeper or alternate index relationships. This change accepts arbitrary multi-segment index references, for example `parent/parent/foo`, `host/host/bar`, or `parent/host/parent/baz` — provided each segment follows the same naming rules already required for case property names: starts with a letter and contains only letters, digits, `_`, or `-`.

## Technical Summary
- Broadened the case-property validation regex in `src/caseManagement.js` to allow zero or more `segment/` prefixes before the final property name.
- Per-segment rules are unchanged, so previously-accepted values stay valid and clearly malformed input is still rejected.

## Feature Flag
`FORMBUILDER_SAVE_TO_CASE`

### Safety story
This is a strict relaxation of a single client-side validation regex. It accepts more strings than before and rejects nothing that was previously accepted, so existing forms and saved data are unaffected. There is no change to XML parsing, generation, or any server interaction. Confidence comes from local test runs and the targeted scope of the diff.

### Automated test coverage

Yes.

### Rollback instructions

### DIFF
--- a/tests/caseManagement.js
+++ b/tests/caseManagement.js
@@ -731,6 +731,8 @@ describe("The Case Management plugin", function () {
             "c-",
             "d1",
             "E1",
+            "parent/f",
+            "host/G",
         ]).forEach(value => {
             it(`should allow ${JSON.stringify(value)}`, function () {
                 mug.p.caseProperty = value;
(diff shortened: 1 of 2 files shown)


tests/caseManagement.js
```diff
@@ -731,6 +731,8 @@ describe("The Case Management plugin", function () {
             "c-",
             "d1",
             "E1",
+            "parent/f",
+            "host/G",
```
jingcheng16, line 735: Do you mind add "mother/a" here?

jingcheng16 (approved):

tests/caseManagement.js
jingcheng16: Just think it is great that our test indicates we will support arbitrary relationship identifiers

jingcheng16: Can be addressed in a follow up PR if you want to merge this PR to fix the issue ASAP.

millerdev: I intentionally kept the tests conservative and within the boundaries of what the Case Management tab currently allows, even though the code allows arbitrary identifiers. I would prefer to wait to add a test for a non-standard index identifier until we add a feature to validate the identifier against the case relationships in the domain.

PR #1242, opened 2026-05-09
Add question type ConnectWorkAreaUpdate to connect plugin

mkangia:
## Product Description

https://dimagi.atlassian.net/browse/CCCT-2397

On same lines as https://github.com/dimagi/Vellum/pull/1194 add a new question type for work area update.

image

## Technical Summary

This adds a new question type ConnectWorkAreaUpdate that has all xpaths which will get values from other questions in the forms like all other connect plugin mugs, including a image capture field.

## Feature Flag

`COMMCARE_CONNECT`

### Safety story

tested on staging. also tests added.
The blast radius is limited to projects with the connect feature flag enabled.

### Rollback instructions

### Labels & Review

### DIFF
--- a/src/commcareConnect.js
+++ b/src/commcareConnect.js
@@ -1,7 +1,7 @@
 /**
  * CommCare Connect plugin for Vellum
  *
- * This plugin add new mug types:
+ * This plugin adds new mug types:
  * - Learn Module
  * - Assessment Score
  * - Delivery Unit
@@ -393,6 +393,7 @@ let mugConfigs = {
                     widget: widgets.xPath,
                     serialize: mugs.serializeXPath,
                     deserialize: mugs.deserializeXPath,
+                    help: gettext('One of the valid values accepted by CommCare Connect like REQUEST_FOR_INACCESSIBLE')
                 },
                 reason: {
                     lstring: gettext("Reason"),
--- a/src/commcareConnect.js
+++ b/src/commcareConnect.js
@@ -351,6 +351,83 @@ let mugConfigs = {
             }),
             _.clone(logicSection),
         ],
+    },
+    ConnectWorkAreaUpdate: {
+        rootName: "work_area_update",
+        childNodes: [
+            {id: "work_area_id"},
+            {id: "status"},
+            {id: "reason"},
+            {id: "additional_details"},
+            {id: "photo_evidence"}
+        ],
+        mugOptions: util.extend(baseMugOptions, {
+            typeName: 'Work Area Update',
+            icon: 'fa fa-wrench',
+            init: mug => {
+                mug.p.work_area_id = "";
+                mug.p.status = "";
+                mug.p.reason = "";
+                mug.p.additional_details = "";
+                mug.p.photo_evidence = "";
+            },
+            spec: util.extend(baseSpec, {
+                work_area_id: {
+                    lstring: gettext("Work Area ID"),
+                    visibility: 'visible',
+                    presence: 'required',
+                    widget: widgets.xPath,
+                    serialize: mugs.serializeXPath,
+                    deserialize: mugs.deserializeXPath,
+                    help: gettext('XPath expression for the work area ID associated with this update.'),
+                },
+                status: {
+                    lstring: gettext("Status"),
+                    visibility: 'visible',
+                    presence: 'required',
+                    widget: widgets.xPath,
+                    serialize: mugs.serializeXPath,
+                    deserialize: mugs.deserializeXPath,
+                },
+                reason: {
+                    lstring: gettext("Reason"),
+                    visibility: 'visible',
+                    presence: 'required',
+                    widget: widgets.xPath,
+                    serialize: mugs.serializeXPath,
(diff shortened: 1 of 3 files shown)


src/commcareConnect.js
```diff
@@ -351,6 +351,94 @@ let mugConfigs = {
...
+            icon: 'fa fa-wrench',
+            init: mug => {
+                mug.p.work_area_id = "";
+                mug.p.status = "";
+                mug.p.reason = "";
+                mug.p.estimated_duration = "";
+                mug.p.additional_details = "";
+                mug.p.photo_evidence = "";
+            },
+            spec: util.extend(baseSpec, {
```
ajeety4, line 340: Confirming that the date of visit should be captured from the form's metadata ?

src/commcareConnect.js
```diff
@@ -351,6 +351,94 @@ let mugConfigs = {
...
+                },
+                additional_details: {
+                    lstring: gettext("Additional Details"),
+                    visibility: 'visible',
+                    presence: 'optional',
+                    widget: widgets.xPath,
+                    serialize: mugs.serializeXPath,
+                    deserialize: mugs.deserializeXPath,
+                },
+                photo_evidence: {
```
ajeety4, line 378: Confirming that this will resolve to the file name of the attachment ?
Also this does not allow for multiple photos right  ? I am not sure if we would need that though

src/commcareConnect.js
mkangia: That is correct yes.

src/commcareConnect.js
mkangia: Yes, the photo_evidence will resolve to the file name.

And, only one photo for now.

src/commcareConnect.js
```diff
@@ -1,10 +1,12 @@
 /**
  * CommCare Connect plugin for Vellum
  *
- * This plugin adds two new mug types:
+ * This plugin add new mug types:
```
ajeety4: nit: `adds`

src/commcareConnect.js
```diff
@@ -351,6 +351,83 @@ let mugConfigs = {
...
+                work_area_id: {
+                    lstring: gettext("Work Area ID"),
+                    visibility: 'visible',
+                    presence: 'required',
+                    widget: widgets.xPath,
+                    serialize: mugs.serializeXPath,
+                    deserialize: mugs.deserializeXPath,
+                    help: gettext('XPath expression for the work area ID associated with this update.'),
+                },
+                status: {
```
ajeety4, line 353: Assuming this is a text field, user can enter any value. 
Wondering if it is helpful to have a help text to indicate what values are allowed ?

src/commcareConnect.js
```diff
@@ -351,6 +351,83 @@ let mugConfigs = {
...
+                photo_evidence: {
+                    lstring: gettext("Photo Evidence"),
+                    visibility: 'visible',
+                    presence: 'required',
+                    widget: widgets.xPath,
+                    serialize: mugs.serializeXPath,
+                    deserialize: mugs.deserializeXPath,
+                    help: gettext('Reference to the image uploaded for photo evidence'),
+                }
+            })
```
ajeety4, line 387: Noticed that there is no relevantAttr which is present for other mugs. Just curious if there a reason for that ?

ajeety4 (approved):
Looks good.
Just a couple of non blocking queries

src/commcareConnect.js
mkangia: Claude helped me understand that the `relevantAttr` is actually already available for all connect plugin's mugs by default and since the ones added in this file are exact copy of the default setup available, they are redundant.

The relevant Attr is used to setup display condition for the question.

mkangia: removed the redundant attrs
https://github.com/dimagi/Vellum/pull/1242/commits/b9e5fb3ead209a1b5faffda58f02dbf893b5f9e5

thanks for the solid review :+1:

src/commcareConnect.js
mkangia: makes sense, added help text
https://github.com/dimagi/Vellum/pull/1242/commits/74baa2327275cc07ac38adc70afb5824838d0b3a

avoided the full list here since it could change on connect

src/commcareConnect.js
mkangia: that's embarrassing, updated

https://github.com/dimagi/Vellum/pull/1242/commits/74baa2327275cc07ac38adc70afb5824838d0b3a

ajeety4 (approved):
Look great. Thanks

millerdev (approved):
Reviewed lightly. I'm not familiar with how Connect is using these question types.

mkangia:
Thank you @millerdev 

> I'm not familiar with how Connect is using these question types.

not sure how to explain this well. Just usual CommCare Apps would have these questions that will be populated by either responses from users to other questions in the form or with data available via session or other non-user data elements. When a form filled with these questions reaches Connect, we parse it there for more data processing.

PR #1249, opened 2026-05-14
Add Edit Form Details modal for in Vellum

jingcheng16:
## Product Description

This PR is to replace the pencil button that injected by CommCare HQ into Vellum. now Vellum owns both the button and the modal.

This is to solve the issue that when update form name using the HQ injected modal, without reloading the page, Vellum will never know, thus the source XML which is written by Vellum won't contain the new `<title>`, but the `<title>` in server side has been changed, thus Vellum would think there is an conflict, thus gives a warning of overwriting changes.

https://github.com/user-attachments/assets/8fc059df-7fa2-4b65-acbb-b198ac96b144
 

## Technical Summary

Ticket: https://dimagi.atlassian.net/browse/SAAS-18436

Except for adding a new modal and set form's name and comment from the new modal, this commit also update the default icon for form in Vellum ( use icon for Survey ).

## Feature Flag

NA

### Safety story

I did thorough testing locally. The change in Vellum side is mainly UI, and sending the `name` and `comment` attr to existing `edit_form_attr` endpoint in HQ.

### Automated test coverage

New test added**`save POST carries name and comment`** in `tests/core.js`

### Rollback instructions

### Labels & Review

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -2356,6 +2356,8 @@ fn.send = function (formText, saveType) {
 
     data = saveType === 'patch' ? {patch: patch} : {xform: formText};
     data = this.augmentSentData(data, saveType);
+    data.name = this.data.core.form.formName;
+    data.comment = this.data.core.form.formComment;
     data.case_references = JSON.stringify(this.data.core.form._logicManager.caseReferences());
     if (checkForConflict) {
         data.sha1 = CryptoJS.SHA1(this.data.core.lastSavedXForm).toString();
--- a/src/core.js
+++ b/src/core.js
@@ -7,6 +7,7 @@ import CodeMirror from "codemirror";
 import main_template from "vellum/templates/main.html";
 import add_question from "vellum/templates/add_question.html";
 import edit_source from "vellum/templates/edit_source.html";
+import edit_form_details_modal from "vellum/templates/edit_form_details_modal.html";
 import confirm_overwrite from "vellum/templates/confirm_overwrite.html";
 import control_group_stdInput from "vellum/templates/control_group_stdInput.html";
 import form_errors_template from "vellum/templates/form_errors_template.html";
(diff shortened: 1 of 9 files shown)


src/core.js
```diff
@@ -1362,7 +1362,7 @@ fn.loadXFormOrError = function (formString, done, updateSaveButton) {
             if (_this.opts().core.formIconClass) {
                 _this.$f.find('.fd-form-icon').addClass(_this.opts().core.formIconClass);
             } else {
-                _this.$f.find('.fd-form-icon').addClass('fa fa-edit');
+                _this.$f.find('.fd-form-icon').addClass('fa fa-regular fa-file');
```
millerdev, line 1397: nit: Maybe use the registration form icon since the `caseManagmeent` plugin is enabled by deafult? (very minor, can ignore)

millerdev (approved):
What do you think of persisting the name and comment in local storage when the save button is clicked in stand-alone vellum? They should also be loaded on click _[Reload saved form]_.

jingcheng16:
@millerdev name and comment is now saved in `window.sessionStorage` 

Do not plan to address the icon I think it doesn't matter... ><

tests/main.js
```diff
@@ -104,18 +107,22 @@ import('jquery.vellum').then(() => Promise.all([
     }
     $('#run-tests').click(runTests);
 
-    function load(form, ready) {
+    function load(form, ready, formName, formComment) {
```
millerdev, line 110: nit: Would it be hard to move `ready` to the last parameter? That would allow removing the `undefined` argument passed in `loadFromSession` below.

millerdev (approved):

tests/main.js
jingcheng16: we have other callers of load: `load("", function () { mocha.run(); });` and `load(e.target.result);` . So if I change the sequence, those two functions will need to pass two `undefined` arguments...

PR #1250, opened 2026-05-14
Fix bubble interaction bugs in rich text editor

millerdev:
## Product Description

Improves how reference bubbles behave inside rich text inputs:

- Clicking a bubble now selects it as a single atomic unit and visibly highlights it (darker background, white text), matching the existing hover treatment.
- Shift-clicking past a bubble extends the selection through the bubble cleanly instead of snapping mid-bubble.
- Cut, copy, and paste of selections containing bubbles now produce the expected output and leave the editor in a valid state.
- Two adjacent bubbles no longer cause one to disappear when the other is edited or removed.

https://dimagi.atlassian.net/browse/SAAS-19367

A video demo of _before_ behavior is in the [ticket](https://dimagi.atlassian.net/browse/SAAS-19367
). Below is an _after_ video demo:

https://github.com/user-attachments/assets/351c55fb-9c51-47d9-bd20-d9d94aa46f76

This PR was written with the help of Claude Code.

## Technical Summary

- Centralize zero-width-space (ZWSP) boundary wrapping in `makeBubble` so every bubble owns its own leading and trailing ZWSP. Previously, `outputToBubble` and `bubbleExpression` produced inconsistent markup and adjacent bubbles in expression mode shared a single ZWSP, which let atomic edits on one bubble strip the neighbor's boundary and trigger collateral removal.
- Add a `mousedown` handler on `.label-datanode` bubbles that focuses the input and calls `setBaseAndExtent` across `ZWSP<span>…</span>ZWSP`. A transient `bubbleClickInProgress` flag suppresses the focus handler's "move caret to end" behavior during the click.
- Add a `selectionchange` listener that iterates all ranges and toggles a `.selected` class on intersected bubbles; the listener is detached in `destroy`. Styles added to suppresses the OS selection overlay so it doesn't compete with the class-based fill.
- Snap selection focus to the nearest bubble boundary and preserve the ZWSP invariant across cut/paste so the editor's input handler doesn't see broken boundaries.
- Simplify ZWSP cleanup now that boundaries are owned per-bubble: the "only trailing ZWSP left" branch and the "except last in editor" guard are removed in favor of a single rule — if a boundary is broken, strip remaining boundary chars and remove the span.
- Minor refactors: `insertHtmlWithSpace` → `insertHtmlWithSuffix`, consistent argument order with `form` last, removal of a superfluous `assertCopy` callback, and a `ZWSP` constant.

### Safety story

Changes are scoped to the rich text editor's bubble handling. The ZWSP-wrapping refactor is the highest-risk piece because it changes the DOM shape produced by `makeBubble`; existing expression-mode tests were updated to reflect the now-consistent wrapping, and the simplified cleanup rule is a strict subset of the previous defensive branches. Manual testing covered click-to-select, shift-click extension, keyboard selection across bubble boundaries, cut/copy/paste of single and adjacent bubbles, and deletion of one of two adjacent bubbles.

### Automated test coverage

New tests in `tests/richText.js` cover:

- Copy, cut, and paste of selections that include bubbles (the paste test reproduces the original bug report).
- Selecting a bubble by click and the resulting `.selected` class.
- Shift-click extending selection past a bubble.
- Selection focus snapping to bubble boundaries.

Existing expression-mode tests were updated to expect the now-consistent ZWSP wrapping in `toRichText` output.

### Rollback instructions

### DIFF
--- a/src/richText.js
+++ b/src/richText.js
@@ -372,6 +372,23 @@ var editor = function(input, form, options) {
         );
     }
 
+    function updateBubbleSelectedState() {
+        const selection = window.getSelection();
+        const editorRanges = [];
+        for (let index = 0; index < selection.rangeCount; index++) {
+            const range = selection.getRangeAt(index);
+            if (inputElement.contains(range.commonAncestorContainer)) {
+                editorRanges.push(range);
+            }
+        }
+        if (!editorRanges.length) { return; }
+        inputElement.querySelectorAll('.label-datanode').forEach(bubble => {
+            const isSelected = editorRanges.some(range => range.intersectsNode(bubble));
+            bubble.classList.toggle('selected', isSelected);
+        });
+    }
+    document.addEventListener('selectionchange', updateBubbleSelectedState);
+
     if (arguments.length === 1) {
         throw new Error("editor not initialized: " +
                         $("<div>").append(input).html());
@@ -475,6 +492,7 @@ var editor = function(input, form, options) {
         },
         destroy: function () {
             if (input !== null) {
+                document.removeEventListener('selectionchange', updateBubbleSelectedState);
                 input.removeData("editorWrapper");
                 input = null;
             }
--- a/src/richText.js
+++ b/src/richText.js
@@ -294,31 +294,18 @@ var editor = function(input, form, options) {
             if (zwspBeforeMissing || zwspAfterMissing) {
                 spansToRemove.push(span);
             }
-
-            // There is only the tailing ZWSP left
-            if (nextNode && nextNode.nodeType === Node.TEXT_NODE &&
-                    nextNode.parentNode.getAttribute('contenteditable') === 'true' &&
-                    nextNode.parentNode.lastChild === nextNode &&
-                    nextNode.nodeValue === ZERO_WIDTH_SPACE) {
-
-                spansToRemove.push(span);
-            }
         });
 
         spansToRemove.forEach(span => {
             const prevNode = span.previousSibling;
             const nextNode = span.nextSibling;
-            // remove the previous node or tailing ZWSP if it has one
             if (prevNode && prevNode.nodeType === Node.TEXT_NODE && prevNode.nodeValue.endsWith(ZERO_WIDTH_SPACE)) {
                 prevNode.nodeValue = prevNode.nodeValue.slice(0, -1);
                 if (prevNode.length === 0) {
                     prevNode.remove();
                 }
             }
-            // remove the next node or leading ZWSP if it has one
-            // except if it is the last one in the editor
-            if (nextNode && nextNode.nodeType === Node.TEXT_NODE && nextNode.nodeValue.startsWith(ZERO_WIDTH_SPACE) &&
-                !(nextNode.parentNode.lastChild === nextNode && nextNode.nodeValue === ZERO_WIDTH_SPACE)) {
+            if (nextNode && nextNode.nodeType === Node.TEXT_NODE && nextNode.nodeValue.startsWith(ZERO_WIDTH_SPACE)) {
                 nextNode.nodeValue = nextNode.nodeValue.slice(1);
                 if (nextNode.length === 0) {
                     nextNode.remove();
--- a/src/richText.js
+++ b/src/richText.js
@@ -101,7 +101,7 @@ var editor = function(input, form, options) {
     });
     observer.observe(inputElement, { childList: true, subtree: true });
 
-    function insertHtmlWithSpace(content, insertSpaces = false) {
+    function insertHtmlWithSuffix(content, suffix=' ') {
         const hasFocus = document.activeElement === inputElement;
         let range;
         if (!hasFocus && x && y) {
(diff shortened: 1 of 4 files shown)


src/richText.js
```diff
@@ -340,6 +340,38 @@ var editor = function(input, form, options) {
         y = e.clientY;
     });
 
+    let bubbleClickInProgress = false;
```
jingcheng16, line 347: Is this necessary?

src/richText.js
```diff
@@ -452,10 +436,7 @@ var editor = function(input, form, options) {
             if (options.isExpression) {
                 insertHtmlWithSpace(bubbleExpression(xpath, form), true);
             } else {
-                var output = makeBubble(form, xpath);
-                insertHtmlWithSpace($('<p>')
```
jingcheng16, line 407: Why `<p>` is removed?

tests/richText.js
```diff
@@ -239,21 +238,22 @@ describe("Rich text utilities", function() {
         var f_1065 = "#case/f_1065",
             ico = icon('fcc-fd-text'),
             opts = {isExpression: true},
+            wrap = function (bubble) { return ZWSP + html(bubble) + ZWSP; },
```
jingcheng16: Can you incorporate `wrap` into `makeBubble` so it mirrors the `makeBubble` in src code?

tests/richText.js
```diff
@@ -598,9 +598,9 @@ describe("The rich text editor", function () {
                     input_[0].dispatchEvent(clipboardEvent);
 
                     const text = removeZWSP(removeSpanId(input_[0].innerHTML));
-                    const expText = removeSpanId(richText
+                    const expText = removeZWSP(removeSpanId(richText
```
jingcheng16: can `removeZWSP` removed from both two lines?

tests/richText.js
```diff
@@ -654,7 +654,7 @@ describe("The rich text editor", function () {
             it("should make bubbles on converting to rich text: " + repr, function () {
                 var text = richText.toRichText(result, form, {isExpression: true}),
                     bubble = makeBubble('#form/text', 'text', icon('fcc-fd-text'), true),
-                    expected = (expr.slice(0, i) + html(bubble) + " " + expr.slice(i)),
+                    expected = (expr.slice(0, i) + ZWSP + html(bubble) + ZWSP + " " + expr.slice(i)),
```
jingcheng16: won't be necessary after incorporating `wrap` into `makeBubble`

src/richText.js
```diff
@@ -514,6 +509,28 @@ var editor = function(input, form, options) {
...
+        inputElement.querySelectorAll('.label-datanode').forEach(bubble => {
+            if (range.intersectsNode(bubble)) return;
+            const prev = bubble.previousSibling;
+            const next = bubble.nextSibling;
+            const leadingZwspInRange =
+                range.comparePoint(prev, prev.nodeValue.length - 1) === 0 &&
+                range.comparePoint(prev, prev.nodeValue.length) === 0;
+            const trailingZwspInRange =
+                range.comparePoint(next, 0) === 0 &&
+                range.comparePoint(next, 1) === 0;
```
jingcheng16: Is it possible to improve this?

jingcheng16 (commented):
Looks good to me and thank you for fix this! Just a few comments that we went over together.

src/richText.js
millerdev: Unfortunately yes. Without this, clicking a bubble does not select it if some other input (than the one with the clicked bubble) is focused.

I'd love to find a less hacky way to do it, and am open to suggestions.

src/richText.js
millerdev: It was a construction device (container used to append the output element), and was immediately discarded two lines below with `.html()`.

tests/richText.js
millerdev: 0523d2e43ee1d8b2af7b3831f88717cc0156de26

tests/richText.js
millerdev: 8d07e74705c2aec418658528c87ff2334c10853c

tests/richText.js
millerdev: 0523d2e43ee1d8b2af7b3831f88717cc0156de26

src/richText.js
millerdev: 3e17ddcb6066b0e976aa48c97dff3ea7b814caba - Attempted to improve comment and variable names. Did not eliminate [`range.comparePoint`](https://developer.mozilla.org/en-US/docs/Web/API/Range/comparePoint) entirely because it seems to be the best way to detect if the range includes a point in a text node (although did remove one unneeded call).

jingcheng16 (approved):

src/richText.js
MartinRiese: I don't think it is that hacky.

MartinRiese (approved):
Super nice PR. Love the commit messages and comments. It is really easy to follow what you are doing. And great suggestions from Jing. I don't have anything to add.

PR #1251, opened 2026-05-15
Refine task processing

mkangia:
## Product Description

For connect plugin, refine ConnectTask by allowing to set a task slug dynamically and removing the redundant fields for name and description which are actually entered on Connect and not HQ.

## Technical Summary

Was originally looking at an issue that task node was not present on the form submitted. While resolving that realized that the ConnectTask setup needed to allow passing the task slug dynamically instead of setting it as the node ID.
So, added that separate field which also resolved the original concern of task node not being present in the xml (details in the commit message).
Additionally removed the redundant fields.

## Feature Flag

`COMMCARE_CONNECT`

### Safety story

Testing on staging.

### Rollback instructions

### Labels & Review

### DIFF
--- a/src/commcareConnect.js
+++ b/src/commcareConnect.js
@@ -281,6 +281,7 @@ let mugConfigs = {
         childNodes: [
             {id: "name", writeToData: true},
             {id: "description", writeToData: true},
+            {id: "task_slug"},
         ],
         mugOptions: util.extend(baseMugOptions, {
             typeName: 'Task',
@@ -288,6 +289,7 @@ let mugConfigs = {
             init: mug => {
                 mug.p.name = "";
                 mug.p.description = "";
+                mug.p.task_slug = "";
             },
             spec: util.extend(baseSpec, {
                 nodeID: {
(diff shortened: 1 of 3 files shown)


calellowitz:
Is this going to break existing apps that configured the old setup?

calellowitz (commented):
all of the other connect question types use an id property rather than a separate xml tag within the block to indicate the slug/id. Why does this one need to do it differently?

mkangia:
Hey @calellowitz 

Thanks for the feedback.

> Is this going to break existing apps that configured the old setup?

For tasks, it won't break anything but the Task question would need to be reset up. I believe that is applicable on staging for anything done for testing. I have already notified Delivery and would do so for QA.
ConnectTask was added in https://github.com/dimagi/Vellum/pull/1194 for the re-learning module so there is no live usage for it yet.

> all of the other connect question types use an id property rather than a separate xml tag within the block to indicate the slug/id. Why does this one need to do it differently?

My understanding is the ID on the block is simply for the question type, which is static. But we need to pass the slug or the ID as a dynamic value just like we have for work area update, so we know the task to be marked as completed.

This is how I see this working,

image

just FYI
Additionally, I am also checking with Delivery if they have access to take unique ID itself, so they can pass that instead of task type slug. But that can be done as a follow up.

mkangia:
Thanks @calellowitz for questioning this.
Turns out this was not needed,
https://dimagi.atlassian.net/browse/CCCT-2422?focusedCommentId=467545

PR #1252, opened 2026-05-19
Locked questions: bugfixes and QA-raised issues

nospame:
## Product Description

A collection of bug fixes for the locked-questions feature, plus a related improvement to the global drag-and-drop behavior:

- Locked Android App Callout questions are now fully read-only — intent extras/responses can't be edited and add/remove buttons are hidden.
- Users without the locked-questions permission can no longer unlock a locked question by tampering with the disabled "Locked" checkbox in DevTools.
- The "This group contains locked questions" message clears once a group no longer contains any locked questions.
- Locked widget inputs (xpath, text, etc.) are no longer valid drop targets — dragging a question from the tree onto a locked input no longer mutates it. The drag helper shows the "error" icon and `jstree-drop-active` highlight is suppressed on disabled drop targets.
- Locked questions are skipped by the auto-case-name flow so it doesn't silently target a question the user can't edit.

## Technical Summary

Vellum side of https://dimagi.atlassian.net/browse/SAAS-19552

Notable changes:
- `src/core.js`: global drag and drop handler now consults `disableDropForEl` (checks `disabled` attribute and `contenteditable='false'`) before treating a `.jstree-drop` element as a valid drop target.
- `src/lock.js`: 
  - Setter for the `locked` property refuses to mutate when the user lacks `edit_locked_questions` permission (after XML load completes). 
  - Clears the `mug-has-locked-children` `nodeID` message on ancestor groups when their last locked descendant becomes unlocked.
  - The "Locked" checkbox itself is excluded from `isPropertyLocked` so it does not inadvertently lock itself.
- `src/widgets.js`: `baseKeyValue` sets `widget.path` so `isDisabled()` resolves through the existing lock check, and `setValue` now disables row inputs and hides add/remove buttons when the widget is disabled.
- `src/widgets/base.js`: a property with a custom `enabled` callback no longer overrides `isPropertyLocked` — locked wins.
- `src/dataSourceWidgets.js`: passes `isDisabled` through both `getUIElement` and `getUIElementWithEditButton`.
- `src/caseManagement.js`: auto-assigned case name now skips locked questions via a new `canAddCaseName` helper that consults `isPropertyLocked`.

## Feature Flag
In HQ, the majority of this functionality is gated behind `LOCKED_ADMIN_QUESTIONS` until release.

### Safety story

Each change is small and locally scoped. The lock-plugin guard against unlocking does not affect users with permission (it short-circuits only when `!isEditable && !isLoadingXForm`), and the drag and drop predicate is additive — existing valid drop targets behave exactly as before. The baseKeyValue change required setting widget.path, which is only used for validation and instance references. Both are safe no-ops for the androidIntentExtra and androidIntentResponse callers, since neither defines a validator or uses instance refs.

Verified locally:
- Android App Callout intent extras + responses become read-only on a locked question (inputs disabled, add/remove hidden).
- A question cannot be unlocked by unchecking the "Locked" checkbox after removing the `disabled` attribute in DevTools (vellum:lock attribute stays set).
- Dragging a tree question into a locked xpath/text/calculate input is rejected; the helper shows the error icon while hovering.
- Dragging into an unlocked input still works as before.
- Locking the last unlocked descendant of a group clears the ancestor's "contains locked questions" message.

### Automated test coverage

New tests in `tests/core.js` for the global drag and drop handler.

Existing `tests/formdesigner.lock.js` tests were updated to pass `edit_locked_questions: true` where they programmatically toggle the locked state (otherwise the new setter guard would silently no-op them).

### QA Plan

https://dimagi.atlassian.net/browse/QA-8497
QA have tested Locked Admin Questions with this branch and its fixes on staging.

### Rollback instructions

### Labels & Review

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/caseManagement.js
+++ b/src/caseManagement.js
@@ -419,8 +419,7 @@ function autoAssignName(vellum) {
     vellum.ensureCurrentMugIsSaved(() => {
         const form = vellum.data.core.form;
         let mug = form.findFirstMatchingChild(null, () => true);
-        if (!mug || mug.p.caseProperty || !mug.spec.caseProperty ||
-                mug.getPresence("caseProperty") !== 'optional') {
+        if (!canAddCaseName(mug)) {
             mug = vellum.addQuestion('DataBindOnly', 'first');
             mug.p.nodeID = form.generate_question_id('case-name');
             mug.p.calculateAttr = 'uuid()';
@@ -430,6 +429,14 @@ function autoAssignName(vellum) {
         vellum.data.core.saveButton.fire('change');
         vellum.setCurrentMug(mug);
     });
+}
+
+function canAddCaseName(mug) {
+    if (!mug) {
+        return false;
+    }
+    const canSetCaseProperty = mug.spec.caseProperty && mug.getPresence("caseProperty") === 'optional';
+    return canSetCaseProperty && !mug.p.caseProperty && !mug.form.vellum.isPropertyLocked(mug, "caseProperty");
 }
 
 function refreshCurrentMug(vellum) {
--- a/tests/caseManagement.js
+++ b/tests/caseManagement.js
@@ -883,6 +883,30 @@ describe("The Case Management plugin", function () {
             });
         });
 
+        it("new mug when first mug's Case Property is locked", function (done) {
+            util.loadXML("");
+            const mug = util.addQuestion("Text", "first");
+            const vellum = mug.form.vellum;
+            const originalIsPropertyLocked = vellum.isPropertyLocked;
+            vellum.isPropertyLocked = () => true;
+            const save = call("getData").core.saveButton.ui;
+            function test() {
+                save.off('shown.bs.popover.test');
+                try {
+                    $(".fd-auto-assign-case-name").trigger("click");
+                    assert.isUndefined(mug.p.caseProperty, "locked mug should not receive case name");
+                    const name = call("getMugByPath", "/data/case-name");
+                    assert.equal(name.p.nodeID, 'case-name');
+                    assert.equal(name.p.caseProperty, 'name');
+                } finally {
+                    vellum.isPropertyLocked = originalIsPropertyLocked;
+                }
+                done();
+            }
+            save.on('shown.bs.popover.test', test);
+            save.popover("show");
+        });
+
         it("new mug with unique node ID", function (done) {
             util.loadXML("");
             const mug = util.addQuestion("Text", "case-name");
--- a/src/core.js
+++ b/src/core.js
@@ -1119,14 +1119,21 @@ fn._createJSTree = function () {
  * drag begins.
  */
 $(document).one("dnd_move.vakata.jstree", function (e, data) {
+    const disableDropForEl = function (el) {
+        return el.getAttribute('disabled') !== null || el.getAttribute('contenteditable') === 'false';
+    };
+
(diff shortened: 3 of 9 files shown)


src/caseManagement.js
```diff
@@ -432,6 +431,14 @@ function autoAssignName(vellum) {
     });
 }
 
+function canAddCaseName(mug) {
+    if (!mug) {
+        return false;
+    }
+    const canSetCaseProperty = mug.spec.caseProperty && mug.getPresence("caseProperty") === 'optional';
```
millerdev, line 438: Can this be simplified? The next line has `&& !mug.p.caseProperty`, which I think makes the final condition something like
```js
return mug.p.caseProperty && ... && !mug.p.caseProperty && ...
```
which I think will never be true.

```suggestion
    const canSetCaseProperty = mug.getPresence("caseProperty") === 'optional';
```

tests/caseManagement.js
```diff
@@ -883,6 +883,30 @@ describe("The Case Management plugin", function () {
...
+            const originalIsPropertyLocked = vellum.isPropertyLocked;
+            vellum.isPropertyLocked = () => true;
+            const save = call("getData").core.saveButton.ui;
+            function test() {
+                save.off('shown.bs.popover.test');
+                try {
+                    $(".fd-auto-assign-case-name").trigger("click");
+                    assert.isUndefined(mug.p.caseProperty, "locked mug should not receive case name");
+                    const name = call("getMugByPath", "/data/case-name");
+                    assert.equal(name.p.nodeID, 'case-name');
```
millerdev, line 899: nit: this assertion seems redundant given the `getMugByPath` call on the previous line.

```suggestion
```

millerdev (dismissed):

src/caseManagement.js
nospame: I don't think it's quite true that we're looking at `mug.p.caseProperty && ... && !mug.p.caseProperty && ...`. The checks are `mug.spec.caseProperty` to see that caseProperty exists on the mug spec, and later `!mug.p.caseProperty` to see that caseProperty isn't already set to a value.

The current logic expands out to something like this:
```javascript
    const casePropertySpec = mug.spec.caseProperty;
    const casePropertyIsOptional = mug.getPresence("caseProperty") === 'optional';  // throws an error if property is undefined on mug.spec
    const canSetCaseProperty = casePropertySpec && casePropertyIsOptional;

    const casePropertyCurrentValue = mug.p.caseProperty;
    const casePropertyWidgetLocked = mug.form.vellum.isPropertyLocked(mug, "caseProperty");
    return canSetCaseProperty && !casePropertyCurrentValue && !casePropertyWidgetLocked;
```

It looks like we set `caseProperty` on the spec for mug types that have a databind. In practice, this is all top-level question types, I think it just excludes things like choices and itemsets. So effectively, yes we could remove the `mug.spec.caseProperty` check, `mug.getPresence("caseProperty") === 'optional'` there would be enough.

I don't think the check is strictly redundant (there could probably _theoretically_ be a top-level question with no databind), but I can remove it if you'd prefer. What do you think?

millerdev: Ignore me. `mug.spec.caseProperty` is not `mug.p.caseProperty`

millerdev (approved):

PR #1253, opened 2026-06-04
Skip advanced case actions when cascading lock from a group

nospame:
## Product Description
Advanced Case Actions cannot be fully locked right now, so the UI shouldn't pretend that they can be. They will also not contribute toward a group's 'unlock' icon: a group with all locked questions except for Advanced Case Actions will still have the 'lock' icon.

## Technical Summary
https://dimagi.atlassian.net/browse/SAAS-19853
Fixes a bug where Advanced Case Actions in a group would _appear_ to be locked after locking the parent group, but the state would not be maintained. Advanced Case Actions do not have a `<bind>` node in the XML, so to actually implement locking on these questions will be a larger change at a later date. Instead, this PR makes sure these questions don't appear to be locked when they shouldn't / aren't really.

### Safety story
Tested locally to see the intended effects and ran `yarn build` to ensure the code builds correctly.

### Automated test coverage
Added automated tests to see that "SaveToCase" (Advanced Case Actions) questions do not appear to get locked when a parent group is locked, and do not contribute to a parent group's `unlock` icon.

### QA Plan
Not planning formal QA.

### Rollback instructions

### Labels & Review

### DIFF
--- a/src/lock.js
+++ b/src/lock.js
@@ -19,6 +19,7 @@ const SELECT_AND_CHOICE_CLASSES = [...STATIC_SELECT_CLASSES, "Choice"];
 const DYNAMIC_SELECT_CLASSES = ["SelectDynamic", "MSelectDynamic"];
 const SELECT_CLASSES = [...STATIC_SELECT_CLASSES, ...DYNAMIC_SELECT_CLASSES];
 const GROUP_CLASSES = ["Group", "Repeat", "FieldList"];
+const NON_LOCKABLE_CLASSES = ["SaveToCase"];  // TODO: allow Advanced Case Actions to be locked
 
 $.vellum.plugin("lock", {}, {
     init: function () {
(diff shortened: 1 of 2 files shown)


jingcheng16:
> Advanced Case Actions do not have a <bind> node in the XML

I guess because I didn't review the most of the locked admin questions PR so I need some help here. Advanced Case Actions don't have a single <bind> node for the question it self, but it has a lot of other `<bind>` nodes with `nodeset="/data/question1/case/xxxxx`, can those be used?

jingcheng16 (approved):
Evan and I offlined on this, Advanced Case Actions will have no bind nodes if you don't set up anything.

nospame:
Force-pushed to rebase onto latest `master` with node pinned so tests run correctly. Seems like this doesn't require reapproval since it's effectively a clean merge with `master`.

PR #1258, opened 2026-06-10
Allow Advanced Case Actions to be locked

nospame:
## Product Description
Allows Advanced Case Actions questions to be locked in the form builder via the "Locked" widget.
image

## Technical Summary
Vellum side of https://dimagi.atlassian.net/browse/SAAS-19855
HQ PR: https://github.com/dimagi/commcare-hq/pull/37817
Updates the lock plugin to add and look for a `vellum:lock` attribute on a question's data node, when there is no `<bind>` node (i.e. with Advanced Case Actions). Tweaks some templates and disabling functionality in the saveToCase and caseManagement plugins to ensure a widget's disabled attribute is respected when set and when toggled (as in locking and unlocking a question).

These changes are UI-only; this will need corresponding changes in HQ (in progress) to ensure we validate correctly against changing a modifying Advanced Case Actions question.

### Safety story
Tested each change locally and on staging, including seeing that all elements and dropdowns in the Advanced Case Actions are functionally disabled, and that other question types are unaffected (or fixing those effects as needed). As noted above, these are UI-only changes and should be deployed after, or at the same time as HQ changes to add backend validation.

### Automated test coverage
Added automated tests to see that Advanced Case Actions can be locked, unlocked, and saved correctly.

### QA Plan
Not planning formal QA. Will test in depth on staging before planning release.

### Rollback instructions

### Labels & Review

### DIFF
--- a/src/caseManagement.js
+++ b/src/caseManagement.js
@@ -444,11 +444,10 @@ function refreshCurrentMug(vellum) {
     if (mugs.length !== 1) { return; }
     const select = $(".fd-content-right").find('fieldset[data-slug="caseManagement"]').find("select");
     if (select.length) {
-        const mappings = vellum.data.caseManagement.caseMappingsByQuestion;
         const widget = widgets.util.getWidget(select, vellum);
         widget.setValue(mugs[0].p.caseProperty || "");
+        select.prop('disabled', widget.isDisabled());
         select.trigger("change.select2");
-        select.prop('disabled', (mappings[mugs[0].absolutePath]?.length || 0) > 1);
     }
 }
 
--- a/src/widgets/base.js
+++ b/src/widgets/base.js
@@ -180,8 +180,10 @@ export var getUIElement = function($input, labelText, isDisabled, help) {
     }));
 
     if (isDisabled) {
-        // Disable anything that can be disabled
-        $input.find("*").addBack().prop('disabled', true);
+        // Disable anything that can be disabled. Exclude <option>/<optgroup>:
+        // a disabled option cannot be selected, which breaks select2's display
+        // of the current value (it falls back to the placeholder).
+        $input.find("*").addBack().not("option, optgroup").prop('disabled', true);
         $input.filter('[contenteditable]').attr({
             'contenteditable': false,
             'disabled': true,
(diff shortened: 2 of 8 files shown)


jingcheng16:
@nospame Can you ping me when this is ready to review? I think I don't get notification for that.

jingcheng16:
~~Do you have a better understanding of the `bind` node and `data` node after this PR? Can you share that knowledge with me? 😂~~
nvm I just googled and refreshed my memory. 😛

src/widgets/base.js
```diff
@@ -180,8 +180,10 @@ export var getUIElement = function($input, labelText, isDisabled, help) {
     }));
 
     if (isDisabled) {
-        // Disable anything that can be disabled
-        $input.find("*").addBack().prop('disabled', true);
+        // Disable anything that can be disabled. Exclude <option>/<optgroup>:
+        // a disabled option cannot be selected, which breaks select2's display
+        // of the current value (it falls back to the placeholder).
+        $input.find("*").addBack().not("option, optgroup").prop('disabled', true);
```
jingcheng16, line 186: What is `optgroup` and where is it used?

src/caseManagement.js
jingcheng16, line 1: I'm currently on commit bd836796b51d1d4f38fd2858ab7da49a28077cbb. I'm curious, before this PR, does the caseManagement dropdown displays the default value when locked? Is my understanding correct, that by figuring out how to lock ACA (unofficial abbreviation... I'm just lazy), you identified a bug with case management dropdown?

jingcheng16 (approved):
Looks great to me. Thank you for making this change!
Just curious, after this PR, do you find it is easier to just bring all mug type's lock attr to data node from bind node?

src/widgets/base.js
nospame: We actually don't use `<optgroup>` at all in Vellum. It is a standard HTML element used to group `<option>`s inside a `<select>`. Because it can use the `disabled` attribute, which would disable any `<option>`s within it, this proactively excludes them, but it's not strictly necessary.

src/caseManagement.js
nospame: It didn't display the default value when locked. I'm not entirely sure why but the behavior with the caseManagement dropdown was a little bit different. The issue there was that, on save, the dropdown would become re-enabled, which is weird on its own but becomes a functional issue when its options are selectable.


https://github.com/user-attachments/assets/3d6db5a4-0c9c-4bca-b323-845625cb2526

